PR #63, opened 2026-01-21
Add getting started tutorial

burkeholland:
Getting started tutorial with examples in TypeScript, Python, Go and .NET

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,0 +1,884 @@
+# Build Your First Copilot-Powered App
+
+In this tutorial, you'll use the Copilot SDK to build a command-line assistant. You'll start with the basics, add streaming responses, then add custom tools - giving Copilot the ability to call your code.
+
+**What you'll build:**
+
+```
+You: What's the weather like in Seattle?
+Copilot: Let me check the weather for Seattle...
+         Currently 62°F and cloudy with a chance of rain.
+         Typical Seattle weather!
+
+You: How about Tokyo?
+Copilot: In Tokyo it's 75°F and sunny. Great day to be outside!
+```
+
+## Prerequisites
+
+Before you begin, make sure you have:
+
+- **GitHub Copilot CLI** installed and authenticated ([Installation guide](https://docs.github.com/en/copilot/how-tos/set-up/install-copilot-cli))
+- Your preferred language runtime:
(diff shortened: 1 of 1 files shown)


steveward (approved):
Looks great from a docs perspective! ✨

docs/getting-started.md
```diff
@@ -0,0 +1,884 @@
...
+<details open>
+<summary><strong>Node.js / TypeScript</strong></summary>
+
+Create `index.ts`:
+
+```typescript
+import { CopilotClient } from "@github/copilot-sdk";
+
+const client = new CopilotClient();
+const session = await client.createSession({ model: "gpt-4.1" });
```
friggeri, line 108: let's default to a more recent model?

docs/getting-started.md
```diff
@@ -0,0 +1,884 @@
...
+session.on((event: SessionEvent) => {
+    if (event.type === "assistant.message_delta") {
+        process.stdout.write(event.data.deltaContent);
+    }
+    if (event.type === "session.idle") {
+        console.log(); // New line when done
+    }
+});
+
+await session.sendAndWait({ prompt: "Tell me a short joke" });
```
friggeri, line 261: I think it would be nice to have an example that uses `send` not just `sendAndWait`. Maybe something that prompts the user to input a message and then sends it?

docs/getting-started.md
```diff
@@ -0,0 +1,884 @@
+# Build Your First Copilot-Powered App
```
friggeri, line 1: would it be worth splitting having two tutorials one for python/one for node?

docs/getting-started.md
burkeholland: We can, but my concern was that we want it to be an included model and GPT-5-mini is just so slow. 

in any event - yes - can update.

docs/getting-started.md
burkeholland: @ashleywolf - thoughts here? Right now everything is in one tutorial and you get to pick your language. But we could punt out to separate files for just Node/Python instead of trying to show everything. Would be cleaner but we lose the Go/.NET.

ashleywolf: I'd keep as is for now

patniko (approved):

PR #81, opened 2026-01-21
Make the .NET library NativeAOT compatible

stephentoub:
- Enabled AOT analyzers
- Disabled STJ reflection in the test project (to help vet NAOT correctness)
- Use source generation for all types that may be serialized
- Remove all use of anonymous types
- Removed `<autogenerated/>` from the source generated code, as it was suppressing the analyzers
- Added support for propagating StreamJsonRpc's tracing to the CopilotClient's ILogger. I used this for debugging and decided to leave it
- Updated StreamJsonRpc to a newly published version on nuget to pick up NativeAOT fixes
- Cleaned up some formatting in the session types generator, in particular using a file-scoped namespace and removing the top-level indentation

### DIFF
--- a/dotnet/test/ToolsTests.cs
+++ b/dotnet/test/ToolsTests.cs
@@ -5,6 +5,7 @@
 using GitHub.Copilot.SDK.Test.Harness;
 using Microsoft.Extensions.AI;
 using System.ComponentModel;
+using System.Text.Json;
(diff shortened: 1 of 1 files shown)


agocke (approved):

SteveSandersonMS (approved):
Thanks, this looks great!

PR #127, opened 2026-01-23
docs: Update Python cookbook examples

RbBtSn0w:
- Convert all 5 recipe examples to async/await patterns matching current SDK interface
  * error_handling.py: Async/await conversion with proper event handling
  * managing_local_files.py: Async conversion with non-blocking user input
  * multiple_sessions.py: Full async/await conversion with proper cleanup
  * persisting_sessions.py: Async/await conversion with session management updates
  * pr_visualization.py: Async conversion with async subprocess handling

Root cause: SDK interface changed from synchronous to asynchronous (async/await) API. All examples now use correct async patterns with proper error handling and cleanup."

### DIFF
--- a/cookbook/python/recipe/error_handling.py
+++ b/cookbook/python/recipe/error_handling.py
@@ -2,6 +2,7 @@
 
 import asyncio
 from copilot import CopilotClient
+from copilot.generated.session_events import SessionEventType
 
 async def main():
     client = CopilotClient()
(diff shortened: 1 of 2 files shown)


RbBtSn0w:
> Is there a reason for moving away from kwargs to dict?

Yes, the move to a dictionary is due to changes in the underlying SDK’s create_session method signature.

In the latest async implementation (see python/copilot/client.py), the method is defined as:async def create_session(self, config: Optional[SessionConfig] = None)

It now expects a single config dictionary rather than individual keyword arguments (**kwargs). The official docstrings and E2E tests in the SDK also follow this pattern (passing a dict for model, tools, and other configurations). I updated the cookbook examples to align with this new interface.

PR #304, opened 2026-01-30
Fail CI if snapshots aren't present

friggeri:


nodejs/test/e2e/session.test.ts
```diff
@@ -387,7 +387,9 @@ describe("Send Blocking Behavior", async () => {
         expect(events).toContain("assistant.message");
     });
 
-    it("sendAndWait throws on timeout", async () => {
+    // Skip in CI - this test validates client-side timeout behavior, not LLM responses.
+    // The test intentionally times out before receiving a response, so there's no snapshot to replay.
```
SteveSandersonMS, line 388: I don't follow - it should get back a response instructing it to do the shell tool call, so that will be snapshotted won't it?

SteveSandersonMS (dismissed):
Other than hopefully not needing to skip that one test in CI, LGTM!

nodejs/test/e2e/session.test.ts
friggeri: it doesn't get snapshotted, because we timeout on purpose before the response gets here

SteveSandersonMS: OK I see. I thought it was timing out while executing the shell call (because of the `sleep 2` part of it) but I see it actually times out before receiving the initial response, so this makes sense.

PR #369, opened 2026-02-04
fix(python): populate usage_info when resuming session

### LINKED ISSUE #312: Resuming session doesn't automatically fetch 'session.usage_info' until a prompt is sent
On py module. I'm creating a '{currentTokens}/{maxTokens}' indicator for the current session, and that data is only stored in SessionEvent 'session.usage_info', which only fires upon sending a prompt, and not on resuming a session through CopilotClient.resume_session. This causes the init phase (period between session start and first prompt sent) where I just can't know what the current context tokens used are.

I know 'session.usage_info' is an event, but maybe also store this data directly onto the session object as well?

Coderxrohan:
### Issue

When resuming a session, `session.usage_info` is not populated until a prompt is sent, making it impossible to access token usage during the initial resumed session phase.

### Fix

Populate `usage_info` directly from the `session.resume` response and attach it to the `CopilotSession` object during resume.

### Impact

Token usage is now immediately available after `resume_session`, enabling clients to display current usage without waiting for the first prompt.

### Testing

- Resumed an existing session locally
- Verified `session.usage_info` is available immediately after resume

Fixes: #312


SteveSandersonMS:
Closing as stale

PR #374, opened 2026-02-04
Move to slnx

jaredpar:
This is the modern solution format for .NET. The advantages are that it has a defined format and more resistant to merge conflicts. This is supported by all modern .NET Tools (VS and VS Code)


SteveSandersonMS (approved):
Awesome, thanks!

PR #396, opened 2026-02-07
Update @github/copilot to 0.0.405

devm33:


jmoseley (approved):
Where do you change log entries come from the SDK? It'd be nice to have the addition of the request ID field recorded in the change log

devm33:
> Where do you change log entries come from the SDK? It'd be nice to have the addition of the request ID field recorded in the change log

I don't think there is a changelog for the sdk. Just the PR titles on the releases.

PR #414, opened 2026-02-09
[go] support bundling and auto installing the GitHub Copilot CLI

qmuntal:
This pull request introduces support for embedding the GitHub Copilot CLI binary directly within Go applications using the SDK, removing the need for users to have the CLI installed separately. It adds a robust mechanism for embedding, unpacking, and managing the Copilot CLI binary, including versioning, file locking for concurrency, and hash validation. The changes also include comprehensive tests and documentation updates.

The bundling works as follows:

```bash
# One-time setup step per project
go get -tool github.com/github/copilot-sdk/go/cmd/bundler 

# Downloads the CLI
# Places it at the current directory
# Creates a Go file that embed the CLI
# Instruct the SDK to use it
go tool bundler
```

### DIFF
--- a/go/cmd/bundler/main.go
+++ b/go/cmd/bundler/main.go
@@ -3,11 +3,12 @@
 //
 // Usage:
 //
-//	go run github.com/github/copilot-sdk/go/cmd/bundler [--platform GOOS/GOARCH] [--output DIR] [--cli-version VERSION]
+//	go run github.com/github/copilot-sdk/go/cmd/bundler [--platform GOOS/GOARCH] [--output DIR] [--cli-version VERSION] [--check-only]
 //
 //	--platform: Target platform using Go conventions (linux/amd64, linux/arm64, darwin/amd64, darwin/arm64, windows/amd64, windows/arm64). Defaults to current platform.
 //	--output: Output directory for embedded artifacts. Defaults to the current directory.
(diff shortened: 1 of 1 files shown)


go/cmd/bundler/main.go
```diff
@@ -0,0 +1,610 @@
+// Bundler downloads Copilot CLI binaries and packages them as a binary file,
+// along with a Go source file that embeds the binary and metadata.
+//
+// Usage:
+//
+//	go run github.com/github/copilot-sdk/go/cmd/bundler [--platform GOOS/GOARCH] [--output DIR] [--cli-version VERSION]
+//
+//	--platform: Target platform using Go conventions (linux/amd64, linux/arm64, darwin/amd64, darwin/arm64, windows/amd64, windows/arm64). Defaults to current platform.
+//	--output: Output directory for embedded artifacts. Defaults to the current directory.
+//	--cli-version: CLI version to download. If not specified, automatically detects from the copilot-sdk version in go.mod.
```
SteveSandersonMS, line 10: Is there something that forces developers to realise if their local bundled version is out of date?

qmuntal: Good point. I've added a new flag, `--check-only`, that will verify that the bundled version matches the one derived from the go.mod file.

PR #463, opened 2026-02-13
Allow custom npm registry URL for downloading Copilot from dotnet SDK

nayanshah:
Our repos enforce using private feeds and npm registry to fetch packages. But the .NET SDK attempts to download Copilot package from public npm registry. 

This change allows customizing the registry url by defined a MSBuild property in the csproj or props file - `CopilotNpmRegistryUrl`.

Note: an improvement could be attempting to find and parse any existing `.npmrc` files in any parent directories, but that ends up complicating this target a bit, so skipping for now


nayanshah:
@SteveSandersonMS thoughts?

PR #466, opened 2026-02-13
More e2e nodejs tests

friggeri:
As titled


nodejs/test/e2e/client_lifecycle.test.ts
```diff
@@ -0,0 +1,57 @@
...
+            events.push(event);
+        });
+
+        try {
+            const session = await client.createSession();
+
+            await session.sendAndWait({ prompt: "Say hello" });
+
+            // Wait for session data to flush to disk
+            await new Promise((r) => setTimeout(r, 500));
```
SteveSandersonMS, line 44: Can you make sure none of the tests rely on timings like this? Should poll until the desired condition is met.

Issue #341, opened 2026-02-03
Support async context manager for CopilotClient


pamelafox:
As a best practice, many SDKs that have Client() classes that require close()ing also support Python's async context manager protocol, like the Azure Python SDKs. That way, developers can write "async with CopilotClient() as client" and then not worry about explicitly closing. You can add support by implementing __aenter__ and __aexit__ dunder methods.

My motivation for asking:

Using the Python SDK, I mistakenly neglected to call session.destroy()/client.stop() and managed to use up all of my Mac's memory when running evals (200 copilot server processes).

PR #475, opened 2026-02-15
feat: add async context manager support for CopilotClient and Copilot…

Sumanth007:
### Session with automatic resource cleanup #341
This pull request introduces async context manager support for both `CopilotClient` and `CopilotSession`, enabling automatic resource cleanup and following Python best practices for resource management. The documentation is updated to recommend this pattern, and comprehensive tests ensure correct behavior and error handling. These changes make it easier and safer to use the SDK, especially in batch or evaluation scenarios.

Async context manager support and resource management:

* Added async context manager methods (`__aenter__` and `__aexit__`) to `CopilotClient` and `CopilotSession` for automatic startup, cleanup, and error handling, ensuring resources are always released—even if exceptions occur. [[1]](diffhunk://#diff-dcdffbe9a69136dca30ad65a8fecf97dd3e282205a7805110f4c842e496bb184R211-R256) [[2]](diffhunk://#diff-4e9625093c2c42e798587a63a39d2d1774b45582c089e8c2c8c0d6bc9cc01aceR87-R130)
* Updated the `python/README.md` to document and recommend the context manager usage pattern, including example code and benefits, and highlighted this feature in the capabilities list. [[1]](diffhunk://#diff-217ed82f87b78b399e304b07a159b27dff327c0f83adf4a2fc30b03bcbf84b01R17-R55) [[2]](diffhunk://#diff-217ed82f87b78b399e304b07a159b27dff327c0f83adf4a2fc30b03bcbf84b01R98) [[3]](diffhunk://#diff-217ed82f87b78b399e304b07a159b27dff327c0f83adf4a2fc30b03bcbf84b01R183-R220)

Testing and validation:

* Added new end-to-end tests in `python/e2e/test_context_managers.py` to verify correct context manager behavior, including automatic cleanup, error propagation, and nested context scenarios.
* Introduced unit tests in `python/test_client.py` to confirm that context manager methods return the correct values and propagate exceptions as expected.

Internal improvements:

* Imported `logging` and `TracebackType` in relevant files to support error logging and context manager implementation. [[1]](diffhunk://#diff-dcdffbe9a69136dca30ad65a8fecf97dd3e282205a7805110f4c842e496bb184R17-R25) [[2]](diffhunk://#diff-4e9625093c2c42e798587a63a39d2d1774b45582c089e8c2c8c0d6bc9cc01aceR10-R12)

### DIFF
--- a/python/copilot/client.py
+++ b/python/copilot/client.py
@@ -891,7 +891,7 @@ class CopilotClient:
 
         return (host, port)
 
-    async def __aenter__(self) -> "CopilotClient":
+    async def __aenter__(self) -> CopilotClient:
         """
         Enter the async context manager.
 


python/copilot/client.py
```diff
@@ -206,6 +208,56 @@ def __init__(self, options: Optional[CopilotClientOptions] = None):
...
+
+        Returns:
+            False to propagate any exception that occurred in the context.
+        """
+        try:
+            stop_errors = await self.stop()
+            # Log any session destruction errors
+            if stop_errors:
+                for error in stop_errors:
+                    logging.warning("Error during session cleanup in CopilotClient: %s", error)
```
brettcannon: I'm a bit concerned about this as not everyone wants 'logging' in their code as they may use a different logger. Plus this is too broad regardless as there isn't a way to filter out the logging for the library (i.e. this call uses the root logger).

I think more thought has to go into whether there's a desire for logging, and if there is then how to handle it.

python/copilot/client.py
```diff
@@ -206,6 +208,56 @@ def __init__(self, options: Optional[CopilotClientOptions] = None):
...
+        except Exception:
+            # Log the error but don't raise - we want cleanup to always complete
+            logging.warning("Error during CopilotClient cleanup", exc_info=True)
+        return False
```
brettcannon: This is incorrect if you're trying to suppress the exception; [you would want to return `True`](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement:~:text=If%20the%20suite%20was%20exited%20due%20to%20an%20exception%2C%20and%20the%20return%20value%20from%20the%20__exit__()%20method%20was%20false%2C%20the%20exception%20is%20reraised.) in that instance.

python/copilot/client.py
```diff
@@ -206,6 +208,56 @@ def __init__(self, options: Optional[CopilotClientOptions] = None):
...
+            exc_tb: The traceback of the exception that occurred, if any.
+
+        Returns:
+            False to propagate any exception that occurred in the context.
+        """
+        try:
+            stop_errors = await self.stop()
+            # Log any session destruction errors
+            if stop_errors:
+                for error in stop_errors:
```
brettcannon: If `StopError` was an exception then you would want to use `ExceptionGroup`, but since that's a Python 3.11 thing and this library supports 3.9 it would need a shim for 3.9 and 3.10.

python/copilot/session.py
```diff
@@ -82,6 +85,50 @@ def __init__(self, session_id: str, client: Any, workspace_path: Optional[str] =
...
+
+        Returns:
+            False to propagate any exception that occurred in the context.
+        """
+        try:
+            await self.destroy()
+        except Exception:
+            # Log the error but don't raise - we want cleanup to always complete
+            logging.warning("Error during CopilotSession cleanup", exc_info=True)
+        return False
```
brettcannon: I don't think the exception that was raised should be suppressed without at least a configuration option.

python/e2e/test_context_managers.py
```diff
@@ -0,0 +1,156 @@
...
+
+import pytest
+
+from copilot import CopilotClient
+
+from .testharness import CLI_PATH
+
+pytestmark = pytest.mark.asyncio(loop_scope="module")
+
+
```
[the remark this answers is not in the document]
SteveSandersonMS: To clarify this, please make sure that any new snapshots are created automatically by running the tests. Don't write them manually.

Also I'm not sure why it would be relevant to have new E2E tests that make LLM calls for this change given that asynchrony is an implementation detail and doesn't change the set of usage scenarios.

python/copilot/client.py
Sumanth007: Agreed on the semantics. The updated ```__aexit__``` always return False to propagate the original context exception. Cleanup exceptions are only suppressed when ```exc_type``` is not ```None```— meaning something already went wrong inside the ```with``` block and we don't want the cleanup error to mask it. That suppression happens via a bare except (not return True), so the original context exception still propagates correctly. In all other cases, cleanup errors are re-raised directly.

python/copilot/session.py
Sumanth007: Addressed. CopilotSession. ```__aexit__``` now uses the same pattern as ```CopilotClient.__aexit__: ``` cleanup exceptions from ```destroy()``` are only suppressed when the context block already raised an exception (exc_type is not None), to avoid masking the original error. If nothing went wrong inside the with block, any cleanup error from destroy() is re-raised directly so the caller sees it.

python/e2e/test_context_managers.py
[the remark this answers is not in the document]
Sumanth007: Agreed — deleted e2e/test_context_managers.py entirely.

python/copilot/client.py
Sumanth007: Removed logging entirely from ```__aexit__``` in both files. import logging also removed. Cleanup failures now surface as exceptions instead.

python/copilot/client.py
```diff
@@ -207,23 +213,45 @@ def __init__(self, options: CopilotClientOptions | None = None):
...
+            >>> async with CopilotClient() as client:
+            ...     session = await client.create_session()
+            ...     await session.send({"prompt": "Hello!"})
+        """
+        await self.start()
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: Optional[type[BaseException]],
```
brettcannon: Please use Python 3.11+ type annotation syntax and set default values, e.g.:

```suggestion
        exc_type: type[BaseException] | None  = None,
```

python/README.md
brettcannon, line 1: All examples should be updated to use context managers and the README the call-out should be you can do what the context manager does manually and how to do that.

python/copilot/session.py
```diff
@@ -649,20 +518,33 @@ async def disconnect(self) -> None:
...
-        with self._permission_handler_lock:
-            self._permission_handler = None
+            if self._destroyed:
+                return
+            self._destroyed = True
+
+        try:
+            await self._client.request("session.destroy", {"sessionId": self.session_id})
+        finally:
+            # Clear handlers even if the request fails
```
brettcannon:
```suggestion
            # Clear handlers even if the request fails.
```

python/copilot/client.py
Sumanth007: Done. Updated `__aexit__` signatures in both `CopilotClient` and `CopilotSession` to use `type[BaseException] | None = None` syntax. Also removed the `Optional` import since it's no longer used.

python/README.md
Sumanth007: Done. Restructured the entire README so all code examples use `async with` as the default pattern. Added a "Manual Resource Management" section under Quick Start that shows how to call `start()`/`stop()`/`disconnect()` directly, and linked to it from the API Reference section. Also replaced all `"gpt-5"` references with `"gpt-4o"`.

python/copilot/session.py
Sumanth007: Done.

python/README.md
```diff
@@ -28,14 +28,46 @@ import asyncio
...
+
+            session.on(on_event)
+
+            # Send a message and wait for completion
+            await session.send("What is 2+2?")
+            await done.wait()
+
+        # Session automatically disconnected here
+
+    # Client automatically stopped here
```
brettcannon: These sorts of comments are unnecessary.

```suggestion
```

python/README.md
```diff
@@ -199,10 +222,11 @@ async def lookup_issue(params: LookupIssueParams) -> str:
     issue = await fetch_issue(params.id)
     return issue.summary
 
-session = await client.create_session({
-    "model": "gpt-5",
+async with await client.create_session({
+    "model": "gpt-4o",
```
brettcannon: The various model changes are superfluous and unrelated to the PR.

python/test_client.py
```diff
@@ -490,3 +492,55 @@ async def mock_request(method, params):
...
+            mock_disconnect.assert_awaited_once()
+
+    @pytest.mark.asyncio
+    async def test_aexit_returns_none(self):
+        from copilot.session import CopilotSession
+
+        session = CopilotSession.__new__(CopilotSession)
+        with patch.object(session, "disconnect", new_callable=AsyncMock):
+            result = await session.__aexit__(None, None, None)
+            assert result is None
```
brettcannon: This is unnecessary.

python/test_client.py
```diff
@@ -490,3 +492,55 @@ async def mock_request(method, params):
...
+        with patch.object(client, "stop", new_callable=AsyncMock) as mock_stop:
+            await client.__aexit__(None, None, None)
+            mock_stop.assert_awaited_once()
+
+    @pytest.mark.asyncio
+    async def test_aexit_returns_none(self):
+        client = CopilotClient(SubprocessConfig(cli_path=CLI_PATH))
+        with patch.object(client, "stop", new_callable=AsyncMock):
+            result = await client.__aexit__(None, None, None)
+            assert result is None
```
brettcannon: This whole test is unnecessary.

python/test_client.py
Sumanth007: Done.

python/test_client.py
Sumanth007: Done.

PR #479, opened 2026-02-16
Fix ObjectDisposedException when disposing session after client.StopAsync()

### LINKED ISSUE #306: Disposing of a CopilotSession after stopping the client results in exception
Opus wrote code that explicitly issued the StopAsync, which is how I noticed. Presumably the StopAsync isn't actually necessary, and that the client ideally shouldn't be stopped while there are active sessions, but disposal also should ideally be tolerant to such things.

Repro:
```csharp
using GitHub.Copilot.SDK;

await using var session = await client.CreateSessionAsync();

await client.StopAsync();
```
results in this exception from the DisposeAsync from the await using cleanup:
```
Unhandled exception. System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'StreamJsonRpc.JsonRpc'.
   at Microsoft.Verify.NotDisposed(IDisposableObservable disposedValue, String message)
   at StreamJsonRpc.JsonRpc.InvokeCoreAsync[TResult](RequestId id, String targetName, IReadOnlyList`1 arguments, IReadOnlyList`1 positionalArgumentDeclaredTypes, IReadOnlyDictionary`2 namedArgumentDeclaredTypes, CancellationToken cancellationToken, Boolean isParameterObject)
   at GitHub.Copilot.SDK.CopilotClient.InvokeRpcAsync[T](JsonRpc rpc, String method, Object[] args, CancellationToken cancellationToken)
   at GitHub.Copilot.SDK.CopilotSession.DisposeAsync()
   at Program.<Main>$(String[] args)
```

SteveSandersonMS:
## Summary

Supersedes #371 (resolves merge conflicts against current main).
Fixes #306

When `await client.StopAsync()` is called before the `await using var session` goes out of scope, `session.DisposeAsync()` is called twice:
1. First by `StopAsync()` (while connection is alive) ✓
2. Second by `await using` cleanup (connection is now disposed) ✗ → `ObjectDisposedException`

## Solution

Make `CopilotSession.DisposeAsync()` idempotent and tolerant to already-disposed connections:

1. **Added `_isDisposed` flag** - Thread-safe using `Interlocked.Exchange`
2. **Made `DisposeAsync` idempotent** - Returns immediately if already disposed (required by `IAsyncDisposable` contract)
3. **Handle connection failures gracefully** - Catches `ObjectDisposedException` and `IOException` since dispose methods should never throw

This follows the same pattern already used in `CopilotClient.DisposeAsync()`.

## Changes
- `dotnet/src/Session.cs`: Add idempotency and exception handling to `DisposeAsync()`
- `dotnet/test/ClientTests.cs`: Add regression test

Credit to @parthtrivedi2492 for the original PR #371.

### DIFF
--- a/dotnet/src/Session.cs
+++ b/dotnet/src/Session.cs
@@ -54,6 +54,7 @@ public partial class CopilotSession : IAsyncDisposable
     private SessionHooks? _hooks;
     private readonly SemaphoreSlim _hooksLock = new(1, 1);
     private SessionRpc? _sessionRpc;
+    private int _isDisposed;
 
     /// <summary>
     /// Gets the unique identifier for this session.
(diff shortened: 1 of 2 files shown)


richlanpowers5-afk:
Fix n changed this from a dirty hoe to what it was pose to be bitch hot n
std and claim i give to her lol dirty slut

PR #519, opened 2026-02-19
Refactor CopilotClient initialization and enhance documentation

brettcannon:
Refactor the `CopilotClient` constructor to use keyword arguments for improved IDE support and type checking. Update the constructor to accept path-like objects and document changes in the README, including valid log levels and new parameters. Add overload signatures for better type checking and clarify nullable types in the documentation.

### DIFF
--- a/python/copilot/client.py
+++ b/python/copilot/client.py
@@ -109,30 +109,30 @@ class CopilotClient:
         self,
         *,
         cli_url: str,
-        cwd: Union[str, os.PathLike[str], None] = None,
-        port: int = 0,
-        log_level: LogLevel = "info",
-        auto_start: bool = True,
-        auto_restart: bool = True,
-        env: Optional[dict[str, str]] = None,
-    ) -> None: ...
+        cwd: Union[str, os.PathLike[str], None] = ...,
+        port: int = ...,
+        log_level: LogLevel = ...,
+        auto_start: bool = ...,
+        auto_restart: bool = ...,
+        env: Optional[dict[str, str]] = ...,
+    ) -> ...: ...
 
     @overload
     def __init__(
         self,
         *,
-        cli_path: Union[str, os.PathLike[str], None] = None,
-        cli_args: Optional[list[str]] = None,
-        cwd: Union[str, os.PathLike[str], None] = None,
-        port: int = 0,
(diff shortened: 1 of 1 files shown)


python/copilot/client.py
```diff
@@ -100,44 +101,106 @@ class CopilotClient:
...
+        cwd: Union[str, os.PathLike[str], None] = None,
+        port: int = 0,
+        use_stdio: Optional[bool] = None,
+        log_level: LogLevel = "info",
+        auto_start: bool = True,
+        auto_restart: bool = True,
+        github_token: Optional[str] = None,
+        use_logged_in_user: Optional[bool] = None,
+        env: Optional[dict[str, str]] = None,
+    ):
```
SteveSandersonMS, line 152: Managing all these overloads feels painful but if this is what you recommend as the most idiomatic Python solution then I'll gladly defer to you.

Do IDEs really not provide good hinting when passing an object literal in as an argument?

python/copilot/client.py
```diff
@@ -100,44 +101,106 @@ class CopilotClient:
...
+                or path-like objects. If not provided, uses the bundled
+                CLI binary.
+            cli_args: Additional command-line arguments to pass to the CLI
+                process.
+            cli_url: URL of an existing Copilot CLI server to connect to.
+                Format: "host:port", "http://host:port", or just "port".
+                Mutually exclusive with cli_path and use_stdio.
+            cwd: Working directory for the CLI process. Accepts strings
+                or path-like objects (default: current working directory).
+            port: Port for the CLI server in TCP mode (default: 0 for random).
```
SteveSandersonMS:
```suggestion
            port: Port for the CLI server in TCP mode (default: 0 for OS-assigned).
```

SteveSandersonMS (approved):
Looks good, though I think there are also a lot of samples to update under https://github.com/github/copilot-sdk/blob/main/test/scenarios

python/copilot/client.py
brettcannon: I don't quite follow what you mean by "good hinting when passing an object literal"? The overloads are there for type checkers because you have arguments that are valid only in certain situations (e.g. `cli_args` only applies when `cli_path` is provided).

PR #522, opened 2026-02-20
feat: Support overriding built-in tools

### LINKED ISSUE #411: Support overriding built-in tools
I'm working on a feature using the SDK where I would like to register my own tools and handle their execution externally (e.g. forwarding tool calls to another piece of software and returning the results back).

While digging into the SDK, I noticed that the `SessionOptions` in the SDK already has `externalToolDefinitions` and `externalToolDispatcher` fields that seem to do exactly this, where they get merged into the `tools` array alongside the built-in CLI and MCP tools and the dispatcher callback is called when the model invokes them.

Would it be possible to surface these two fields on the `public SessionConfig`? It would just be passing them through to the internal session, so no new logic is required since the all the plumbing is already there.

```ts
interface SessionConfig {
  // ... existing fields ...
  externalToolDefinitions?: ExternalToolDefinition[];
  externalToolDispatcher?: ExternalToolDispatcher;
}
```

patniko:
## Summary

Closes #411

When users register tools via `SessionConfig.tools` with names matching built-in CLI tools (e.g., `edit_file`, `read_file`), the CLI previously handled them itself instead of dispatching to the user's handler.

This PR makes user-registered tools automatically take precedence by adding their names to `excludedTools` in the RPC payload, so the CLI's built-in version is excluded and the SDK's local dispatch handles the call instead.

## Changes

Implemented consistently across all 4 SDKs:

| SDK | Code | Tests | Docs |
|-----|------|-------|------|
| **Node.js** | `mergeExcludedTools()` helper in `client.ts` | 4 tests in `client.test.ts` | README updated |
| **Python** | Inline merge in `client.py` | 4 tests in `test_client.py` | README updated |
| **Go** | `mergeExcludedTools()` helper in `client.go` | 4 tests in `client_test.go` | README updated |
| **.NET** | `MergeExcludedTools()` helper in `Client.cs` | 5 tests in `MergeExcludedToolsTests.cs` | README updated |

## How it works

1. Before sending `session.create` / `session.resume` RPC, collect tool names from user-registered tools
2. Merge them into `excludedTools` (with deduplication)
3. CLI receives the merged list and won't handle those tools itself
4. SDK dispatches tool calls to the user's local handler

## Backward compatibility

- Fully backward-compatible: users who don't name tools after built-ins see no behavior change
- If a user already explicitly lists a tool in `excludedTools`, deduplication avoids issues

### DIFF
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -415,6 +415,19 @@ var session = await client.CreateSessionAsync(new SessionConfig
 
 When Copilot invokes `lookup_issue`, the client automatically runs your handler and responds to the CLI. Handlers can return any JSON-serializable value (automatically wrapped), or a `ToolResultAIContent` wrapping a `ToolResultObject` for full control over result metadata.
 
+#### Overriding Built-in Tools
+
+If you register a tool with the same name as a built-in CLI tool (e.g. `edit_file`, `read_file`), your tool takes precedence. The SDK automatically adds the tool name to `ExcludedTools`, so the built-in is disabled and your handler is called instead. This is useful when you need custom behavior for built-in operations.
+
+```csharp
+AIFunctionFactory.Create(
+    async ([Description("File path")] string path, [Description("New content")] string content) => {
+        // your logic
+    },
+    "edit_file",
+    "Custom file editor with project-specific validation")
+```
+
 ### System Message Customization
 
 Control the system prompt using `SystemMessage` in session config:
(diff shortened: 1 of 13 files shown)


theblixguy:
Would this work if my tools didn't have the same name, but performed the same functionality? (e.g. instead of my tool being called "update_todo" it was called "TodoWrite"). Basically, if I have my own tools that can do the same job, I would like to use that over the CLI's built in tools.

PR #546, opened 2026-02-23
fix(nodejs): add CJS compatibility for VS Code extensions

### LINKED ISSUE #528: Node SDK: getBundledCliPath() breaks in CJS bundles (VS Code extensions)
## Summary

`getBundledCliPath()` in `nodejs/src/client.ts` uses `import.meta.resolve("@github/copilot/sdk")`, which is an ESM-only API. VS Code extensions must bundle to CJS format using esbuild, which replaces `import.meta` with `{}` in CJS output. This causes a runtime crash when the SDK tries to locate the CLI.

## Affected Versions

- 0.1.23, 0.1.24, 0.1.25, 0.1.26-preview.0 (all contain the same `getBundledCliPath()` implementation)
- 0.1.22 and earlier are not affected — they used `cliPath: options.cliPath || "copilot"` (simple PATH lookup)

## Reproduction

1. Create a VS Code extension that uses `@github/copilot-sdk`
2. Bundle with esbuild in CJS format (`format: 'cjs'`)
3. Package as VSIX and install
4. Attempt to start a session

**Error 1** (without any polyfill):
```
TypeError: Zo.resolve is not a function
```

esbuild replaces `import.meta` with `{}`, so `import.meta.resolve` becomes `{}.resolve` → undefined.

**Error 2** (with `import.meta.resolve` polyfilled to `require.resolve`):
```
Error: Cannot find module '@github/copilot/sdk'
```

`require.resolve("@github/copilot/sdk")` fails because `@github/copilot` only exports `./sdk` under the ESM `import` condition, not `require`. Additionally, the VSIX does not include `node_modules`.

## Root Cause

The "Bundling (#382)" PR changed the default CLI resolution from a simple PATH lookup to:

```typescript
function getBundledCliPath(): string {
    const sdkUrl = import.meta.resolve("@github/copilot/sdk");
    const sdkPath = fileURLToPath(sdkUrl);
    return join(dirname(dirname(sdkPath)), "index.js");
}
```

This function uses two ESM-only features:
1. `import.meta.resolve()` — not available in CJS
2. The `@github/copilot` package's `./sdk` export — only exposed under ESM `import` condition

## Workaround

Pass an explicit `cliPath` to the `CopilotClient` constructor, which short-circuits `getBundledCliPath()` via `options.cliPath || getBundledCliPath()`:

```typescript
import { execFileSync } from 'child_process';

function resolveCliFromPath(): string {
    const cmd = process.platform === 'win32' ? 'where' : 'which';
    return execFileSync(cmd, ['copilot'], { encoding: 'utf-8' }).trim().split(/\r?\n/)[0];
}

const client = new CopilotClient({
    cliPath: resolveCliFromPath(),
    // ...
});
```

This restores the pre-0.1.23 behavior of resolving `copilot` from PATH.

## Suggestion

The unreleased commit `736b17e` ("Remove @github/copilot dependency, point all SDKs at copilot-core") changes `getBundledCliPath()` to `return "copilot-core"`, which would resolve this issue. However, SDK 0.1.23+ also added an `existsSync(cliPath)` check that rejects bare command names. If the intent is to support PATH-based resolution, the `existsSync` guard would need to handle bare command names (or use `which`/`where` internally).

Alternatively, a simpler fix would be to catch `getBundledCliPath()` failures and fall back to `"copilot"` on PATH, similar to the 0.1.22 behavior:

```typescript
cliPath: options.cliPath || tryGetBundledCliPath() || "copilot"
```

## Environment

- Node.js 24
- esbuild 0.25.x (CJS output format)
- VS Code extension host (CJS module system)
- Linux x86_64 (also affects Windows/macOS — any CJS environment)

darthmolen:
## Summary

Fixes #528 — `getBundledCliPath()` uses `import.meta.resolve("@github/copilot/sdk")` which throws in CJS contexts (VS Code extensions bundled with esbuild `format:"cjs"`).

**Changes:**
- Replace `import.meta.resolve` with `createRequire` + module resolution path walking (works in both ESM and CJS)
- Add bundled CJS output (`dist/cjs/index.cjs`) via esbuild
- Add conditional `exports` in package.json (`import` + `require`)
- Add CJS compatibility test suite (4 tests)

## Problem

The `@github/copilot` package only exposes an ESM `"import"` condition for its `./sdk` subpath export — no `"require"` condition. This means:

1. `import.meta.resolve("@github/copilot/sdk")` — works in ESM only
2. `require.resolve("@github/copilot/sdk")` — fails (`Package subpath './sdk' is not defined by "exports"`)
3. `require.resolve("@github/copilot/package.json")` — also fails (not exported)

When a consumer bundles the SDK with esbuild `format:"cjs"`, `import.meta` becomes an empty object, and `import.meta.resolve` is undefined.

## Solution

Instead of resolving through the `@github/copilot` package's strict exports, we:

1. Use `createRequire(import.meta.url ?? pathToFileURL(__filename).href)` — `import.meta.url` in ESM, `__filename` fallback in CJS
2. Walk `require.resolve.paths("@github/copilot")` to get the Node module search directories
3. Check each path for `@github/copilot/index.js` using `existsSync`

This avoids the ESM-only exports entirely and works in both module systems.

## Test plan

- [x] Existing unit tests pass (19/19 passing, 3 pre-existing CLI spawn failures unchanged)
- [x] New CJS tests pass (4/4):
  - CJS dist file exists
  - ESM dist file exists  
  - CJS build is requireable and exports `CopilotClient`
  - `CopilotClient` constructor resolves bundled CLI path in CJS context
- [ ] Manual: bundle SDK in a VS Code extension with esbuild `format:"cjs"`, verify `getBundledCliPath()` works

### DIFF
--- a/nodejs/esbuild-copilotsdk-nodejs.ts
+++ b/nodejs/esbuild-copilotsdk-nodejs.ts
@@ -26,6 +26,7 @@ await esbuild.build({
     target: "es2022",
     sourcemap: false,
     outExtension: { ".js": ".js" },
+    logOverride: { "empty-import-meta": "silent" },
 });
 
 // Mark the CJS directory so Node treats .js files as CommonJS


patniko (dismissed):
## Review: Checking Steve Sanderson's feedback was addressed

### Steve's feedback (2 comments):

**1. "Do you have a reason to think we need to support both [CJS and ESM] and not just CJS?"**
> Steve wanted a single build output, not dual CJS+ESM builds.

**✅ Addressed.** The contributor removed the dual build approach entirely — no changes to `esbuild-copilotsdk-nodejs.ts` or `package.json`. The PR now keeps the single ESM build and only modifies `getBundledCliPath()` to handle CJS-shimmed environments via fallback logic. The contributor confirmed this in their [follow-up comment](https://github.com/github/copilot-sdk/pull/546#issuecomment-3948261723).

**2. "Perhaps the best outcome altogether would be having only a single ESM build, but putting in place whatever fallback logic is needed to get `__filename`-based path resolution to work if you are loading in a shimmed CJS environment."**

**✅ Addressed.** This is exactly what the current code does:
```typescript
const require = createRequire(import.meta.url ?? pathToFileURL(__filename).href);
```
`import.meta.url` works in native ESM; `__filename` via `pathToFileURL` works in CJS-shimmed contexts.

### Copilot Reviewer feedback (3 threads, all unresolved):

**1. Test comment about `@github/copilot` not being installed** — suggestion to clarify the test assumes the package is in node_modules. **Partially addressed** — the test was rewritten but doesn't include the suggested NOTE comment. Minor.

**2. Error message for empty search paths** — suggestion to add a descriptive error when `require.resolve.paths()` returns null/empty. **✅ Addressed.** The code handles `?? []` and the error message includes: `Searched ${searchPaths.length} paths. Ensure it is installed.`

**3. Document the requirement that `@github/copilot` must be in node_modules** — suggestion to add README documentation. **❌ Not addressed.** No documentation changes were included.

---

### 🔴 Critical issue: PR conflicts with `onPermissionRequest` requirement

This PR is based on `main` from before PR #554 (`Require permission handler on session creation`, merged Feb 24). The current diff **removes the `onPermissionRequest` requirement** from both `createSession()` and `resumeSession()` — making the parameter optional and deleting the validation that throws if it's missing.

This is a regression against the intentional security decision in #554. The PR needs to be rebased on current `main` and these unrelated `onPermissionRequest` changes must be dropped, keeping only the `getBundledCliPath()` fix and the new test file.

### Summary

Steve's two concerns are fully addressed — single ESM build with CJS fallback logic ✅. But the PR needs a rebase to avoid regressing the `onPermissionRequest` requirement from #554.

darthmolen:
Hi @patniko @SteveSandersonMS — just a gentle ping. The PR has been reworked per Steve's feedback (single ESM build, `__filename`-fallback for shimmed CJS — no changes to the build config or package.json). @hacetin asked about the status last week too.

Is there anything blocking a merge? Happy to make further changes if needed.

PR #561, opened 2026-02-24
chore(python): Bump the oldest supported Python version to 3.11

brettcannon:
Python 3.10 gives us nicer type hint syntax and 3.11 gives us exception groups which will be used in the future. Along the way, expand the test matrix for Python versions in CI.

### DIFF
--- a/go/generated_session_events.go
+++ b/go/generated_session_events.go
@@ -35,120 +35,120 @@ type SessionEvent struct {
 }
 
 type Data struct {
-	Context                                              *ContextUnion            `json:"context"`
-	CopilotVersion                                       *string                  `json:"copilotVersion,omitempty"`
-	Producer                                             *string                  `json:"producer,omitempty"`
-	SelectedModel                                        *string                  `json:"selectedModel,omitempty"`
-	SessionID                                            *string                  `json:"sessionId,omitempty"`
-	StartTime                                            *time.Time               `json:"startTime,omitempty"`
-	Version                                              *float64                 `json:"version,omitempty"`
-	EventCount                                           *float64                 `json:"eventCount,omitempty"`
-	ResumeTime                                           *time.Time               `json:"resumeTime,omitempty"`
-	ErrorType                                            *string                  `json:"errorType,omitempty"`
-	Message                                              *string                  `json:"message,omitempty"`
-	ProviderCallID                                       *string                  `json:"providerCallId,omitempty"`
-	Stack                                                *string                  `json:"stack,omitempty"`
-	StatusCode                                           *int64                   `json:"statusCode,omitempty"`
-	Title                                                *string                  `json:"title,omitempty"`
-	InfoType                                             *string                  `json:"infoType,omitempty"`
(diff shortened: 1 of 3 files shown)


.github/workflows/python-sdk-tests.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.11", "3.14", "3.14t"]
```
SteveSandersonMS: Is it really important to test across all these? We end up with 9 different CI runs for Python if we do (3 versions * 3 OSes). It makes it harder to track down what's going on if there's a failure as it will tend to be repeated 3 or 9 times.

For other languages we test on a single representative version. Can we do that for Python?

python/copilot/generated/rpc.py
```diff
@@ -10,7 +10,8 @@
 
 
 from dataclasses import dataclass
-from typing import Any, Optional, List, Dict, TypeVar, Type, cast, Callable
+from typing import Any, Optional, List, Dict, TypeVar, Type, cast
+from collections.abc import Callable
```
SteveSandersonMS, line 14: It seems wrong that the codegenerated code (rpc.py and session_events.py) have changed given that this PR didn't change the code generator.

If this file was updated manually or through some automated process, can you revert it and only update this file by running the code generator?

SteveSandersonMS: Actually it looks as if CI has detected this change is out of sync with the generator and failed the *Verify generated files are up-to-date* job

.github/workflows/python-sdk-tests.yml
brettcannon: > Is it really important to test across all these?

It depends on your definition of "important". If you test a single version you have to decide if you want to test against the oldest version to make sure you don't break compatibility, or the newest so you don't use stuff that's been removed. Do you have a feeling as where problems have historically shown up in PRs?

python/copilot/generated/rpc.py
brettcannon: > If this file was updated manually or through some automated process,

It was automated.



> update this file by running the code generator?

I have an update, but unfortunately the npm package being used for Python is woefully out of date, so Copilot had to add a post-process step.

.github/workflows/python-sdk-tests.yml
SteveSandersonMS: So far I don't recall seeing any Python version-specific problems. Is it reasonable to test against the oldest version we claim to support, and rely on the Python maintainers not to break backcompat? That's more or less the approach we have for .NET and Node at least. At an extreme, we're not going to test against every individual patch release (as a set that expands forever) so we have to accept a compromise.

If we have reason to think there's a breaking change in a newer version of the platform we can account for that in a more targeted way.

brettcannon: I changed it to be just 3.11.

PR #583, opened 2026-02-25
Move client creation to functions

brettcannon:
The `cli_client()` and `network_client()` functions separate out the two ways to launch the client to simplify things. The initializer for `CopilotClient` is made private by making the config objects it accepts private to keep the public API simple.

### DIFF
--- a/docs/auth/index.md
+++ b/docs/auth/index.md
@@ -255,7 +255,7 @@ const client = new CopilotClient({
 
 <!-- docs-validate: skip -->
 ```python
-client = copilot.cli_client()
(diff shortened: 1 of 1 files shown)


brettcannon:
> What would be the payoff from doing this split?

Simpler parameters based on needs and what Python folks are more used to. The config object approach is just not common in the Python community (hence why the builder pattern isn't either) as Python's signature support makes building up arguments isn't as necessary.

But I understand if you're concerned about support overhead. Let me think about whether config objects or everything as parameters but without overloads makes the most sense.

PR #586, opened 2026-02-25
Fix Go client data races

chlowell:
The client's `running bool` field is accessed concurrently by `Stop()` and `readLoop()` methods without synchronization, creating a data race. This can be prevented by simply changing the field's type to `atomic.Bool`. I also enabled the `go test` race detector in the test script so future CI runs will fail when the test suite triggers a race.

### DIFF
--- a/go/client.go
+++ b/go/client.go
@@ -1152,7 +1152,10 @@ func (c *Client) startCLIServer(ctx context.Context) error {
 			select {
 			case <-timeout:
 				killErr := c.killProcess()
-				return errors.Join(fmt.Errorf("timeout waiting for CLI server to start"), killErr)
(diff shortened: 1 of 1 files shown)


chlowell:
Sorry I missed that one; I haven't been able to repro it on my machine with the current test suite. I added a test that reliably triggers it and other races in `copilot.Client`, which I've now fixed. I also fixed some process leaks that came up while I was working on synchronization. The new synchronization is a bit involved because some data shared by `Start` and `[Force]Stop` is already synchronized and a simple mutex covering all three methods won't work because `ForceStop` needs to function concurrently with `Start` and `Stop`. I was aiming for a smallish fix; let me know what you think.

PR #599, opened 2026-02-27
docs: clarify session destroy vs delete semantics

### LINKED ISSUE #526: Clarify session disposal
The API and documentation around session disposal are counterintuitive and contradictory such that it isn't clear how to go about this:
 
- [Session.Destroy()](https://github.com/github/copilot-sdk/blob/c4b3b366c4bd8dfba9ba4aa05e4019825360ad78/go/session.go#L528) docs say it "destroys this session and releases all associated resources" and that the session can no longer be used after calling this method. So far so good: "destroy" implies an irreversible operation leaving nothing behind. However, the docs go on to contradict this by saying that the conversation can be resumed by calling `Client.ResumeSession()`. Maybe "destroy" is the wrong verb here; how about "suspend" or "close"?
- [Client.DeleteSession()](https://github.com/github/copilot-sdk/blob/c4b3b366c4bd8dfba9ba4aa05e4019825360ad78/go/client.go#L675) naively appears equivalent to `Session.Destroy()` but apparently it's a superset: it does everything `Destroy` does plus it deletes files. It isn't clear how I should decide whether to "delete" or "destroy" a session.
- [Client.Stop()](https://github.com/github/copilot-sdk/blob/c4b3b366c4bd8dfba9ba4aa05e4019825360ad78/go/client.go#L297) is documented as both "closing" and "destroying" all active sessions. This is confusing because "destroy" implies an irreversible operation leaving nothing behind while "close" is ambiguous (what happens to session resources on closure?). Because this method calls `Session.Destroy()` on each active session, it apparently leaves their files behind. Maybe this is again suggesting "close" as a better verb than "destroy". In any case, the docs should at least call out that `Client.Stop()` leaves session artifacts behind and SDK users who want to remove them must explicitly delete sessions.

patniko:
## Summary

Clarifies the distinction between `destroy()` and `deleteSession()` across all four SDK languages and the session persistence guide.

Fixes #526

## Problem

As reported in #526, the API documentation around session disposal is counterintuitive:

- **`destroy()`** says it "destroys this session and releases all associated resources" but then says the session can be resumed — contradicting what "destroy" implies
- **`deleteSession()`** is the truly permanent operation, but there's no clear guidance on when to use one vs the other
- **`stop()`** says it "destroys all active sessions" but actually leaves session data on disk, which isn't mentioned

## Changes

**All four SDK languages** (Go, Node.js, Python, .NET) — updated doc comments for:
- `destroy()` / `DisposeAsync()`: Now explicitly states it releases **in-memory** resources only, preserves disk state, and points to `deleteSession()` for permanent removal
- `stop()` / `StopAsync()`: Clarifies it closes sessions (not "destroys") and notes that disk data is preserved — users wanting permanent removal must call `deleteSession()` first
- `deleteSession()` / `DeleteSessionAsync()`: Now explicitly contrasts with `destroy()` to make the difference clear

**Session persistence guide** (`docs/guides/session-persistence.md`):
- Renamed "Explicit Session Destruction" to "Closing a Session" and "Permanently Deleting a Session" — two separate subsections
- Added a callout box summarizing `destroy()` vs `deleteSession()`
- Updated the summary table with clarifying descriptions

### DIFF
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -217,22 +217,18 @@ Abort the currently processing message in this session.
 
 Get all events/messages from this session.
 
-##### `DisconnectAsync(CancellationToken cancellationToken = default): Task`
-
-Disconnect the session and release in-memory resources. Session data on disk is preserved — the conversation can be resumed later via `ResumeSessionAsync()`. To permanently delete session data, use `client.DeleteSessionAsync()`.
-
 ##### `DisposeAsync(): ValueTask`
 
-Calls `DisconnectAsync()`. Enables the `await using` pattern for automatic cleanup:
+Close the session and release in-memory resources. Session data on disk is preserved — the conversation can be resumed later via `ResumeSessionAsync()`. To permanently delete session data, use `client.DeleteSessionAsync()`.
 
 ```csharp
 // Preferred: automatic cleanup via await using
 await using var session = await client.CreateSessionAsync(config);
-// session is automatically disconnected when leaving scope
+// session is automatically disposed when leaving scope
 
-// Alternative: explicit disconnect
+// Alternative: explicit dispose
 var session2 = await client.CreateSessionAsync(config);
-await session2.DisconnectAsync();
+await session2.DisposeAsync();
 ```
 
 ---
--- a/dotnet/src/Client.cs
+++ b/dotnet/src/Client.cs
@@ -243,11 +243,11 @@ public sealed partial class CopilotClient : IDisposable, IAsyncDisposable
         {
             try
             {
-                await session.DisconnectAsync();
+                await session.DisposeAsync();
             }
             catch (Exception ex)
             {
-                errors.Add(new Exception($"Failed to disconnect session {session.SessionId}: {ex.Message}", ex));
+                errors.Add(new Exception($"Failed to dispose session {session.SessionId}: {ex.Message}", ex));
             }
         }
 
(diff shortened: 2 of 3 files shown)


patniko:
@SteveSandersonMS would be great to chat about this

dotnet/README.md
```diff
@@ -217,9 +217,23 @@ Abort the currently processing message in this session.
...
+Calls `DisconnectAsync()`. Enables the `await using` pattern for automatic cleanup:
+
+```csharp
+// Preferred: automatic cleanup via await using
+await using var session = await client.CreateSessionAsync(config);
+// session is automatically disconnected when leaving scope
+
+// Alternative: explicit disconnect
+var session2 = await client.CreateSessionAsync(config);
+await session2.DisconnectAsync();
```
SteveSandersonMS: For .NET we could make the case that we don't need both DisconnectAsync and DisposeAsync since they are exactly equivalent and DisposeAsync is sufficient on its own.

I don't feel strongly about this so if you prefer to keep DisconnectAsync that's fine, but if you don't have opinions I'd lean towards just having DisposeAsync as the more idiomatic standard.

stephentoub: In https://github.com/github/copilot-sdk/pull/682 I propose adding a ShutdownAsync as a mechanism for developers that want to be able to receive shutdown events to get them; you can just use DisposeAsync, but if you want to reliably get shutdown events, you use ShutdownAsync, then wait for the shutdown events, and then DisposeAsync.

I suggest we shouldn't have _three_ "I'm done" APIs. If we like the "disconnect" terminology, could "disconnect" be used instead of what I'm calling "shutdown"?

PR #602, opened 2026-02-27
Add OpenTelemetry GenAI instrumentation to Copilot SDK

stephentoub:
Introduced built-in OpenTelemetry tracing and metrics for agent turns and tool executions, following GenAI semantic conventions (v1.40).

image

image

image

image

image


stephentoub:
I misunderstood a few things about the session events object model. Reworking this a bit, so flipped it back to draft.

dotnet/src/OpenTelemetryConsts.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        public const string Address = "server.address";
+        public const string Port = "server.port";
+    }
+
+    public static class GenAI
+    {
+        public const string Chat = "chat";
+        public const string InvokeAgent = "invoke_agent";
+        public const string ExecuteTool = "execute_tool";
```
MackinnonBuck: Nit: I wonder if the purpose of these would be slightly clearer they were moved to a separate `OperationNames` class. So it would be e.g., `GenAI.OperationNames.Chat` instead of `GenAI.Chat`.

dotnet/src/CopilotTelemetry.cs
```diff
@@ -0,0 +1,1958 @@
...
+                if (!string.IsNullOrWhiteSpace(_systemInstructionsJson))
+                {
+                    activity.SetTag(OpenTelemetryConsts.GenAI.SystemInstructions, _systemInstructionsJson);
+                }
+            }
+
+            if (!string.IsNullOrWhiteSpace(_toolDefinitionsJson))
+            {
+                activity.SetTag(OpenTelemetryConsts.GenAI.Tool.Definitions, _toolDefinitionsJson);
+            }
```
MackinnonBuck: The [Semantic Conventions Doc](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#:~:text=Since%20this%20attribute%20could%20be%20large%2C%20it%E2%80%99s%20NOT%20RECOMMENDED%20to%20populate%20it%20by%20default.%20Instrumentations%20MAY%20provide%20a%20way%20to%20enable%20populating%20this%20attribute.) for `gen_ai.tool.definitions` says that this attribute should be opt-in because it could be large. Should we consider adding a property to configure this?

dotnet/src/CopilotTelemetry.cs
```diff
@@ -0,0 +1,1958 @@
...
+                _lastOutputChunkElapsed = elapsed;
+                _telemetry.RecordTimeToFirstChunk(
+                    elapsed.TotalSeconds,
+                    _requestModel,
+                    null, // response model not yet known during streaming
+                    ProviderName,
+                    ServerAddress,
+                    ServerPort);
+                return;
+            }
```
MackinnonBuck: The `_turnTimestamp` field gets set by `BeginChatTurnLocked()`, which gets called _after_ `RecordOutputChunkMetricLocked()` (this method) from `ProcessEvent()`. Wouldn't this cause the "time to first chunk" metric to be incorrect, as the `elapsed` value is getting computed from an uninitialized `_turnTimestamp`?

I wonder if we should limit `RecordOutputChunkMetricLocked()` to only be called for events that represent actual content. This would exclude e.g., `assistant.turn_start`, which is what's causing the issue described above.

dotnet/src/OpenTelemetryConsts.cs
stephentoub: Sounds good.

dotnet/src/CopilotTelemetry.cs
stephentoub: We don't currently have such a knob in M.E.AI, and we haven't been asked for one (yet). I'd err on the side of not exposing configuration for it yet. We could later add public API for it, or use env vars.

dotnet/src/CopilotTelemetry.cs
stephentoub: I'll take a look at the timing issue. With regards to "events that represent actual content", I have a question in to otel about that: https://github.com/open-telemetry/semantic-conventions/issues/3483

python/copilot/_copilot_telemetry.py
```diff
@@ -0,0 +1,1744 @@
...
+    def _process_subagent_event(self, subagent: _SubagentState, event: SessionEvent) -> None:
+        """Routes an event to its owning subagent's spans."""
+        if event.type == SessionEventType.ASSISTANT_USAGE:
+            data = event.data
+            model = getattr(data, "model", None)
+            subagent.response_model = model
+
+            if model:
+                if subagent.chat_span is not None:
+                    subagent.chat_span.set_attribute(ATTR_GEN_AI_RESPONSE_MODEL, model)
```
MackinnonBuck, line 1531: If a usage event doesn't contain a `model`, would we want to preserve the existing value?

```suggestion
            model = getattr(data, "model", None)
            if model:
                subagent.response_model = model
                if subagent.chat_span is not None:
                    subagent.chat_span.set_attribute(ATTR_GEN_AI_RESPONSE_MODEL, model)
```

stephentoub:
@tonybaloney, could you review this as well, and in particular the python implementation?

PR #606, opened 2026-02-27
Go: include CLI subprocess stderr output in error messages

chlowell:
When the CLI subprocess fails (crash, bad args, etc.), the Go SDK returns a generic error that doesn't explain why the CLI exited. With this PR, the SDK appends the subprocess's stderr to these errors:
```diff
 process exited unexpectedly
 failed to kill CLI process: os: process already finished
+stderr: something went wrong
```

The implementation is a little complex because of the synchronization required to prevent races and leaks, and enable a client instance to start a new subprocess while concurrently cleaning up after an old one. The high-level gist is that the client uses a ring buffer (capped at 64 KB) to track stderr and drains that into an error message when the subprocess exits.

### DIFF
--- a/go/client.go
+++ b/go/client.go
@@ -34,6 +34,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


SteveSandersonMS:
Closing as stale - will leave this for @qmuntal to continue in the future if desired.

PR #611, opened 2026-02-28
fix:  Justfile doesn't install everything necessary to run the test suite

### LINKED ISSUE #584: Justfile doesn't install everything necessary to run the test suite
`test/harness` isn't installed like it is in CI:

https://github.com/github/copilot-sdk/blob/f9144f1e8cf90b805580413112d1d39a61d9bae5/.github/workflows/python-sdk-tests.yml#L77-L79

vivganes:
Fixes #584

### DIFF
--- a/justfile
+++ b/justfile
@@ -78,6 +78,7 @@ install:
     @cd python && uv pip install -e ".[dev]"
     @cd go && go mod download
     @cd dotnet && dotnet restore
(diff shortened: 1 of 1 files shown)


kellyerinmontgomery426-ui:
If it doesn't run right then that means there's a decimal or a jigger Jagger or numeral or something or abbreviations is not in the right place

PR #636, opened 2026-03-03
feat: Support overriding built-in tools

### LINKED ISSUE #411: Support overriding built-in tools

SteveSandersonMS:
## Summary

Adds support for SDK-registered tools to override CLI built-in tools of the same name (Issue #411).

Each tool can now declare `overridesBuiltInTool: true` to intentionally replace a built-in tool. The runtime uses this flag to detect accidental name clashes (error) vs intentional overrides.

## Changes across all 4 SDKs (Node.js, Python, Go, .NET)

- Add `overridesBuiltInTool` field to tool definitions
- Pass `overridesBuiltInTool` through wire protocol to runtime
- Update documentation with 'Overriding Built-in Tools' section in all READMEs
- Add unit tests and E2E tests with snapshot scenarios

Closes #411


dotnet/README.md
```diff
@@ -415,6 +415,27 @@ var session = await client.CreateSessionAsync(new SessionConfig
 
 When Copilot invokes `lookup_issue`, the client automatically runs your handler and responds to the CLI. Handlers can return any JSON-serializable value (automatically wrapped), or a `ToolResultAIContent` wrapping a `ToolResultObject` for full control over result metadata.
 
+#### Overriding Built-in Tools
+
+If you register a tool with the same name as a built-in CLI tool (e.g. `edit_file`, `read_file`), the SDK will throw an error unless you explicitly opt in by setting `BuiltInToolOverrides` on the session config. This flag signals that you intend to replace the built-in tool with your custom implementation.
```
stephentoub: Is there a way to query for all built-in tools?

SteveSandersonMS: Not currently. Can you say what the use case would be?

stephentoub: I'm wondering if there's a) a way I can proactively determine whether I'm going to get conflicts with built-in tools or if it'll just be "oops, that one conflicts now, better change it" and b) whether I can determine what tools are in use to see whether I might want to replace one with my own logic.

SteveSandersonMS: OK thanks for explaining. I suspect even then it might be tricky to make a programmatic decision since it would depend on understanding the semantics and usage patterns of the built-in tool. But we can certainly consider adding a way to enumerate them if we get demand for it.

Issue #490, opened 2026-02-17
How to receive the session.shutdown event?


theblixguy:
I'm trying to access `totalPremiumRequests` and other data from the `session.shutdown` event but never see it fire. I log every event type from `session.on()` and get `assistant.usage`, `session.idle`, etc but never `session.shutdown`.

My sessions are long-lived and reused, so `destroy()` isn't called during normal use. On process exit I call `client.stop()` (which calls `destroy()` internally) but still nothing shows up. Looking at the [source](https://github.com/github/copilot-sdk/blob/5016587a62652f3d184b3c6958dfc63359921aa8/nodejs/src/session.ts#L517), it looks like `destroy()` clears all event handlers right after the RPC resolves, which might be before the shutdown notification arrives.

What's the intended way to get this data?

Stono:
ahhh I was just about to raise the same problem - I need the shutdown event! 
At the moment there is no way that I can see to end a session and get the session usage metrics.

jackcan2:
Any progress on this?  I could use the shutdown event as well.  I'm currently evaluating token usage for a variety of agent POCs. In the mean time I can get from the event logs if needed, but it would be nice to collect the stats with an event handler.

kaawmnain24-netizen:
- [ ] @ > . <

PR #664, opened 2026-03-04
Register sessions before RPC to prevent dropped events

stephentoub:
## Summary

Register sessions in the client's sessions map **before** issuing the \session.create\ and \session.resume\ RPC calls, so that events emitted by the CLI during the RPC (e.g. \session.start\, permission requests, tool calls) are not dropped. Previously, sessions were registered only after the RPC completed, creating a window where notifications for the session had no target.

## Changes

Across all four SDKs (Node.js, Python, Go, .NET):

- **Generate \sessionId\ client-side** (via UUID) rather than extracting it from the server response
- **Create and register the session** in the sessions map before the RPC
- **Set \workspacePath\** from the RPC response after it completes
- **Clean up the session** from the map if the RPC fails

## Tests

12 new unit tests across Node.js, Python, and Go:

| Test | Node.js | Python | Go |
|------|---------|--------|----|
| Session in map during \session.create\ RPC | ✅ | ✅ | ✅ |
| Session in map during \session.resume\ RPC | ✅ | ✅ | ✅ |
| Cleanup on \session.create\ RPC failure | ✅ | ✅ | ✅ |
| Cleanup on \session.resume\ RPC failure | ✅ | ✅ | ✅ |

Go tests use a fake JSON-RPC server via \io.Pipe()\ to verify map state during the RPC without needing a real CLI process. Node.js and Python tests mock the RPC layer and assert the session is present in the sessions map when the RPC is invoked.

### DIFF
--- a/dotnet/src/Client.cs
+++ b/dotnet/src/Client.cs
@@ -403,34 +403,11 @@ public sealed partial class CopilotClient : IDisposable, IAsyncDisposable
             config.Hooks.OnSessionEnd != null ||
             config.Hooks.OnErrorOccurred != null);
 
-        var request = new CreateSessionRequest(
-            config.Model,
-            config.SessionId,
-            config.ClientName,
-            config.ReasoningEffort,
-            config.Tools?.Select(ToolDefinition.FromAIFunction).ToList(),
-            config.SystemMessage,
-            config.AvailableTools,
-            config.ExcludedTools,
-            config.Provider,
-            (bool?)true,
-            config.OnUserInputRequest != null ? true : null,
-            hasHooks ? true : null,
-            config.WorkingDirectory,
-            config.Streaming is true ? true : null,
-            config.McpServers,
-            "direct",
-            config.CustomAgents,
-            config.Agent,
-            config.ConfigDir,
-            config.SkillDirectories,
-            config.DisabledSkills,
-            config.InfiniteSessions);
+        var sessionId = config.SessionId ?? Guid.NewGuid().ToString();
 
-        var response = await InvokeRpcAsync<CreateSessionResponse>(
-            connection.Rpc, "session.create", [request], cancellationToken);
-
-        var session = new CopilotSession(response.SessionId, connection.Rpc, response.WorkspacePath);
+        // Create and register the session before issuing the RPC so that
+        // events emitted by the CLI (e.g. session.start) are not dropped.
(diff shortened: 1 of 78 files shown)


stephentoub:
> @stephentoub Excellent, thanks. Will review shortly. Interested in whether this might also address #490

Hmm... I don't think it will. We can look at that separately.

go/client_test.go
```diff
@@ -569,3 +574,230 @@ func TestClient_StartStopRace(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// fakeJSONRPCServer reads one JSON-RPC request from r and sends a response to w.
+// onRequest is called with the parsed method and params before the response is sent,
+// allowing the caller to inspect state (e.g. the sessions map) during the RPC.
+func fakeJSONRPCServer(t *testing.T, r io.Reader, w io.WriteCloser, onRequest func(method string, params json.RawMessage)) {
```
SteveSandersonMS: Instead of adding new infrastructure for a mock JSON-RPC server, could we cover this functionality in E2E tests? It seems like an E2E test would be ideal as it would directly show what events the client receives.

Also if we're adding new tests for Go/Node/Python should we also be adding them for .NET?

stephentoub: > Also if we're adding new tests for Go/Node/Python should we also be adding them for .NET?

I couldn't figure out a good way to do this with unit tests for .NET, as there wasn't anything mockable.

I can try to switch all of them to E2E tests instead.

SteveSandersonMS:
@stephentoub The implementation looks great! Thanks very much for this. Only change I'm hoping for is doing the test coverage for all four languages in the existing E2E tests as it seems like it should be trivial to do that there, and will eliminate a lot of mocking and the fake JSON-RPC server.

go/client_test.go
SteveSandersonMS: It should be much easier as E2E but if Copilot can't figure it out first time, LMK and I can handle it tomorrow.

stephentoub:
> > @stephentoub Excellent, thanks. Will review shortly. Interested in whether this might also address #490
> 
> Hmm... I don't think it will. We can look at that separately.

https://github.com/github/copilot-sdk/pull/682 addresses [shutdown](https://github.com/github/copilot-sdk/issues/490).

SteveSandersonMS (approved):
Fantastic - thanks!

PR #783, opened 2026-03-10
fix(dotnet): add AOT-safe RequestId serialization for StreamJsonRpc

PureWeen:
## Problem

When `CancellationToken` fires during a JSON-RPC operation, StreamJsonRpc's `StandardCancellationStrategy` sends a `$/cancelRequest` notification containing a `RequestId`. The SDK's AOT serialization setup (source-generated contexts + `MakeReadOnly()`) has no type info for `RequestId`, causing a `NotSupportedException` that silently kills the JSON-RPC connection.

Fixes: https://github.com/PureWeen/PolyPilot/issues/319

## Root Cause

Commit 4e7319e locked serialization to source-generated contexts only. `RequestId` couldn't be added via `[JsonSerializable]` because it has `[JsonConverter(typeof(RequestIdSTJsonConverter))]` and that converter is `internal` to StreamJsonRpc (SYSLIB1220). The converter also can't be used directly (CS0122).

## Solution

Add AOT-safe `RequestIdTypeInfoResolver` + `RequestIdJsonConverter`:

- **`RequestIdJsonConverter`** — hand-written converter using `RequestId`'s public API (`Number`, `String`, constructors). Uses `TryGetInt64()` with string fallback for overflow/decimal safety.
- **`RequestIdTypeInfoResolver`** — `IJsonTypeInfoResolver` that returns type info via `JsonMetadataServices.CreateValueInfo<T>()` (the same API the source generator uses internally). Zero reflection.

This follows the existing converter pattern in the SDK (see `PermissionRequestResultKind.Converter` in Types.cs).

## Tests

- `RequestId_CanBeSerializedAndDeserialized_WithSdkOptions` — round-trips long, string, and null RequestIds
- `SerializerOptions_CanResolveRequestIdTypeInfo` — verifies the resolver chain produces type info
- `RequestId_NumericEdgeCases_RoundTrip` — `[Theory]` covering 0, -1, `long.MaxValue`

All 5 tests fail without the fix (`NotSupportedException`) and pass with it.

### DIFF
--- a/dotnet/src/Client.cs
+++ b/dotnet/src/Client.cs
@@ -14,6 +14,7 @@ using System.Net.Sockets;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using System.Text.Json.Serialization.Metadata;
 using System.Text.RegularExpressions;
 using GitHub.Copilot.SDK.Rpc;
 using System.Globalization;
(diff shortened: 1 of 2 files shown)


PureWeen:
> Hmm... why can't this instead be done just by adding `[JsonSerializable(typeof(RequestId))]` to one of the existing json serialization contexts, like here? https://github.com/PureWeen/copilot-sdk/blob/d512e5a718a1917d2130f5e21af99f57eadcc76b/dotnet/src/Client.cs#L1648
> 
> EDIT: Nevermind, I see the PR comment: "RequestId couldn't be added via [JsonSerializable] because it has [JsonConverter(typeof(RequestIdSTJsonConverter))] and that converter is internal to StreamJsonRpc (SYSLIB1220). The converter also can't be used directly (CS0122)" Can you file an issue on StreamJsonRpc about that? Such converters need to be public for this very reason.

https://github.com/microsoft/vs-streamjsonrpc/issues/1407

PR #791, opened 2026-03-11
Serialize event dispatch in .NET and Go SDKs

stephentoub:
## Summary

Decouple user event handler dispatch from the transport in both .NET and Go SDKs. Events are routed through a channel, consumed by a single background goroutine/task that invokes user handlers serially in FIFO order. Broadcast handlers (tool calls, permission requests) are fired as fire-and-forget directly from the dispatch entry point, outside the channel.

This brings .NET and Go in line with the Node.js and Python SDKs, which get natural serialization from their single-threaded event loops.

## Problem

In .NET, StreamJsonRpc dispatches notifications concurrently on the ThreadPool, which meant:
- Event handlers could run **concurrently with themselves** (two events arriving at once)
- Event handlers could run **out of order**

In Go, the JSON-RPC read loop was single-threaded so user handlers were already serialized, but broadcast handlers (tool calls, permission requests) ran inline, which meant:
- A broadcast handler issuing an RPC request would **deadlock** — the response can only be read by the read loop, which is blocked waiting for the handler to return
- A slow or stalled broadcast handler would **block all event delivery**

## Design

Both SDKs now use the same pattern, adapted to their runtime:

| | User handlers | Broadcast handlers | Blocks reader? |
|---|---|---|---|
| **Node.js** | Sync inline (event loop) | `void` fire-and-forget | No |
| **Python** | Sync inline (with lock) | `asyncio.ensure_future` | No |
| **Go** | Channel → consumer goroutine | `go` goroutine | No |
| **.NET** | Channel → consumer task | `_ =` fire-and-forget task | No |

Key properties:
- **Serial dispatch**: user handlers see events one at a time, never concurrently
- **FIFO ordering**: events delivered in arrival order
- **No transport blocking**: the reader thread/goroutine returns immediately after enqueueing
- **No deadlocks**: broadcast handlers run in their own goroutine/task, decoupled from the reader and the consumer

## .NET Changes

- **Unbounded `Channel<SessionEvent>`** with single-reader consumer task (`ProcessEventsAsync`) that invokes user handlers serially
- **`DispatchEvent`** enqueues to channel and fires broadcast handler as `_ = HandleBroadcastEventAsync(...)` (fire-and-forget)
- **Per-handler error catching** via `ImmutableArray<SessionEventHandler>` iteration with cached snapshot
- **`HandleBroadcastEventAsync`** wraps broadcast work in try/catch directly (no separate Safe wrapper)
- Channel completed in `DisposeAsync`; consumer task exits naturally
- `ILogger` threaded from `CopilotClient` to `CopilotSession` for diagnostics

## Go Changes

- **`eventCh` channel** added to Session; `processEvents` consumer goroutine started in `newSession`
- **`dispatchEvent`** enqueues to channel and fires `go s.handleBroadcastEvent(event)`
- Channel closed in `Disconnect` to stop the consumer
- Panic recovery per user handler to prevent one handler from halting delivery

## Tests

| SDK | Test | What it validates |
|-----|------|-------------------|
| .NET | `Should_Receive_Session_Events` (modified) | Serial dispatch — concurrent count never exceeds 1; awaits async delivery |
| .NET | `Handler_Exception_Does_Not_Halt_Event_Delivery` (new) | Consumer loop survives handler exceptions |
| .NET | `DisposeAsync_From_Handler_Does_Not_Deadlock` (new) | No deadlock when handler calls DisposeAsync |
| Go | `events are dispatched serially` (new) | Serial dispatch — max concurrent count is 1 |
| Go | `handler panic does not halt delivery` (new) | Panic recovery keeps dispatch alive |
| Go | `should_receive_session_events` (modified) | Awaits async delivery of session.start |

### DIFF
--- a/go/session.go
+++ b/go/session.go
@@ -67,7 +67,8 @@ type Session struct {
 
 	// eventCh serializes user event handler dispatch. dispatchEvent enqueues;
 	// a single goroutine (processEvents) dequeues and invokes handlers in FIFO order.
-	eventCh chan SessionEvent
+	eventCh   chan SessionEvent
+	closeOnce sync.Once // guards eventCh close so Disconnect is safe to call more than once
 
 	// RPC provides typed session-scoped RPC methods.
 	RPC *rpc.SessionRpc
@@ -451,7 +452,15 @@ func (s *Session) handleHooksInvoke(hookType string, rawInput json.RawMessage) (
 // serial, FIFO dispatch without blocking the read loop.
 func (s *Session) dispatchEvent(event SessionEvent) {
 	go s.handleBroadcastEvent(event)
-	s.eventCh <- event
+
+	// Send to the event channel in a closure with a recover guard.
+	// Disconnect closes eventCh, and in Go sending on a closed channel
+	// panics — there is no non-panicking send primitive. We only want
+	// to suppress that specific panic; other panics are not expected here.
+	func() {
+		defer func() { recover() }()
+		s.eventCh <- event
+	}()
 }
 
 // processEvents is the single consumer goroutine for the event channel.
@@ -657,7 +666,7 @@ func (s *Session) Disconnect() error {
 		return fmt.Errorf("failed to disconnect session: %w", err)
 	}
 
-	close(s.eventCh)
+	s.closeOnce.Do(func() { close(s.eventCh) })
 
 	// Clear handlers
 	s.handlerMutex.Lock()


stephentoub:
## Re: Cross-SDK Consistency Review

> *This response was generated by GitHub Copilot CLI.*

The consistency review above analyzed a stale version of the PR. The current implementation fires broadcast handlers as **fire-and-forget in all four SDKs**, which is the correct design. The review's recommendation to \wait\ broadcast handlers in Python would reintroduce the deadlock/blocking problems this PR fixes.

### Actual state after this PR

| SDK | User handlers | Broadcast handlers |
|---|---|---|
| **Node.js** | Serial (sync loop) | Fire-and-forget (\oid this._executeToolAndRespond()\) |
| **Python** | Serial (sync loop) | Fire-and-forget (\syncio.ensure_future()\) |
| **Go** | Serial (channel consumer) | Fire-and-forget (\go s.handleBroadcastEvent()\) |
| **.NET** | Serial (channel consumer) | Fire-and-forget (\_ = HandleBroadcastEventAsync()\) |

All four SDKs are consistent. Broadcast handlers **must** run concurrently/fire-and-forget because:

1. **Deadlock prevention** — in Go, the read loop goroutine calls \dispatchEvent\. If a broadcast handler issues an RPC request (e.g., \HandlePendingToolCall\), the response can only be read by the read loop — but it's blocked waiting for the handler. Inline execution deadlocks.

2. **Multi-client scenarios** — in multi-client tests, one client's permission handler intentionally never completes (\select {}\ in Go, \
ew TaskCompletionSource<T>().Task\ in .NET). If broadcast handlers block event delivery, the second client never sees events.

The review's table incorrectly shows .NET and Go as having serial/inline broadcast handlers. That was an earlier iteration that was reverted for exactly these reasons.

PR #797, opened 2026-03-12
fix(python): honor auto_restart on process exit

### LINKED ISSUE #789: Python has an `auto_restart` argument for `CopilotClient`, but it is never used
Should it be removed, supported, or just left as a no-op?

haosenwang1018:
## Summary
- route Python SDK requests through a stable request proxy so `auto_restart` is actually honored
- reconnect once and retry the failed RPC when the underlying CLI process exits unexpectedly
- keep existing `CopilotSession` objects wired to the refreshed transport after reconnect
- add focused unit coverage for both the retrying and non-retrying paths

## Testing
- python -m ruff check python/copilot/client.py python/test_client.py
- python -m pytest python/test_client.py -q

Closes #789.

### DIFF
--- a/python/copilot/client.py
+++ b/python/copilot/client.py
@@ -75,6 +75,16 @@ def _get_bundled_cli_path() -> str | None:
     return None
 
 
+class _AutoRestartRequestProxy:
+    """Stable request facade that can retry once after reconnecting."""
+
+    def __init__(self, owner: "CopilotClient"):
+        self._owner = owner
+
+    async def request(self, method: str, params: dict | None = None, **kwargs: Any) -> Any:
+        return await self._owner._request_with_auto_restart(method, params, **kwargs)
+
+
 class CopilotClient:
     """
     Main client for interacting with the Copilot CLI.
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks for propsing this.

However, this feature has never actually worked across any of the languages - there's non-implemented options for it on Python and Go, and then on TypeScript it tries to restart in certain cases but there are *many* cases where it won't work.

The feature dates back to the original spike of this SDK. I think we'd be better off removing it until we can make a more solid implementation, and of course only in reaction to real-world evidence of the need that can provide a clear case for what specifically should be supported.

I appreciate the PR you've added here makes it a little closer to working on Python, but it still won't work in a lot of cases (e.g., the CLI subprocess becomes frozen, or it was launched externally in the first place).

SteveSandersonMS:
Thanks for proposing this, @haosenwang1018.

On investigation, we realised that the "auto restart" feature was an artifact of the original spike implementation of this SDK and didn't really work in many cases (e.g., if the child process hangs or with `cliUrl`). Plus there wasn't even any implementation for it in the Go, Python, or .NET SDKs despite the option being present.

To clean this all up we decided to take a different approach, which is to remove the option since it doesn't really work (https://github.com/github/copilot-sdk/pull/803). Since there's no longer a feature here for the PR to be about, I hope you don't mind me closing the PR.

We'd love your contributions in other areas though if you find other things you'd like to fix. Thanks again!

PR #808, opened 2026-03-12
Allow tools to set `skipPermission`

MRayermannMSFT:
A nice to have for extensions adding relatively safe tools.

### DIFF
--- a/dotnet/src/Types.cs
+++ b/dotnet/src/Types.cs
@@ -283,6 +283,9 @@ public readonly struct PermissionRequestResultKind : IEquatable<PermissionReques
     /// <summary>Gets the kind indicating the permission was denied interactively by the user.</summary>
     public static PermissionRequestResultKind DeniedInteractivelyByUser { get; } = new("denied-interactively-by-user");
 
+    /// <summary>Gets the kind indicating the permission was denied interactively by the user.</summary>
+    public static PermissionRequestResultKind NoResult { get; } = new("no-result");
+
     /// <summary>Gets the underlying string value of this <see cref="PermissionRequestResultKind"/>.</summary>
     public string Value => _value ?? string.Empty;
 
--- a/dotnet/test/ToolsTests.cs
+++ b/dotnet/test/ToolsTests.cs
@@ -195,12 +195,15 @@ public partial class ToolsTests(E2ETestFixture fixture, ITestOutputHelper output
                 new Dictionary<string, object?> { ["skip_permission"] = true })
         });
 
-        // TODO: Once the CLI respects skip_permission, use a tracking permission handler
-        // and assert it was NOT called for this tool.
+        var didRunPermissionRequest = false;
         var session = await CreateSessionAsync(new SessionConfig
         {
             Tools = [tool],
-            OnPermissionRequest = PermissionHandler.ApproveAll,
+            OnPermissionRequest = (_, _) =>
+            {
+                didRunPermissionRequest = true;
+                return Task.FromResult(new PermissionRequestResult { Kind = PermissionRequestResultKind.NoResult });
+            }
         });
 
         await session.SendAsync(new MessageOptions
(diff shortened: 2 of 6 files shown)


dotnet/src/Client.cs
```diff
@@ -1470,13 +1470,16 @@ internal record ToolDefinition(
...
         JsonElement Parameters, /* JSON schema */
-        bool? OverridesBuiltInTool = null)
+        bool? OverridesBuiltInTool = null,
+        bool? SkipPermission = null)
     {
         public static ToolDefinition FromAIFunction(AIFunction function)
         {
             var overrides = function.AdditionalProperties.TryGetValue("is_override", out var val) && val is true;
+            var skipPerm = function.AdditionalProperties.TryGetValue("skip_permission", out var skipVal) && skipVal is true;
             return new ToolDefinition(function.Name, function.Description, function.JsonSchema,
```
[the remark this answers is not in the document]
MRayermannMSFT: fixed

python/copilot/client.py
```diff
@@ -507,6 +507,8 @@ async def create_session(self, config: SessionConfig) -> CopilotSession:
                     definition["parameters"] = tool.parameters
                 if tool.overrides_built_in_tool:
                     definition["overridesBuiltInTool"] = True
+                if tool.skip_permission:
+                    definition["skipPermission"] = True
```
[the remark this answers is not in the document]
MRayermannMSFT: fixed

python/copilot/tools.py
```diff
@@ -47,6 +49,7 @@ def define_tool(
     handler: Callable[[Any, ToolInvocation], Any] | None = None,
     params_type: type[BaseModel] | None = None,
     overrides_built_in_tool: bool = False,
+    skip_permission: bool = False,
 ) -> Tool | Callable[[Callable[[Any, ToolInvocation], Any]], Tool]:
```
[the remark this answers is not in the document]
MRayermannMSFT: fixed

go/client_test.go
```diff
@@ -509,6 +509,47 @@ func TestOverridesBuiltInTool(t *testing.T) {
...
+			t.Fatalf("failed to marshal: %v", err)
+		}
+		var m map[string]any
+		if err := json.Unmarshal(data, &m); err != nil {
+			t.Fatalf("failed to unmarshal: %v", err)
+		}
+		if _, ok := m["skipPermission"]; ok {
+			t.Errorf("expected skipPermission to be omitted, got %v", m)
+		}
+	})
```
SteveSandersonMS: It's strange that for go it doesn't actually test the feature (it only tests how the type is serialized), whereas for other languages it does.

nodejs/test/client.test.ts
```diff
@@ -378,6 +378,64 @@ describe("CopilotClient", () => {
...
+                ],
+            });
+
+            const payload = spy.mock.calls.find((c) => c[0] === "session.resume")![1] as any;
+            expect(payload.tools).toEqual([
+                expect.objectContaining({ name: "my_tool", skipPermission: true }),
+            ]);
+            spy.mockRestore();
+        });
+    });
```
SteveSandersonMS: My guess is we probably have inconsistencies in the tools tests prior to this PR, and the agent is just following the patterns.

But now we're paying attention to it, I don't *think* we want to have an E2E test for .NET for this but a mock-based test for Node. Could you see if this switches over to a true E2E test without much difficulty? All the new features are being tested E2E already so it should have all the infrastructure it needs across all the languages, and you've already shown it works and has a valid snapshot based on .NET.

python/test_client.py
```diff
@@ -237,6 +237,100 @@ def grep(params) -> str:
             await client.force_stop()
 
 
+class TestSkipPermission:
+    @pytest.mark.asyncio
+    async def test_skip_permission_sent_in_tool_definition(self):
```
SteveSandersonMS: Looks like this could be an E2E test too

MRayermannMSFT: fixed

nodejs/test/client.test.ts
MRayermannMSFT: fixed

go/client_test.go
MRayermannMSFT: fixed

PR #823, opened 2026-03-13
test: cover python list_sessions empty filter

haosenwang1018:
## Summary
- add a focused Python client test for `list_sessions()` without a filter
- verify the client sends an empty payload when no `SessionListFilter` is provided
- verify the empty server response is returned as an empty list

## Why
The Python client now has targeted coverage for filtered `list_sessions()` calls, but not for the no-filter default path. This PR locks down the default wire contract so regressions in payload shape or empty-response handling are caught in CI.

## Validation
- `python -m pytest -q python/test_client.py -k 'list_sessions_sends_empty_payload_without_filter'`
- `git diff --check`

### DIFF
--- a/python/test_client.py
+++ b/python/test_client.py
@@ -480,3 +480,26 @@ class TestSessionConfigForwarding:
             assert captured["session.model.switchTo"]["modelId"] == "gpt-4.1"
         finally:
             await client.force_stop()
+
+    @pytest.mark.asyncio
+    async def test_list_sessions_sends_empty_payload_without_filter(self):
+        client = CopilotClient({"cli_path": CLI_PATH})
+        await client.start()
+
+        try:
+            captured = {}
+            original_request = client._client.request
+
+            async def mock_request(method, params):
+                captured[method] = params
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks @haosenwang1018 for submitting all these PRs with test coverage.

As it stands these are not something we would merge because:

1. They only cover Python, and we need to cover all the languages we support. Making the test suites diverge significantly will make it harder to keep the languages in sync.
2. They are largely mock-oriented, which is not the test pattern we're trying to focus on. Using mocks means we're locking down internal implementation details instead of observing the true behavior. For tests that describe interactions with the underlying CLI server, we would prefer to use the E2E test infrastructure. This also helps to ensure we're only testing things that can actually happen in reality.

If you're interested in working with us on test coverage that's wonderful and I'd be happy to discuss more. I'd recommend not submitting large numbers of PRs without coordination though as we are really focused on keeping the project moving and we won't necessarily have capacity to review and merge things that are not on our roadmap.

PR #825, opened 2026-03-13
test: cover python get_last_session_id none case

haosenwang1018:
## Summary
- add a focused Python client test for `get_last_session_id()` when the server returns no session id
- verify the client still sends the expected empty payload to `session.getLastId`
- verify the Python API returns `None` for the no-session case

## Why
The happy path for `get_last_session_id()` is now covered, but the default/no-session path was still untested. This PR locks down that behavior so regressions in payload shape or `None` handling are caught in CI.

## Validation
- `python -m pytest -q python/test_client.py -k 'get_last_session_id_returns_none_when_absent'`
- `git diff --check`

### DIFF
--- a/python/test_client.py
+++ b/python/test_client.py
@@ -480,3 +480,26 @@ class TestSessionConfigForwarding:
             assert captured["session.model.switchTo"]["modelId"] == "gpt-4.1"
         finally:
             await client.force_stop()
+
+    @pytest.mark.asyncio
+    async def test_get_last_session_id_returns_none_when_absent(self):
+        client = CopilotClient({"cli_path": CLI_PATH})
+        await client.start()
+
+        try:
+            captured = {}
+            original_request = client._client.request
+
+            async def mock_request(method, params):
+                captured[method] = params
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks @haosenwang1018 for submitting all these PRs with test coverage.

As it stands these are not something we would merge because:

1. They only cover Python, and we need to cover all the languages we support. Making the test suites diverge significantly will make it harder to keep the languages in sync.
2. They are largely mock-oriented, which is not the test pattern we're trying to focus on. Using mocks means we're locking down internal implementation details instead of observing the true behavior. For tests that describe interactions with the underlying CLI server, we would prefer to use the E2E test infrastructure. This also helps to ensure we're only testing things that can actually happen in reality.

If you're interested in working with us on test coverage that's wonderful and I'd be happy to discuss more. I'd recommend not submitting large numbers of PRs without coordination though as we are really focused on keeping the project moving and we won't necessarily have capacity to review and merge things that are not on our roadmap.

PR #863, opened 2026-03-15
fix(go): capture CLI stderr and fix SetProcessDone race

claudiogodoy99:
# Fix: Capture CLI stderr and resolve SetProcessDone race

## Problem

When the CLI process exits during startup, the SDK reports an opaque error with no indication of *why* the process failed:

```
panic: process exited unexpectedly
        failed to kill CLI process: os: process already finished
```

The actual cause (e.g. missing module, auth failure, version mismatch) is written to stderr, which the SDK silently discards.

## Root Causes

1. **Stderr not captured** — `client.go` never sets `exec.Cmd.Stderr`, so all CLI diagnostic output is lost.
2. **Race in `SetProcessDone`** — `jsonrpc2.go` copies the process error inside an async goroutine. When `Request()` detects `processDone` is closed and immediately calls `getProcessError()`, the goroutine hasn't run yet, so the error is `nil` and callers get the generic `"process exited unexpectedly"` message.

## Changes

| File | Change |
|------|--------|
| `client.go` | Set `c.process.Stderr = &bytes.Buffer{}` before `Start()` in both stdio and TCP branches |
| `client.go` | Read stderr buffer in `monitorProcess` after `Wait()` and include contents in error |
| `internal/jsonrpc2/jsonrpc2.go` | Replace async goroutine in `SetProcessDone` with direct error pointer storage |
| `internal/jsonrpc2/jsonrpc2.go` | `getProcessError()` now dereferences the stored pointer, which is set before channel close |
| `client_test.go` | Tests verifying stderr is captured on both zero and non-zero exits |
| `internal/jsonrpc2/jsonrpc2_test.go` | Tests verifying error is available immediately after `processDone` closes |

## After

```
panic: CLI process exited: exit status 1
        stderr: Error: Cannot find module '.../copilot/index.js'
            at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
            ...
```

## Testing

- `TestMonitorProcess_StderrCaptured` — stderr included in error on non-zero exit
- `TestMonitorProcess_StderrCapturedOnZeroExit` — stderr included on exit code 0
- `TestStartCLIServer_StderrFieldSet` — confirms `Stderr` is assigned
- `TestSetProcessDone_ErrorAvailableImmediately` — no race: error accessible right after channel close
- `TestSetProcessDone_RequestMissesProcessError` — `Request()` returns actual error, not generic message
- `TestSetProcessDone_ErrorCopiedEventually` — sanity check that error pointer works after yield

## Manual Test Evidence

**Before the fix** — stderr is discarded, error is opaque:

```
panic: process exited unexpectedly
        failed to kill CLI process: os: process already finished
```

**After the fix** — stderr is captured and included in the error:

```
$ go run samples/chat.go
panic: CLI process exited: exit status 1
        stderr: node:internal/modules/cjs/loader:1386
          throw err;
          ^

        Error: Cannot find module '/home/claudiogodoy/repos/fork/nodejs/node_modules/@github/copilot/index.js'
            at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
            at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
            at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
            at Function._load (node:internal/modules/cjs/loader:1192:37)
            at TracingChannel.traceSync (node:diagnostics_channel:328:14)
            at wrapModuleLoad (node:internal/modules/cjs/loader:237:24)
            at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
            at node:internal/main/run_main_module:36:49 {
          code: 'MODULE_NOT_FOUND',
          requireStack: []
        }

        Node.js v22.21.0
        failed to kill CLI process: os: process already finished

goroutine 1 [running]:
main.main()
        /home/claudiogodoy/repos/fork/copilot-sdk/go/samples/chat.go:22 +0x4a7
exit status 2
```

The error now clearly shows the CLI failed because of a missing Node.js module, making the root cause immediately actionable.

### DIFF
--- a/go/client.go
+++ b/go/client.go
@@ -30,7 +30,6 @@ package copilot
 
 import (
 	"bufio"
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -49,6 +48,7 @@ import (
 
 	"github.com/github/copilot-sdk/go/internal/embeddedcli"
 	"github.com/github/copilot-sdk/go/internal/jsonrpc2"
+	"github.com/github/copilot-sdk/go/internal/truncbuffer"
 	"github.com/github/copilot-sdk/go/rpc"
 )
 
@@ -1443,6 +1443,11 @@ func (c *Client) verifyProtocolVersion(ctx context.Context) error {
 	return nil
 }
 
+// stderrBufferSize is the maximum number of bytes kept from the CLI process's
+// stderr. Only the tail is retained so that memory stays bounded even when the
+// process produces a large amount of diagnostic output.
+const stderrBufferSize = 64 * 1024
+
 // startCLIServer starts the CLI server process.
 //
 // This spawns the CLI server as a subprocess using the configured transport
@@ -1559,7 +1564,7 @@ func (c *Client) startCLIServer(ctx context.Context) error {
 			return fmt.Errorf("failed to create stdout pipe: %w", err)
 		}
 
-		c.process.Stderr = &bytes.Buffer{}
+		c.process.Stderr = truncbuffer.NewTruncBuffer(stderrBufferSize)
 
 		if err := c.process.Start(); err != nil {
 			return fmt.Errorf("failed to start CLI server: %w", err)
@@ -1592,7 +1597,7 @@ func (c *Client) startCLIServer(ctx context.Context) error {
 			return fmt.Errorf("failed to create stdout pipe: %w", err)
 		}
 
-		c.process.Stderr = &bytes.Buffer{}
+		c.process.Stderr = truncbuffer.NewTruncBuffer(stderrBufferSize)
 
 		if err := c.process.Start(); err != nil {
 			return fmt.Errorf("failed to start CLI server: %w", err)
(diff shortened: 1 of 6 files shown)


go/client.go
```diff
@@ -48,6 +48,7 @@ import (
 
 	"github.com/github/copilot-sdk/go/internal/embeddedcli"
 	"github.com/github/copilot-sdk/go/internal/jsonrpc2"
+	"github.com/github/copilot-sdk/go/internal/truncbuffer"
```
qmuntal, line 51: This package does not exist 🐱

qmuntal (changes requested):

go/client.go
claudiogodoy99: My bad I forgot adding the new files on the las commit, sorry!

SteveSandersonMS:
@qmuntal Would you be able to decide if this is correct, and if so merge it? Thanks!

qmuntal (approved):
LGTM, thanks!

stephentoub (approved):

PR #868, opened 2026-03-16
Fix(dotnet): serialization of SessionEvent

xoofx:
## Summary

This fixes a .NET SDK serialization gap where `SessionEvent.ToJson()` could fail for valid session events containing payloads that are stored through `object`-typed members and materialize as `JsonElement` after deserialization.

Affected event shapes include:

- `assistant.message` with `toolRequests[].arguments`
- `tool.execution_start` with `arguments`
- `tool.execution_complete` with nested `toolTelemetry`
- `session.shutdown` with nested `modelMetrics`

## Problem

`SessionEvent.FromJson(...)` could successfully deserialize these events, but `SessionEvent.ToJson()` could then throw `System.NotSupportedException` because the source-generated `SessionEventsJsonContext` did not include metadata for `JsonElement`.

This is a real issue for SDK consumers that subscribe to events and re-serialize them for logging, capture, persistence, or forwarding.

A concrete symptom is that richer runtime events can appear to be “missing” from application-side captures even though the Copilot session emitted and persisted them correctly.

## Root cause

Some generated session event models contain members such as:

- `object? Arguments`
- `Dictionary<string, object>? ToolTelemetry`
- `Dictionary<string, object> ModelMetrics`

When those are populated from JSON, their runtime values are often `JsonElement`. The generated `SessionEventsJsonContext` did not include `JsonElement` / `JsonElement?`, so re-serialization failed.

## Fix

Update the C# code generator to include:

- `JsonElement`
- `JsonElement?`

in the generated `SessionEventsJsonContext`.

This keeps the fix source-generated and AOT/trimming-friendly, instead of relying on reflection-based serialization.

## Generator change

The fix was made in the generator, not only in generated output:

- [csharp.ts](C:/code/github/copilot-sdk/scripts/codegen/csharp.ts)

and then regenerated into:

- [SessionEvents.cs](C:/code/github/copilot-sdk/dotnet/src/Generated/SessionEvents.cs)

## Tests

Added a regression test that builds typed C# session event objects and verifies `ToJson()` round-trips event payloads that contain `JsonElement`-backed dynamic members.

Covered cases:

- `assistant.message`
- `tool.execution_start`
- `tool.execution_complete`
- `session.shutdown`

The test intentionally constructs real SDK event/data types so it will fail if the generated C# model shape changes.

## Validation

Verified locally with:

```powershell
dotnet build C:\code\github\copilot-sdk\dotnet\src\GitHub.Copilot.SDK.csproj -c Release
dotnet test C:\code\github\copilot-sdk\dotnet\test\GitHub.Copilot.SDK.Test.csproj -c Release --filter SessionEventSerializationTests
```

The targeted regression tests pass.

### DIFF
--- a/dotnet/src/Generated/SessionEvents.cs
+++ b/dotnet/src/Generated/SessionEvents.cs
@@ -3527,4 +3527,5 @@ public enum PermissionCompletedDataResultKind
 [JsonSerializable(typeof(UserMessageDataAttachmentsItemSelectionSelectionEnd))]
 [JsonSerializable(typeof(UserMessageDataAttachmentsItemSelectionSelectionStart))]
 [JsonSerializable(typeof(UserMessageEvent))]
+[JsonSerializable(typeof(JsonElement))]
 internal partial class SessionEventsJsonContext : JsonSerializerContext;
--- a/dotnet/test/SessionEventSerializationTests.cs
+++ b/dotnet/test/SessionEventSerializationTests.cs
@@ -1,0 +1,180 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------------------------------------------*/
+
+using System.Collections.Generic;
+using System.Text.Json;
+using Xunit;
+
+namespace GitHub.Copilot.SDK.Test;
+
+public class SessionEventSerializationTests
+{
+    public static TheoryData<SessionEvent, string> JsonElementBackedEvents => new()
+    {
+        {
+            new AssistantMessageEvent
+            {
+                Id = Guid.Parse("11111111-1111-1111-1111-111111111111"),
+                Timestamp = DateTimeOffset.Parse("2026-03-15T21:26:02.642Z"),
+                ParentId = Guid.Parse("22222222-2222-2222-2222-222222222222"),
+                Data = new AssistantMessageData
+                {
+                    MessageId = "msg-1",
+                    Content = "",
+                    ToolRequests =
+                    [
+                        new AssistantMessageDataToolRequestsItem
+                        {
+                            ToolCallId = "call-1",
+                            Name = "view",
+                            Arguments = ParseJsonElement("""{"path":"README.md"}"""),
+                            Type = AssistantMessageDataToolRequestsItemType.Function,
+                        },
+                    ],
+                },
+            },
+            "assistant.message"
+        },
+        {
+            new ToolExecutionStartEvent
+            {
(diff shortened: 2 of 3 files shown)


dotnet/src/Generated/SessionEvents.cs
```diff
@@ -3527,4 +3527,6 @@ public enum PermissionCompletedDataResultKind
 [JsonSerializable(typeof(UserMessageDataAttachmentsItemSelectionSelectionEnd))]
 [JsonSerializable(typeof(UserMessageDataAttachmentsItemSelectionSelectionStart))]
 [JsonSerializable(typeof(UserMessageEvent))]
+[JsonSerializable(typeof(JsonElement))]
+[JsonSerializable(typeof(JsonElement?))]
```
stephentoub: What requires a `JsonElement?`?

xoofx: When you receive events in C#, they are deserialized from RPCJson, for example this property:

https://github.com/github/copilot-sdk/blob/485ea5ed1ce43125075bab2f3d2681f1816a4f9a/dotnet/src/Generated/SessionEvents.cs#L1638

Will be actually a `JsonElement`

If you try to serialize back a `SessionEvent` to Json, any of the C# types that are implicitly using `JsonElement` will fail to serialize.

Basically, I discovered this problem when trying to log all C# SessionEvent to json, as many of them were not serializing at all silently. These are the other cases discovered:

> - `object? Arguments`
> - `Dictionary<string, object>? ToolTelemetry`
> - `Dictionary<string, object> ModelMetrics`

Maybe a better fix long term would be to type formalize these elements, but it goes beyond this fix.

stephentoub: Yes, I understand why `JsonElement` is needed... I'm wondering why `JsonElement?` is needed, i.e. `Nullable<JsonElement>`?

xoofx: > Yes, I understand why JsonElement is needed... I'm wondering why JsonElement? is needed, i.e. Nullable<JsonElement>?

I encountered on another project a problem where the source generator could generate code that generates warning without `type(JsonElement?)` (in the case of `JsonElement` => `object?`) and I believe it will be same case here. But I will double check, it might be an oversight (Or I'm conflicting it with some source generators I developed recently)

xoofx: Double checked, and it is not necessary, mixed up with another project I was working on. Fixed via push d51be51

stephentoub: Thanks for confirming.

PR #873, opened 2026-03-17
[Node] Add Support For Commands and Simple UI

MRayermannMSFT:


nodejs/test/client.test.ts
```diff
@@ -650,4 +650,265 @@ describe("CopilotClient", () => {
             expect(params.tracestate).toBeUndefined();
         });
     });
+
+    describe("commands in session creation", () => {
+        it("forwards commands metadata in session.create request", async () => {
```
SteveSandersonMS, line 655: I was about to ask for these mock-based tests to be replaced with E2E ones, but is that actually possible? My guess is it would only be possible if there's also a way to *trigger* commands from an SDK client. Right now we may be lacking that and only the CLI can trigger them. Is that right?

Of course I understand that for elicitation, it would only be possible to have an E2E test if there was also some UI implementation we can observe, which I don't think will be the case from this repo.

PR #883, opened 2026-03-18
fix: Go codegen enum prefixes and type name reconciliation

MackinnonBuck:
## Summary

Fixes two Go codegen bugs introduced in #878 (Update @github/copilot to 1.0.7).

### Bug 1: Build-breaking type name mismatch
RPC wrapper code computed type names via `toPascalCase()` (e.g., `SessionMcpListResult`) but quicktype generated different names following Go initialism rules (e.g., `SessionMCPListResult`). **The Go SDK didn't compile.**

**Fix:** After quicktype runs, extract actual generated struct names and use a `resolveType()` lookup instead of recomputing via `toPascalCase`. Also added `mcp` to `goInitialisms`.

### Bug 2: Nonsense/missing enum prefixes
quicktype used whimsical prefixes for colliding enum constants (`PurpleDisabled`, `FluffyFailed`, `TentacledFailed`) and left non-colliding constants unprefixed (`Running`, `Connected`, `Error`). The canonical Go convention is `TypeNameValue`.

**Fix:** `postProcessEnumConstants()` renames all unprefixed constants inside `const` blocks and their associated marshal/unmarshal functions to follow `TypeNameValue` convention, while preserving struct field names.

### Additional fix: Field name reconciliation
quicktype sometimes renames struct fields to avoid Go keyword conflicts. The wrapper code now extracts actual field names from quicktype output instead of recomputing them.

## Breaking change
All Go enum constants now use `TypeNameValue` prefixes. Downstream code referencing old names (e.g., `copilot.SessionIdle` -> `copilot.SessionEventTypeSessionIdle`, `rpc.Interactive` -> `rpc.ModeInteractive`) must be updated.

## Verification
- `go build ./...` passes
- `go test ./...` passes (2 pre-existing E2E failures unrelated to this change)

### DIFF
--- a/docs/features/image-input.md
+++ b/docs/features/image-input.md
@@ -315,7 +315,7 @@ func main() {
 		Prompt: "Describe what you see in this image",
 		Attachments: []copilot.Attachment{
 			{
-				Type:        copilot.Blob,
+				Type:        copilot.AttachmentTypeBlob,
 				Data:        &base64ImageData,
 				MIMEType:    &mimeType,
 				DisplayName: &displayName,
@@ -333,7 +333,7 @@ session.Send(ctx, copilot.MessageOptions{
     Prompt: "Describe what you see in this image",
     Attachments: []copilot.Attachment{
         {
-            Type:        copilot.Blob,
+            Type:        copilot.AttachmentTypeBlob,
             Data:        &base64ImageData, // base64-encoded string
             MIMEType:    &mimeType,
             DisplayName: &displayName,
--- a/test/scenarios/prompts/attachments/README.md
+++ b/test/scenarios/prompts/attachments/README.md
@@ -39,7 +39,7 @@ Demonstrates sending **file attachments** alongside a prompt using the Copilot S
 |----------|------------------------|
 | TypeScript | `attachments: [{ type: "blob", data: base64Data, mimeType: "image/png" }]` |
 | Python | `"attachments": [{"type": "blob", "data": base64_data, "mimeType": "image/png"}]` |
-| Go | `Attachments: []copilot.Attachment{{Type: copilot.Blob, Data: &data, MIMEType: &mime}}` |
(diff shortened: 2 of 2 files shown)


scripts/codegen/go.ts
```diff
@@ -153,6 +224,22 @@ async function generateRpc(schemaPath?: string): Promise<void> {
...
+    const actualTypeNames = new Map<string, string>();
+    const structRe = /^type\s+(\w+)\s+struct\b/gm;
+    let sm;
+    while ((sm = structRe.exec(qtCode)) !== null) {
+        actualTypeNames.set(sm[1].toLowerCase(), sm[1]);
+    }
+    const resolveType = (name: string): string => actualTypeNames.get(name.toLowerCase()) ?? name;
+
+    // Extract field name mappings (quicktype may rename fields to avoid Go keyword conflicts)
+    const fieldNames = extractFieldNames(qtCode);
```
SteveSandersonMS, line 244: We're getting to a pretty high level of customization for the Go output. I wonder if, at some point soon, it will be a net simplification to stop using Quicktype for Go entirely in favour of a standalone generator like we have for C#.

Not suggesting doing that here. Just a thought for the future.

go/rpc/generated_rpc.go
```diff
@@ -506,67 +506,67 @@ type SessionShellKillParams struct {
...
 
 // Connection status: connected, failed, pending, disabled, or not_configured
 type ServerStatus string
 
 const (
-	Connected      ServerStatus = "connected"
-	NotConfigured  ServerStatus = "not_configured"
-	Pending        ServerStatus = "pending"
-	PurpleDisabled ServerStatus = "disabled"
-	PurpleFailed   ServerStatus = "failed"
```
SteveSandersonMS: Ugh, glad to see the end of these bizarre names.

scripts/codegen/go.ts
MackinnonBuck: It does seem like we're nearing that point. A lot of this code only exists to work around quicktype limitations.

PR #899, opened 2026-03-20
feat: add session.getMetadata to all SDK languages

MackinnonBuck:
## Motivation

The runtime now exposes a `session.getMetadata` JSON-RPC method (copilot-agent-runtime#5021) that provides efficient O(1) lookup of a single session's metadata by ID — avoiding the need to call `session.list` and filter client-side. This PR exposes that endpoint in all four SDK language bindings.

## Approach

Each SDK already has `listSessions` returning `SessionMetadata[]`. The new method reuses the **same `SessionMetadata` type** (no new public types needed) but returns a single optional result. When the session is not found, the method returns the language-idiomatic "absent" value rather than throwing an error.

| SDK | Method | Return type | Not-found value |
|-----|--------|-------------|-----------------|
| **Node.js** | `getSessionMetadata(sessionId)` | `Promise<SessionMetadata \| undefined>` | `undefined` |
| **Python** | `get_session_metadata(session_id)` | `SessionMetadata \| None` | `None` |
| **Go** | `GetSessionMetadata(ctx, sessionID)` | `(*SessionMetadata, error)` | `nil` |
| **.NET** | `GetSessionMetadataAsync(sessionId, ct)` | `Task<SessionMetadata?>` | `null` |

## Changes

- **Node.js** — New method in `client.ts` with full JSDoc; skipped E2E test (matching existing `listSessions` skip pattern)
- **Python** — New method in `client.py` with docstring; **running** E2E test (Python/Go session lifecycle tests are live)
- **Go** — New method in `client.go`, internal request/response structs in `types.go`; **running** E2E test
- **.NET** — New method in `Client.cs`, internal records + `[JsonSerializable]` registration; skipped E2E test
- **Snapshot** — `test/snapshots/session/should_get_session_metadata.yaml` for the replay proxy

## Notes

- E2E tests in Node.js and .NET are skipped with the same TODO as the existing `listSessions` tests ("Re-enable once test harness CAPI proxy supports this test's session lifecycle"). Python and Go E2E tests are running, consistent with their existing session lifecycle tests.
- All E2E tests cover both the happy path (create session → get metadata → verify fields) and the not-found case.

### DIFF
--- a/test/snapshots/session/should_get_session_metadata_by_id.yaml
+++ b/test/snapshots/session/should_get_session_metadata_by_id.yaml
@@ -1,0 +1,11 @@
+models:
+  - claude-sonnet-4.5
+conversations:
+  - messages:
+      - role: system
+        content: ${system}
+      - role: user
+        content: Say hello
+      - role: assistant
+        content: Hello! I'm GitHub Copilot CLI, ready to help you with your software engineering tasks. What can I assist you
+          with today?


python/copilot/client.py
```diff
@@ -1643,6 +1643,40 @@ async def list_sessions(self, filter: SessionListFilter | None = None) -> list[S
         sessions_data = response.get("sessions", [])
         return [SessionMetadata.from_dict(session) for session in sessions_data]
 
+    async def get_session_metadata(
+        self, session_id: str
+    ) -> "SessionMetadata | None":
```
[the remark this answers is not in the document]
MackinnonBuck: Fixed — removed the unnecessary string quotes since the file already uses `from __future__ import annotations`.

nodejs/src/client.ts
```diff
@@ -1088,6 +1088,55 @@ export class CopilotClient {
...
+
+        if (!session) {
+            return undefined;
+        }
+
+        return {
+            sessionId: session.sessionId,
+            startTime: new Date(session.startTime),
+            modifiedTime: new Date(session.modifiedTime),
+            summary: session.summary,
```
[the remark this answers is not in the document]
MackinnonBuck: Good call — extracted a private `toSessionMetadata` helper and updated both `listSessions` and `getSessionMetadata` to use it.

dotnet/test/SessionTests.cs
```diff
@@ -405,6 +405,23 @@ public async Task Should_List_Sessions_With_Context()
         }
     }
 
+    // TODO: Re-enable once test harness CAPI proxy supports this test's session lifecycle
+    [Fact(Skip = "Needs test harness CAPI proxy support")]
```
SteveSandersonMS: @MackinnonBuck Is this a change we should be doing in this PR?

dotnet/src/Client.cs
```diff
@@ -840,6 +840,36 @@ public async Task<List<SessionMetadata>> ListSessionsAsync(SessionListFilter? fi
...
+    ///     Console.WriteLine($"Session started at: {metadata.StartTime}");
+    /// }
+    /// </code>
+    /// </example>
+    public async Task<SessionMetadata?> GetSessionMetadataAsync(string sessionId, CancellationToken cancellationToken = default)
+    {
+        var connection = await EnsureConnectedAsync(cancellationToken);
+
+        var response = await InvokeRpcAsync<GetSessionMetadataResponse>(
+            connection.Rpc, "session.getMetadata", [new GetSessionMetadataRequest(sessionId)], cancellationToken);
```
SteveSandersonMS, line 868: Just curious why the need to hand-write the RPC rather than have it generated from JSON schema. I see the runtime side of it is also hand-written, but am interested in what made that necessary. If we can improve the schema-based RPC to avoid the need for that I'd love to do so.

dotnet/test/SessionTests.cs
MackinnonBuck: This is a new test and the "Skip" was copied from the existing test above it, but maybe it's safe to unskip this one. I'll try that.

dotnet/src/Client.cs
MackinnonBuck: This was following the same pattern as other root-level session methods like `ListSessionsAsync` and `DeleteSessionAsync`. Finding a way to auto-generate these seems reasonable. I'll plan to investigate that in a follow-up.

dotnet/test/SessionTests.cs
MackinnonBuck: Fixed.

SteveSandersonMS (approved):

PR #906, opened 2026-03-23
[Node] Add Commands and UI Elicitation Support to SDK

MRayermannMSFT:
## What

Adds commands and UI elicitation support to the Node.js SDK, enabling SDK clients to register slash commands that CLI users can invoke and to request interactive form dialogs via the CLI's elicitation component. The public API includes `commands` in session config for registering command handlers with automatic event routing, a `session.ui` object with convenience methods (`confirm`, `select`, `input`, and generic `elicitation`), and a `session.capabilities` getter for runtime feature detection. Includes unit tests for all new code paths, E2E tests for both features, and updated README documentation.ah

## Why

The CLI recently shipped support for SDK-registered commands and UI elicitation dialogs, but the Node.js SDK had no way to use these features. Without this change, SDK consumers building extensions or agents that participate in interactive CLI sessions cannot register slash commands or prompt users with structured input dialogs — limiting the SDK to headless-only workflows.

### DIFF
--- a/nodejs/test/e2e/commands.test.ts
+++ b/nodejs/test/e2e/commands.test.ts
@@ -23,37 +23,41 @@ describe("Commands", async () => {
         await client2.stop();
     });
 
-    it("client receives commands.changed when another client joins with commands", { timeout: 20_000 }, async () => {
-        const session1 = await client1.createSession({
-            onPermissionRequest: approveAll,
-        });
-
-        type CommandsChangedEvent = Extract<SessionEvent, { type: "commands.changed" }>;
-
-        // Wait for the commands.changed event deterministically
-        const commandsChangedPromise = new Promise<CommandsChangedEvent>((resolve) => {
-            session1.on((event) => {
-                if (event.type === "commands.changed") resolve(event);
+    it(
+        "client receives commands.changed when another client joins with commands",
+        { timeout: 20_000 },
+        async () => {
+            const session1 = await client1.createSession({
+                onPermissionRequest: approveAll,
             });
-        });
 
-        // Client2 joins with commands
-        const session2 = await client2.resumeSession(session1.sessionId, {
-            onPermissionRequest: approveAll,
-            commands: [
-                { name: "deploy", description: "Deploy the app", handler: async () => {} },
-            ],
-            disableResume: true,
-        });
+            type CommandsChangedEvent = Extract<SessionEvent, { type: "commands.changed" }>;
 
-        // Rely on default vitest timeout
-        const commandsChanged = await commandsChangedPromise;
-        expect(commandsChanged.data.commands).toEqual(
-            expect.arrayContaining([
-                expect.objectContaining({ name: "deploy", description: "Deploy the app" }),
-            ]),
-        );
+            // Wait for the commands.changed event deterministically
+            const commandsChangedPromise = new Promise<CommandsChangedEvent>((resolve) => {
+                session1.on((event) => {
+                    if (event.type === "commands.changed") resolve(event);
+                });
+            });
 
-        await session2.disconnect();
-    });
+            // Client2 joins with commands
+            const session2 = await client2.resumeSession(session1.sessionId, {
+                onPermissionRequest: approveAll,
+                commands: [
+                    { name: "deploy", description: "Deploy the app", handler: async () => {} },
+                ],
+                disableResume: true,
+            });
+
+            // Rely on default vitest timeout
+            const commandsChanged = await commandsChangedPromise;
+            expect(commandsChanged.data.commands).toEqual(
+                expect.arrayContaining([
+                    expect.objectContaining({ name: "deploy", description: "Deploy the app" }),
+                ])
+            );
+
+            await session2.disconnect();
+        }
+    );
 });


nodejs/src/session.ts
```diff
@@ -480,6 +565,108 @@ export class CopilotSession {
...
+
+    private async _input(message: string, options?: InputOptions): Promise<string | null> {
+        this.assertElicitation();
+        const field: Record<string, unknown> = { type: "string" as const };
+        if (options?.title) field.title = options.title;
+        if (options?.description) field.description = options.description;
+        if (options?.minLength != null) field.minLength = options.minLength;
+        if (options?.maxLength != null) field.maxLength = options.maxLength;
+        if (options?.format) field.format = options.format;
+        if (options?.default) field.default = options.default;
```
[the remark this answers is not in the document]
MRayermannMSFT: Fixed — changed to `options?.default != null` so empty strings are preserved.

nodejs/test/e2e/commands.test.ts
```diff
@@ -0,0 +1,57 @@
...
+            onPermissionRequest: approveAll,
+            commands: [
+                { name: "deploy", description: "Deploy the app", handler: async () => {} },
+            ],
+            disableResume: true,
+        });
+
+        // Wait for events to propagate
+        await new Promise((resolve) => setTimeout(resolve, 2000));
+
```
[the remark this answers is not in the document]
MRayermannMSFT: Fixed — replaced the sleep with a deterministic `Promise` that resolves on the `commands.changed` event.

nodejs/test/e2e/commands.test.ts
```diff
@@ -0,0 +1,57 @@
...
+describe("Commands", async () => {
+    // Use TCP mode so a second client can connect to the same CLI process
+    const ctx = await createSdkTestContext({ useStdio: false });
+    const client1 = ctx.copilotClient;
+
+    // Trigger connection so we can read the port
+    const initSession = await client1.createSession({ onPermissionRequest: approveAll });
+    await initSession.disconnect();
+
+    const actualPort = (client1 as unknown as { actualPort: number }).actualPort;
```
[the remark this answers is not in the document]
MRayermannMSFT: This is the same pattern used in `multi-client.test.ts` (line 20) — `actualPort` is private on `CopilotClient` with no public accessor, and `createSdkTestContext` doesn't expose it. Leaving as-is for consistency; can revisit if we add a public getter.

nodejs/test/e2e/ui_elicitation.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+describe("UI Elicitation", async () => {
+    const { copilotClient: client } = await createSdkTestContext();
+
+    it("elicitation methods throw in headless mode", async () => {
+        const session = await client.createSession({
+            onPermissionRequest: approveAll,
+        });
+
+        // The SDK spawns the CLI headless - no TUI means no elicitation support.
```
MRayermannMSFT, line 17: E2E testing ability limited here since SDK users can't create their own elicitation handler. Actively working on closing that gap.

nodejs/src/session.ts
```diff
@@ -480,6 +565,108 @@ export class CopilotSession {
         return this.toolHandlers.get(name);
     }
 
+    /**
+     * Registers command handlers for this session.
+     *
+     * @param commands - An array of command definitions with handlers, or undefined to clear
+     * @internal This method is typically called internally when creating/resuming a session.
```
SteveSandersonMS, line 572: In that case, could the set of commands be a property on `SessionConfig` so that we pass them in the same way as tools, and they can only come from the original session creation/resumption options?

Would be ideal not to have any reachable APIs for setting commands dynamically unless we want people to do that.

nodejs/src/session.ts
```diff
@@ -480,6 +565,108 @@ export class CopilotSession {
...
+        for (const cmd of commands) {
+            this.commandHandlers.set(cmd.name, cmd.handler);
+        }
+    }
+
+    /**
+     * Sets the host capabilities for this session.
+     *
+     * @param capabilities - The capabilities object from the create/resume response
+     * @internal This method is typically called internally when creating/resuming a session.
```
SteveSandersonMS, line 588: Also would be good to pass this in via `SessionConfig` instead of having a reachable API for it (even if marked as `@internal`).

nodejs/test/e2e/ui_elicitation.test.ts
SteveSandersonMS: Do you want to do that as part of this PR or is it a follow-up?

MRayermannMSFT: Follow-up, the runtime work is taking a bit.

nodejs/src/session.ts
MRayermannMSFT: Commands are already on `SessionConfig`,  that's how users pass them in. The `registerCommands` is just the internal plumbing called from `client.ts` to wire the handlers onto the session object, same pattern as the existing `registerTools`.

nodejs/src/session.ts
MRayermannMSFT: Same situation as pior comment, this mirrors how `registerTools` works. The `client.ts` calls it after getting the create/resume response.

nodejs/src/session.ts
SteveSandersonMS: Got it. Thanks. Maybe we could find a way to change this for both of them, but no reason to block this PR.

MRayermannMSFT: I'll try to be a good citizen and come back to this. 😄

PR #960, opened 2026-03-31
Close Language Gaps for Commands + Dialogs/Elicitations

MRayermannMSFT:
## What

Adds Commands and UI Elicitation support to the Python, Go, and .NET SDKs, closing the feature gap with the Node.js reference implementation. Each SDK now supports registering slash commands, handling elicitation requests from the server, and driving interactive form dialogs (confirm, select, input) through a capability-gated UI API — with full documentation, unit tests, and E2E tests matching the Node.js test surface.

## Why

These two features were recently shipped in the Node.js SDK only, leaving the other three languages unable to participate in command dispatch or elicitation flows. Any multi-language deployment that relied on commands or form-based user input would have been forced to use the Node.js SDK exclusively. Closing this gap ensures all four SDKs offer the same capabilities, preventing fragmentation and letting teams choose the language that fits their stack.

## ⚠️ Breaking Change (Node.js only)

The Node.js `ElicitationHandler` signature has changed from two arguments to one. The separate `ElicitationRequest` and `{ sessionId }` invocation types have been merged into a single `ElicitationContext` type that includes `sessionId` alongside the request fields. This mirrors the existing `CommandContext` single-argument pattern.

**Before:**
```typescript
onElicitationRequest: async (request, invocation) => {
    console.log(invocation.sessionId, request.message);
}
```

**After:**
```typescript
onElicitationRequest: async (context) => {
    console.log(context.sessionId, context.message);
}
```

Python, Go, and .NET are not affected — Commands and Elicitation are new APIs in those SDKs and ship with the `ElicitationContext` single-argument pattern from the start.

### DIFF
--- a/nodejs/src/types.ts
+++ b/nodejs/src/types.ts
@@ -421,10 +421,6 @@ export interface ElicitationParams {
 }
 
 /**
- * Request payload passed to an elicitation handler callback.
- * Extends ElicitationParams with optional metadata fields.
- */
-/**
  * Context for an elicitation handler invocation, combining the request data
  * with session context. Mirrors the single-argument pattern of {@link CommandContext}.
  */


go/session.go
```diff
@@ -586,6 +921,62 @@ func (s *Session) handleBroadcastEvent(event SessionEvent) {
...
+		if handler == nil {
+			return
+		}
+		message := ""
+		if event.Data.Message != nil {
+			message = *event.Data.Message
+		}
+		var requestedSchema map[string]any
+		if event.Data.RequestedSchema != nil {
+			requestedSchema = event.Data.RequestedSchema.Properties
```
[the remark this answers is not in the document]
MRayermannMSFT: Fixed in ec612a4. Now passes the full schema (type, properties, required) as a map instead of just Properties. Added unit test TestSession_ElicitationRequestSchema that verifies all three fields are passed through to the handler. Also added equivalent test in Python.

go/session.go
```diff
@@ -498,6 +505,334 @@ func (s *Session) handleSystemMessageTransform(sections map[string]systemMessage
...
+		return "", false, err
+	}
+	rpcResult, err := ui.session.RPC.Ui.Elicitation(ctx, &rpc.SessionUIElicitationParams{
+		Message: message,
+		RequestedSchema: rpc.RequestedSchema{
+			Type: rpc.RequestedSchemaTypeObject,
+			Properties: map[string]rpc.Property{
+				"selection": {
+					Type: rpc.PropertyTypeString,
+					Enum: options,
```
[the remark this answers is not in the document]
MRayermannMSFT: Acknowledged. The `default: null` comes from the generated `rpc.Property` struct (`json:"default"` without `omitempty`). Since this is codegen output we cannot add `omitempty` without changing the codegen pipeline. In practice, `"default": null` is valid JSON Schema (means no default) and the CLI server handles it correctly. The proper fix would be in the codegen (`scripts/codegen/go.ts`). Leaving as-is to avoid touching codegen in this PR.

go/session_test.go
```diff
@@ -204,3 +205,192 @@ func TestSession_On(t *testing.T) {
...
+		}
+	})
+}
+
+func containsString(s, substr string) bool {
+	return len(s) >= len(substr) && searchSubstring(s, substr)
+}
+
+func searchSubstring(s, substr string) bool {
+	for i := 0; i <= len(s)-len(substr); i++ {
```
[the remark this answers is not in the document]
MRayermannMSFT: Fixed in ec612a4. Replaced custom `containsString`/`searchSubstring` helpers with `strings.Contains` from the standard library.

dotnet/src/Session.cs
```diff
@@ -98,6 +100,30 @@ public sealed partial class CopilotSession : IAsyncDisposable
...
+    /// <value>
+    /// An <see cref="ISessionUiApi"/> implementation with convenience methods for
+    /// confirm, select, input, and custom elicitation dialogs.
+    /// </value>
+    /// <remarks>
+    /// All methods on this property throw <see cref="InvalidOperationException"/>
+    /// if the host does not report elicitation support via <see cref="Capabilities"/>.
+    /// Check <c>session.Capabilities.Ui?.Elicitation == true</c> before calling.
+    /// </remarks>
+    public ISessionUiApi Ui => new SessionUiApiImpl(this);
```
MackinnonBuck: Two things here:

1. This allocates a new `SessionUiApiImpl` on every `.Ui` access. Could we make this a `{ get; set; }` property instead of a computed (`=>`) property?

2. `ISessionUiApi` is an interface; are we expecting there to be any implementors other than the one provided by the SDK? If not, should we remove the interface and just have a concrete type?

dotnet/src/Session.cs
```diff
@@ -436,6 +462,59 @@ private async Task HandleBroadcastEventAsync(SessionEvent sessionEvent)
...
                         break;
                     }
+
+                case CommandExecuteEvent cmdEvent:
+                    {
+                        var data = cmdEvent.Data;
+                        if (string.IsNullOrEmpty(data.RequestId))
+                            return;
+
+                        _ = ExecuteCommandAndRespondAsync(data.RequestId, data.CommandName, data.Command, data.Args);
```
MackinnonBuck: `HandleBroadcastEventAsync` is already fire-and-forget at the call site, so I don't think we need to fire-and-forget here. Any reason we can't `await` these calls normally? Same applies to `HandleElicitationRequestAsync` on line 493.

dotnet/src/Session.cs
```diff
@@ -436,6 +462,59 @@ private async Task HandleBroadcastEventAsync(SessionEvent sessionEvent)
...
+                                    Required = data.RequestedSchema.Required?.ToList()
+                                }
+                                : null;
+
+                            _ = HandleElicitationRequestAsync(
+                                new ElicitationRequest
+                                {
+                                    Message = data.Message,
+                                    RequestedSchema = schema,
+                                    Mode = data.Mode?.ToString().ToLowerInvariant(),
```
MackinnonBuck: Should `ElicitationRequest.Mode` should be `ElicitationRequestedDataMode` instead of `string`?. The generated enum already exists, is public, and handles JSON serialization correctly, so making this change would allow us to remove this `.ToString().ToLowerInvariant()` conversion.

dotnet/src/Types.cs
```diff
@@ -500,6 +501,260 @@ public class UserInputInvocation
...
+    /// <summary>
+    /// Identifier of the session that triggered the elicitation request.
+    /// </summary>
+    public string SessionId { get; set; } = string.Empty;
+}
+
+/// <summary>
+/// Delegate for handling elicitation requests from the server.
+/// </summary>
+public delegate Task<ElicitationResult> ElicitationHandler(ElicitationRequest request, ElicitationInvocation invocation);
```
MackinnonBuck: I see that we have both `ElicitationRequest` and `ElicitationInvocation` as separate arguments. Can we merge these types into a single type like `ElicitationContext`? This would be consistent with `CommandContext`. I do appreciate that this is an inconsistency already present in other places in the SDK.

go/session.go
```diff
@@ -586,6 +921,68 @@ func (s *Session) handleBroadcastEvent(event SessionEvent) {
...
+		if event.Data.URL != nil {
+			url = *event.Data.URL
+		}
+		s.handleElicitationRequest(ElicitationRequest{
+			Message:           message,
+			RequestedSchema:   requestedSchema,
+			Mode:              mode,
+			ElicitationSource: elicitationSource,
+			URL:               url,
+		}, *requestID)
```
MackinnonBuck, line 985: It looks like `handleBroadcastEvent` doesn't handle `SessionEventTypeCapabilitiesChanged`, unlike the other SDKs.

dotnet/src/Session.cs
MRayermannMSFT: Fixed in 0c4ff33. Now cached via `Lazy<ISessionUiApi>` — single allocation on first access. Keeping the `ISessionUiApi` interface for now since it aids testability (consumers can mock it), but happy to revisit if the team prefers a concrete type.

dotnet/src/Session.cs
MRayermannMSFT: Fixed in 0c4ff33. Both `ExecuteCommandAndRespondAsync` and `HandleElicitationRequestAsync` are now `await`ed inside `HandleBroadcastEventAsync`. The outer call from `DispatchEvent` is already fire-and-forget, so these inner calls get proper error propagation now.

dotnet/src/Session.cs
MRayermannMSFT: Fixed in 0c4ff33. `ElicitationRequest.Mode` is now `ElicitationRequestedDataMode?` instead of `string?`. The `.ToString().ToLowerInvariant()` conversion is removed — the generated enum value passes through directly.

dotnet/src/Types.cs
MRayermannMSFT: Good suggestion for consistency with `CommandContext`. However, this would be a public API change across all four SDKs (Python and Go also have the two-argument pattern, matching Node.js). I think this warrants its own PR to align all languages simultaneously rather than changing the contract in one SDK here. Happy to file a follow-up issue if you agree.

go/session.go
MRayermannMSFT: Good catch! Fixed in 0c4ff33. Added `SessionEventTypeCapabilitiesChanged` handling in `handleBroadcastEvent` — now calls `setCapabilities()` when the event fires, matching Python and .NET behavior. Added a test that verifies both enable and disable transitions via the event.

dotnet/src/Types.cs
MRayermannMSFT: Changing my mind. I'll take care of this.

MRayermannMSFT: Done in dff5fe1. Merged `ElicitationRequest` + `ElicitationInvocation` into a single `ElicitationContext` type across all three SDKs (Python, Go, .NET), matching the `CommandContext` single-argument pattern. Handler signatures are now `(ElicitationContext) -> ElicitationResult` everywhere. All tests, READMEs, and E2E tests updated.

MRayermannMSFT: @MackinnonBuck I decided to go ahead and make NodeJS consistent as well. Lmk if you think this type of breaking change is problematic. 😬

dotnet/src/Session.cs
```diff
@@ -98,6 +101,30 @@ public sealed partial class CopilotSession : IAsyncDisposable
...
+    /// <value>
+    /// An <see cref="ISessionUiApi"/> implementation with convenience methods for
+    /// confirm, select, input, and custom elicitation dialogs.
+    /// </value>
+    /// <remarks>
+    /// All methods on this property throw <see cref="InvalidOperationException"/>
+    /// if the host does not report elicitation support via <see cref="Capabilities"/>.
+    /// Check <c>session.Capabilities.Ui?.Elicitation == true</c> before calling.
+    /// </remarks>
+    public ISessionUiApi Ui => _ui.Value;
```
MackinnonBuck: Nit:

Apologies if my earlier comment wasn't clearer - I was mainly suggesting that this property should either have an explicit backing field or be an auto property (`{ get; }`). Otherwise an allocation is incurred for every access of the `Ui` property.

However, using `Lazy<T>` is probably overkill because the initialization cost of `SessionUiApiImpl` is negligible.

We can probably just change the `_ui` field to be a `SessionUiApiImpl` and initialize it in the constructor. Then the `Ui` property would just be:

```cs
public ISessionUiApi Ui => _ui;
```

MackinnonBuck (dismissed):
Looks great! Just a minor nit.

dotnet/src/Session.cs
MackinnonBuck: Or we could do:

```cs
public ISessionUiApi Ui { get; }
```

And initialize it in the constructor without the backing field. No preference either way!

MRayermannMSFT: No worries. I'm not a C# expert, so misunderstanding is on my side. Will fix tomorrow. :)

dotnet/test/CommandsTests.cs
```diff
@@ -0,0 +1,138 @@
...
+            OnPermissionRequest = PermissionHandler.ApproveAll,
+            Commands =
+            [
+                new CommandDefinition { Name = "deploy", Description = "Deploy", Handler = _ => Task.CompletedTask },
+            ],
+        });
+
+        Assert.NotNull(session2);
+        Assert.Equal(sessionId, session2.SessionId);
+        await session2.DisposeAsync();
```
SteveSandersonMS, line 50: These tests don't really seem to show what the name implies - it shows that we don't throw on the `ResumeSessionAsync` call but not that commands were forwarded.

Perhaps it's impractical to show that commands work e2e because they are only exposed to the CLI and we don't have a way to drive the CLI from here.

I did at one point add a whole method for observing the RPC traffic but took it back out because it adds a bunch more complexity. Maybe we should put that back in.

For now, no reason to block this PR. If we do have any way to query for the list of commands available on a session then it would be good to add assertions here, but if not we can plug this gap later.

test/snapshots/commands/forwards_commands_in_session_create.yaml
```diff
@@ -0,0 +1,4 @@
+models:
+  - claude-sonnet-4.5
+conversations: []
+
```
SteveSandersonMS: I'm curious about all these empty conversation snapshots.

Did these get created by running the tests? I didn't think we created snapshots at all unless there was traffic to capture. If all these empty snapshots were deleted, do the tests still pass?

Is there a possibility that Copilot hand-wrote these snapshots? It often tries to do that, and I have to keep asking it not to do so.

Bottom line - I'm hoping we can just delete them!

dotnet/src/Types.cs
SteveSandersonMS: Since this is a pretty new feature anyway, I think we'll be OK taking the breaking change. We just need to be sure to call it out in the release notes.

test/snapshots/commands/forwards_commands_in_session_create.yaml
MRayermannMSFT: > Did these get created by running the tests? I didn't think we created snapshots at all unless there was traffic to capture. If all these empty snapshots were deleted, do the tests still pass?

They were created by running the tests, I think. I will try deleting them nonetheless and seeing if we still pass.

MRayermannMSFT: Tests passed!

dotnet/test/CommandsTests.cs
MRayermannMSFT: Good point — renamed the tests to `Session_With_Commands_Creates_Successfully` / `Session_With_Commands_Resumes_Successfully` since they only verify the session creates without throwing. The actual verification that commands were forwarded to the server is in the multi-client test `Client_Receives_Commands_Changed_When_Another_Client_Joins_With_Commands`, which proves the server received and registered the commands by checking the `commands.changed` event on another client. There's no single-client API to query registered commands back.

python/copilot/session.py
```diff
@@ -250,6 +259,284 @@ class UserInputResponse(TypedDict):
...
+    message: Required[str]
+    """Message describing what information is needed from the user."""
+    requestedSchema: dict[str, Any]
+    """JSON Schema describing the form fields to present."""
+    mode: Literal["form", "url"]
+    """Elicitation mode: ``"form"`` for structured input, ``"url"`` for browser redirect."""
+    elicitationSource: str
+    """The source that initiated the request (e.g. MCP server name)."""
+    url: str
+    """URL to open in the browser (when mode is ``"url"``)."""
```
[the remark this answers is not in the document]
MRayermannMSFT: The camelCase field names here match the JSON-RPC wire protocol keys, which is the existing convention throughout the Python SDK for TypedDicts that map directly to wire format. For example, on main: `SelectionAttachment.filePath`, `BlobAttachment.mimeType`, `UserInputRequest.allowFreeform`, `PreToolUseHookInput.toolName`/`toolArgs`, `PostToolUseHookOutput.modifiedResult`/`additionalContext`, etc. all use camelCase. Only SDK-internal config types like `SessionConfig` and `CommandContext` use snake_case. Normalizing this across all wire-mapped TypedDicts would be a worthwhile follow-up but is a broader effort beyond this PR.

python/copilot/session.py
```diff
@@ -250,6 +259,284 @@ class UserInputResponse(TypedDict):
...
+    description: str
+    """Descriptive text shown below the field."""
+    minLength: int
+    """Minimum text length."""
+    maxLength: int
+    """Maximum text length."""
+    format: str
+    """Input format hint (e.g. ``"email"``, ``"uri"``, ``"date"``)."""
+    default: str
+    """Default value for the input field."""
```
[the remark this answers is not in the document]
MRayermannMSFT: The camelCase field names here match the JSON-RPC wire protocol keys, which is the existing convention throughout the Python SDK for TypedDicts that map directly to wire format. For example, on main: `SelectionAttachment.filePath`, `BlobAttachment.mimeType`, `UserInputRequest.allowFreeform`, `PreToolUseHookInput.toolName`/`toolArgs`, `PostToolUseHookOutput.modifiedResult`/`additionalContext`, etc. all use camelCase. Only SDK-internal config types like `SessionConfig` and `CommandContext` use snake_case. Normalizing this across all wire-mapped TypedDicts would be a worthwhile follow-up but is a broader effort beyond this PR.

python/copilot/session.py
```diff
@@ -250,6 +259,284 @@ class UserInputResponse(TypedDict):
...
+
+
+class ElicitationParams(TypedDict):
+    """Parameters for a raw elicitation request."""
+
+    message: str
+    """Message describing what information is needed from the user."""
+    requestedSchema: dict[str, Any]
+    """JSON Schema describing the form fields to present."""
+
```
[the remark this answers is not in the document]
MRayermannMSFT: The camelCase field names here match the JSON-RPC wire protocol keys, which is the existing convention throughout the Python SDK for TypedDicts that map directly to wire format. For example, on main: `SelectionAttachment.filePath`, `BlobAttachment.mimeType`, `UserInputRequest.allowFreeform`, `PreToolUseHookInput.toolName`/`toolArgs`, `PostToolUseHookOutput.modifiedResult`/`additionalContext`, etc. all use camelCase. Only SDK-internal config types like `SessionConfig` and `CommandContext` use snake_case. Normalizing this across all wire-mapped TypedDicts would be a worthwhile follow-up but is a broader effort beyond this PR.

Issue #959, opened 2026-03-30
BUG: Subagents don't inherit session-defined external tools


knifeyspoony:
Parent copilot session can define external tools and use them normally. When spawning agents via task(), these tools are not available.

repro:

1. Start a Copilot session and register one external/session tool, e.g. hello_tool.
2. Verify the parent agent can call hello_tool successfully.
3. Define a custom agent / task child whose tools allowlist includes hello_tool.
4. Ask the parent to use built-in task() to delegate to that child, with instructions: “Call hello_tool.”
5. Observe the child cannot use it: it reports the tool is unavailable, or no external-tool callback fires.

Observed on CLI 1.0.12

Issue #947, opened 2026-03-27
Custom SDK tools (defineTool) not available to sub-agents via CustomAgentConfig.tools whitelist


dt-benedict:
## Bug

When a `CustomAgentConfig` specifies custom SDK tool names (registered via `defineTool()`) in its `tools` whitelist, the sub-agent does not receive those tools at runtime. Only built-in CLI tools (like `view`) survive the filtering. The custom tool names are silently ignored.

This is the same class of issue as #860 (MCP tool names not expanded), but affects custom tools registered through the SDK's `defineTool()` API.

## Versions

- **SDK:** `@github/copilot-sdk 0.2.0`
- **Runtime:** Node.js 22

## Expected Behavior

When a session registers custom tools via `defineTool()` and a custom agent's `tools` array references those tool names, the sub-agent should have access to those custom tools.

```typescript
const myTool = defineTool("run_bash", {
  description: "Execute a bash command",
  parameters: z.object({ command: z.string() }),
  handler: async ({ command }) => { /* ... */ },
});

const session = await client.createSession({
  tools: [myTool],
  customAgents: [
    {
      name: "investigator",
      description: "Investigates problems using shell commands",
      tools: ["run_bash"],  // Should give this agent access to the custom tool
      prompt: "You are an investigator...",
    },
  ],
  onPermissionRequest: async () => ({ kind: "approved" }),
});
```

The investigator agent should have `run_bash` available.

## Actual Behavior

The sub-agent reports it does not have `run_bash` available — it only sees built-in tools like `view`. The custom tool name `"run_bash"` in the agent's `tools` array is not matched against the session's registered custom tools.

The agent correctly reports "I do not have a run_bash tool available" — this is not a hallucination, the tool genuinely isn't injected into the sub-agent's context.

## Reproduction

1. Register a custom tool via `defineTool("my_custom_tool", { ... })`
2. Pass it in `tools` when creating a session
3. Define a custom agent with `tools: ["my_custom_tool"]`
4. Send a prompt that triggers the sub-agent
5. Ask the sub-agent to use `my_custom_tool`
6. Sub-agent reports the tool is not available

## Workaround

Setting `tools: null` (all tools) on the custom agent config does give the sub-agent access to custom tools. The issue is specifically with the whitelist filtering — it only matches built-in CLI tool names, not custom SDK tool names.

[...]

PR #962, opened 2026-03-31
feat(go): support custom tools in subagents

ductrung-nguyen:
All three SDKs (Go, Python, Node) use **per-session tool handler lookup** keyed to the exact session ID. When the CLI creates child sessions for subagents, those child session IDs are never registered in the SDK's `sessions` map. Tool calls arriving with a child session ID fail with `"unknown session <id>"`.

This PR only brings the ability to call **custom tools** for **sub-agent** in Go, by following the proposal in https://github.com/github/copilot-sdk/issues/947

Related issues:
- https://github.com/github/copilot-sdk/issues/959


SteveSandersonMS:
The underlying issue was already fixed so I'll close this. Thanks @ductrung-nguyen.

PR #966, opened 2026-03-31
Add provider model and token limit overrides to ProviderConfig

MackinnonBuck:
## Summary

Extends `ProviderConfig` across all four language SDKs with optional fields for **model overrides** and **token limits**. These fields match the runtime's wire format and let BYOK users decouple the model name sent to their provider from the well-known model ID used for agent configuration and capability lookup.

### New fields

| Wire name | Node.js | Python | .NET | Go |
|-----------|---------|--------|------|----|
| `modelId` | `modelId?: string` | `model_id: str` | `string? ModelId` | `ModelID string` |
| `wireModel` | `wireModel?: string` | `wire_model: str` | `string? WireModel` | `WireModel string` |
| `maxInputTokens` | `maxInputTokens?: number` | `max_input_tokens: int` | `int? MaxInputTokens` | `MaxInputTokens int` |
| `maxOutputTokens` | `maxOutputTokens?: number` | `max_output_tokens: int` | `int? MaxOutputTokens` | `MaxOutputTokens int` |

All fields are optional.

- **`modelId`** — Well-known model ID used to look up agent configuration (tools, prompts, reasoning behavior) and default token limits from the capability catalog. Useful for fine-tunes that should inherit a base model''s configuration. Defaults to the session''s configured model (`SessionConfig.model`) when unset.
- **`wireModel`** — Model identifier sent to the provider API for inference. Use this when the name your provider knows (e.g. an Azure deployment name or custom fine-tune name) differs from the well-known model ID. Defaults to the session''s configured model when unset.
- **`maxPromptTokens`** — Maximum tokens allowed in the prompt for a single request. The runtime triggers conversation compaction before sending a request when the prompt exceeds this limit.
- **`maxOutputTokens`** — Maximum tokens the model can generate in a single response.

`modelId` and `wireModel` are independent knobs — set just one or both depending on whether you need to override config lookup, wire transmission, or both.

### Changes

- **Node.js** (`nodejs/src/types.ts`) — Added fields to `ProviderConfig` interface
- **Python** (`python/copilot/session.py`) — Added fields to `ProviderConfig` TypedDict
- **Python** (`python/copilot/client.py`) — Updated `_convert_provider_to_wire_format` to map new snake_case fields to camelCase
- **.NET** (`dotnet/src/Types.cs`) — Added nullable properties with `[JsonPropertyName]` attributes
- **Go** (`go/types.go`) — Added fields with `json:"...,omitempty"` tags

### Testing

Extended existing provider-forwarding/serialization unit tests across all four SDKs to cover the new fields:

- **Node.js** (`nodejs/test/client.test.ts`) — Existing `forwards provider headers` tests for `session.create` / `session.resume` now also assert all four new fields on the wire payload
- **Python** (`python/test_client.py`) — Same — extended `test_create_session_forwards_provider_headers` / `_resume_` to assert camelCase mapping for all four new fields
- **.NET** (`dotnet/test/Unit/SerializationTests.cs`) — Extended `ProviderConfig_CanSerializeHeaders_WithSdkOptions` to round-trip all four new fields with correct JSON property names
- **Go** (`go/types_test.go`) — Added `TestProviderConfig_JSONIncludesAllFields` (correct JSON tags) and `TestProviderConfig_JSONOmitsUnsetTokenFields` (`omitempty` works)

All four SDKs build clean (`tsc --noEmit`, `go build ./...`, `dotnet build`, `python -c "import copilot"`) and all unit tests pass.


go/types.go
```diff
@@ -601,6 +601,19 @@ type ProviderConfig struct {
...
+	// MaxPromptTokens overrides the maximum number of prompt/input tokens.
+	// When set, takes precedence over the default limit from the model's capability catalog entry.
+	MaxPromptTokens int `json:"maxPromptTokens,omitempty"`
+	// MaxContextWindowTokens overrides the maximum context window size in tokens.
+	// When set, takes precedence over the default limit from the model's capability catalog entry.
+	MaxContextWindowTokens int `json:"maxContextWindowTokens,omitempty"`
+	// ModelLimitsId specifies the model ID used to look up default token limits from the capability catalog.
+	// When unset, the session's configured model ID is used.
+	// Useful for fine-tuned models that share the same limits as a base model.
+	ModelLimitsId string `json:"modelLimitsId,omitempty"`
```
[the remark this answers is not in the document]
MackinnonBuck: Good catch — renamed to `ModelLimitsID` to follow Go initialism conventions, consistent with `APIKey`, `BaseURL`, etc.

python/copilot/client.py
```diff
@@ -1962,6 +1962,16 @@ def _convert_provider_to_wire_format(
...
+        if "max_output_tokens" in provider:
+            wire_provider["maxOutputTokens"] = provider["max_output_tokens"]
+        if "max_prompt_tokens" in provider:
+            wire_provider["maxPromptTokens"] = provider["max_prompt_tokens"]
+        if "max_context_window_tokens" in provider:
+            wire_provider["maxContextWindowTokens"] = provider[
+                "max_context_window_tokens"
+            ]
+        if "model_limits_id" in provider:
+            wire_provider["modelLimitsId"] = provider["model_limits_id"]
```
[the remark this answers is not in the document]
MackinnonBuck: I looked into this — there are no existing unit tests for `_convert_provider_to_wire_format` in the Python test suite. Adding a new test file or pattern just for this PR seems like unnecessary scope creep. The wire format mapping is straightforward (snake_case → camelCase key renaming), and the E2E tests will exercise the full path. I'd suggest we defer adding unit tests for wire format conversion to a separate effort if desired.

dotnet/src/Types.cs
```diff
@@ -1115,6 +1115,35 @@ public class ProviderConfig
...
+    [JsonPropertyName("maxContextWindowTokens")]
+    public int? MaxContextWindowTokens { get; set; }
+
+    /// <summary>
+    /// Specifies the model ID used to look up default token limits from the capability catalog.
+    /// When unset, the session's configured model ID (see <see cref="SessionConfig.Model"/>) is used.
+    /// This is useful for fine-tuned models that share the same limits as a base model.
+    /// </summary>
+    [JsonPropertyName("modelLimitsId")]
+    public string? ModelLimitsId { get; set; }
```
MackinnonBuck: Curious to know what others think about this name. We might want to consider extending this to also affect inferred model capabilities (streaming, tool calling, multimodality, etc.), at which point this name is a bit too specific.

We could consider something like `ModelCatalogId` or `ModelProfileId`.

dotnet/src/Types.cs
```diff
@@ -1115,6 +1115,35 @@ public class ProviderConfig
     /// </summary>
     [JsonPropertyName("azure")]
     public AzureOptions? Azure { get; set; }
+
+    /// <summary>
+    /// Overrides the maximum number of output tokens the model can generate.
+    /// When set, takes precedence over the default limit resolved from the model's capability catalog entry.
+    /// </summary>
+    [JsonPropertyName("maxOutputTokens")]
+    public int? MaxOutputTokens { get; set; }
```
stephentoub: What is the scope of this? Is it per request to the LLM? Total across all operations performed by the agent until idle?

dotnet/src/Types.cs
```diff
@@ -1115,6 +1115,35 @@ public class ProviderConfig
...
+
+    /// <summary>
+    /// Overrides the maximum number of prompt/input tokens.
+    /// When set, takes precedence over the default limit resolved from the model's capability catalog entry.
+    /// </summary>
+    [JsonPropertyName("maxPromptTokens")]
+    public int? MaxPromptTokens { get; set; }
```
stephentoub: Should this be called MaxInputTokens instead of MaxPromptTokens?

Does this include cached tokens?

Same question as above... is this about one request or across a sequence of calls?

dotnet/src/Types.cs
```diff
@@ -1115,6 +1115,35 @@ public class ProviderConfig
...
+
+    /// <summary>
+    /// Overrides the maximum context window size in tokens.
+    /// When set, takes precedence over the default limit resolved from the model's capability catalog entry.
+    /// </summary>
+    [JsonPropertyName("maxContextWindowTokens")]
+    public int? MaxContextWindowTokens { get; set; }
```
stephentoub: What happens when these Max values are hit?

dotnet/src/Types.cs
MackinnonBuck: Per individual LLM API request. This value is sent as `max_tokens` in every call to the model provider. Each turn in the agent loop independently gets this cap, and it's not accumulated across the session. It controls how many tokens the model can generate in a single response (text, tool calls, reasoning) before control returns.

dotnet/src/Types.cs
MackinnonBuck: Per-request, but not sent to the API. The runtime uses this internally to decide when to truncate or compact conversation history before each LLM call. "Prompt" here means everything sent to the model in one request: system message, full conversation history up to that point, tool definitions, and the new user message. Cached tokens are counted toward the limit.

The name matches the upstream CAPI `/models` field (`max_prompt_tokens`), though `MaxInputTokens` would also be reasonable.

dotnet/src/Types.cs
MackinnonBuck: Each limit is handled differently:

* `MaxOutputTokens` hit: The model stops generating and returns `stop_reason: "max_tokens"`. Currently the runtime treats the truncated response as normal (silent failure, but I'm going to propose a fix for that soon).
* `MaxPromptTokens` hit: The runtime triggers background compaction before sending the request.
* `MaxContextWindowTokens` hit: Largely a fallback for `MaxPromptTokens`. The runtime prefers `max_prompt_tokens` for truncation/compaction decisions and only falls back to `max_context_window_tokens` when the former isn't set. If the API still rejects the request, the runtime adjusts its stored limit downward and retries. It's probably fine to not expose this value as public API in the SDK (we don't make it configurable for BYOK in the CLI).

MackinnonBuck: Update: Removed `MaxContextWindowTokens`.

dotnet/src/Types.cs
MackinnonBuck: Updated XML docs to clarify

dotnet/src/Types.cs
SteveSandersonMS: `ModelCatalogId` is good except if you think it's identifying the catalog rather than the model. `ModelIdFromCatalog` would be more explicit but feels clunky.

Conceptually I like it though. The only alternative I can think of is adding a different property `ModelDeploymentId` and then having the rules:

 * If `ModelId` and `ModelDeploymentId` are both set then we use the deployment ID when identifying the model in inference requests, and the model ID when looking up capabilities/limits etc
 * If only one is set then we use that for both things

What do you think? Main drawback would be if neither is set, we end up with a cumbersome error like "At least one of ModelId or ModelDeploymentId is required".

Or we could keep ModelId as required, and say ModelDeploymentId is the optional override on inference requests.

MackinnonBuck: Good ideas!

I think the right separation would be:
* `Model` means "the model used in inference requests"
* `ModelCapabilitiesId` (or something) means "the model used to look up capabilities and limits"

That way, all configuration around BYOK model limits is in the same place (`ProviderConfig`) rather than split between `ProviderConfig` and `SessionConfig`.

That said, we've also had requests to support multiple models in BYOK mode (for the CLI specifically, but we should probably expose similar functionality via the SDK). So I'm wondering if it would actually make sense to allow configuring a list of model configurations in `ProviderConfig`, where each model config defines both the ID used in inference as well as the ID used to look up model capabilities.

I'll mark this PR as a draft while I think about how such a design could look.

rramos-seidor:
Hi @MackinnonBuck! Just wanted to follow up on this PR — is it still pending review? Adding token limit fields to ProviderConfig across all SDKs seems like a key capability for BYOK users who need to configure custom providers with specific limits. Would be great to see this move forward. Is there anything blocking it from being marked as ready for review? Thanks.

stephentoub:
@MackinnonBuck, are you still working on this?

MackinnonBuck:
@stephentoub Thanks for the ping. I was holding off on this because we've had requests on the CLI side to allow for multiple provider configs / models per provider config, and I wanted to make sure we were properly accounting for that on the SDK APIs. However, adding new fields to the existing provider config SDK type probably doesn't make it any harder for us to adjust the API later, so it's probably fine to take this now. I'll work on getting this in a ready-to-review state tomorrow.

dotnet/src/Types.cs
```diff
@@ -1528,6 +1528,43 @@ public class ProviderConfig
...
+    /// <summary>
+    /// Well-known model ID used to look up agent configuration (tools, prompts,
+    /// reasoning behavior) and default token limits from the capability catalog.
+    /// Useful for fine-tuned models that should inherit the configuration of a
+    /// known base model.
+    /// Defaults to the session's configured model (see <see cref="SessionConfig.Model"/>)
+    /// when not explicitly set.
+    /// </summary>
+    [JsonPropertyName("modelId")]
+    public string? ModelId { get; set; }
```
stephentoub, line 1539: So there's:
SessionConfig.Model
ProviderConfig.ModelId
ProviderConfig.WireModel

Help me understand the relationship? If I specify SessionConfig.Model, it's used as the default for both options on ProviderConfig, and those options on provider config then represent the two different groupings in which a model would be used, such that I can override one of them? Is there any situation where I would specify the same model ID for both ModelId and WireModel?

dotnet/src/Types.cs
```diff
@@ -1528,6 +1528,43 @@ public class ProviderConfig
...
+    /// <summary>
+    /// Model identifier sent to the provider API for inference.
+    /// Use this when the name your provider knows (e.g. an Azure deployment name
+    /// or a custom fine-tune name) differs from the well-known model ID used for
+    /// configuration lookup.
+    /// Defaults to the session's configured model (see <see cref="SessionConfig.Model"/>)
+    /// when not explicitly set.
+    /// </summary>
+    [JsonPropertyName("wireModel")]
+    public string? WireModel { get; set; }
```
stephentoub, line 1548: Out of the three:
SessionConfig.Model
ProviderConfig.ModelId
ProviderConfig.WireModel

what's the meaning behind ProviderConfig.ModelId having an "Id" suffix and the other two not?

dotnet/src/Types.cs
stephentoub: I'd prefer MaxInputTokens. That's the more modern terminology, right? And it maps to what we show in the CLI UI?

dotnet/src/Types.cs
```diff
@@ -1528,6 +1528,43 @@ public class ProviderConfig
...
+    [JsonPropertyName("maxPromptTokens")]
+    public int? MaxPromptTokens { get; set; }
+
+    /// <summary>
+    /// Maximum number of tokens the model can generate in a single response.
+    /// When hit, the model stops generating and returns a truncated response.
+    /// </summary>
+    [JsonPropertyName("maxOutputTokens")]
+    public int? MaxOutputTokens { get; set; }
 }
```
stephentoub, line 1565: Is it possible to add some E2E tests that cover all of these?

stephentoub (approved):

dotnet/src/Types.cs
MackinnonBuck: If only `ProviderConfig.ModelId` is specified, then that controls multiple things:
* The model name used by the runtime to determine model limits and model-specific agent configuration
* The model name sent to the custom provider for inference

If the model provider recognizes a model name that doesn't match the model ID known by the runtime, then `ProviderConfig.WireModel` can specify that.

`SessionConfig.Model` acts a default in case neither option is specified.

> Is there any situation where I would specify the same model ID for both ModelId and WireModel?

It has the same effect as just specifying `ModelId`, so it's not really necessary.

dotnet/src/Types.cs
MackinnonBuck: I figured the "ID" more strongly implied that the value was identifying a well-known model kind. We could also consider `ModelFamily`? It would just require changing the runtime as well.

dotnet/src/Types.cs
MackinnonBuck: Added!

dotnet/src/Types.cs
MackinnonBuck: We could change this; it would just require changing the runtime representation as well (including the `COPILOT_PROVIDER_MAX_PROMPT_TOKENS` environment variable, which would probably need to be renamed).

stephentoub: Changing the sdk name wouldn't, right? Only if we also wanted to change the wire name?

MackinnonBuck: Oh, if we just changed the public API but configured it to serialize with the `maxInputTokens` naming? Yeah that would work.

patniko (approved):

PR #986, opened 2026-04-02
feat: support custom headers for model backends

### LINKED ISSUE #355: Passing custom headers to model and tool backends.
Is it possible to pass custom header that LLM gateways (AIGateways might need) - copilot says its not 

The Copilot SDK only supports `ApiKey` (sent as Bearer) and `BearerToken` (also Bearer). There's no way to send custom headers like                   │
│ `Ocp-Apim-Subscription-Key`.

patniko:
## Summary

Adds custom headers support across all 4 SDKs (Node.js, Python, Go, .NET), enabling integrators to pass custom HTTP headers to model/tool backends at session creation and between turns.

#### Types (all 4 SDKs)
- `headers?: Record<string, string>` on `ProviderConfig` — session-level headers
- `HeaderMergeStrategy` type (`"override" | "merge"`)
- `requestHeaders` and `headerMergeStrategy` on `MessageOptions`/`send()` — per-turn headers

#### Session Methods (all 4 SDKs)
- `send()` updated to pass `requestHeaders` and `headerMergeStrategy`
- New `updateProvider()` / `update_provider()` / `UpdateProviderAsync()` method for mid-session provider config updates

#### Documentation
- Comprehensive "Custom Headers" section in `docs/auth/byok.md` with use cases, examples, merge strategies, env var expansion, security notes
- Custom headers examples in all 4 SDK READMEs

### Header Merge Strategies
| Strategy | Behavior |
|----------|----------|
| `"override"` (default) | Per-turn headers completely replace session-level headers |
| `"merge"` | Per-turn headers merge with session-level; per-turn wins on conflicts |

### Tests
- **Node.js**: 8 tests — send RPC params, updateProvider
- **Python**: 14 tests — types, send, updateProvider, wire format
- **Go**: 12 subtests — JSON marshaling, types, send request
- **.NET**: 12 tests — properties, cloning, all header fields

### Related
- Runtime PR: https://github.com/github/copilot-agent-runtime/pull/5679
- Closes #355


docs/auth/byok.md
```diff
@@ -304,6 +305,327 @@ provider: {
...
+
+</details>
+
+### Header Merge Strategy
+
+When you provide both session-level `headers` and per-turn `requestHeaders`, the `headerMergeStrategy` controls how they combine.
+
+| Strategy | Behavior |
+|----------|----------|
+| `"override"` (default) | Per-turn headers **completely replace** session-level headers. No session headers are sent for that turn. This is the safest default — no unexpected header leakage. |
```
stephentoub, line 487: This is an interesting default. I'm curious what drives this. What are the scenarios we have for each of override and merge that drove this choice?

docs/auth/byok.md
```diff
@@ -304,6 +305,327 @@ provider: {
...
+| Language | Field |
+|----------|-------|
+| TypeScript | `headerMergeStrategy: "override" \| "merge"` |
+| Python | `header_merge_strategy: Literal["override", "merge"]` |
+| Go | `HeaderMergeStrategy: copilot.HeaderMergeStrategyOverride \| copilot.HeaderMergeStrategyMerge` |
+| C# | `HeaderMergeStrategy = HeaderMergeStrategy.Override \| HeaderMergeStrategy.Merge` |
+
+### Updating Provider Configuration Mid-Session
+
+Use `updateProvider()` to change provider configuration — including headers — between turns without recreating the session. This is useful for rotating API keys, switching tenants, or adjusting gateway headers on the fly.
```
stephentoub, line 528: "between turns"....

If I Send and then immediately UpdateProvider, do those updates affect the current turn (which might make many LLM requests well after that), or are they delayed taking effect until idle?

docs/auth/byok.md
```diff
@@ -304,6 +305,327 @@ provider: {
...
+</details>
+
+### Environment Variable Expansion
+
+Header values support environment variable expansion at the runtime level. This lets you reference secrets without hardcoding them in your application code.
+
+| Syntax | Behavior |
+|--------|----------|
+| `${VAR}` | Replaced with the value of `VAR`. Fails if `VAR` is not set. |
+| `$VAR` | Same as `${VAR}`. |
```
stephentoub, line 605: Why two different syntaxes for the same thing?

docs/auth/byok.md
```diff
@@ -304,6 +305,327 @@ provider: {
...
+| `${VAR:-default}` | Replaced with the value of `VAR`, or `default` if `VAR` is not set. |
```
stephentoub, line 606: What's the reasoning for why the `-` is necessary after the colon?

docs/auth/byok.md
```diff
@@ -304,6 +305,327 @@ provider: {
...
+        "X-Tenant-Id": "${X_TENANT:-default-tenant}",
+    },
+}
+```
+
+> **Note:** Expansion is performed by the CLI server, not the SDK client. The SDK passes header values as-is to the server, which resolves environment variables before sending requests to your provider.
+
+### Security Considerations
+
+- **Scoped to your endpoint** — Custom headers are sent only to the configured `baseUrl`. They are never sent to GitHub Copilot servers or other endpoints.
```
stephentoub, line 625: Is there any situation where you might be able to connect to a CLI server and use this to exfiltrate environment variable values you may not have otherwise had access to?

docs/auth/byok.md
SteveSandersonMS: This is bash syntax for variable-or-default. Looks a bit weird to me too but probably will make sense to a lot of users.

docs/auth/byok.md
SteveSandersonMS: Generally the threat model is that anyone who can connect to the CLI server as the SDK has full access to the whole machine, since you can always create sessions that have auto-approve-everything, don't set any filesystem virtualization, and then instruct the agent to do anything.

This doesn't mean that users who have sessions can do anything - it comes down to who can control the session options.

SteveSandersonMS: Having said that, we should clarify that these headers are only sent when the runtime issues LLM inference calls, and not in general for all HTTP traffic. If it were the latter, an end user who *doesn't* control session options might be able to tell the agent to do an HTTP request to a URL within `baseUrl` that has some vulnerability (e.g., trigger an error and get back a dump of all incoming headers).

PR #1043, opened 2026-04-08
Naming improvements via codegen changes

SteveSandersonMS:
Changes to RPC naming convention changes to fix https://github.com/github/copilot-sdk/issues/1035

This PR improves generated public type names across the SDKs by moving naming decisions into the shared schema/codegen pipeline instead of relying on language-specific quirks.

The goal is to make generated names:

- **shorter when the old names were just path concatenation noise**
- **more specific when the old names were too generic to be useful**
- **more consistent across TypeScript, C#, Go, and Python**
- **more stable**, because naming now follows a common set of rules plus explicit schema-local overrides where the API really wants a specific public name

## Main naming rules

The naming is now mostly mechanical rather than hand-curated:

1. **Derive names from the RPC method / event identity first**
   - Example: `account.getQuota` becomes an `AccountQuota` root instead of `AccountGetQuotaResult`.

2. **Drop generic namespace and filler words**
   - Strip prefixes like `session.`
   - Strip terminal filler verbs like `get`, `read`, and `list` when they are only structural

3. **Singularize list entities**
   - Example: `models.list` uses `Model` for nested item types and `ModelList` for the top-level result

4. **Keep top-level wrapper words only when they are actually useful**
   - `Request` / `Result` remain for top-level params/results
   - nested helper types stop inheriting repeated `...Request...Result...Value...Item` noise

5. **Drop purely structural wrapper property names in nested helpers**
   - Words like `data`, `result`, `request`, `response`, `kind`, `value`, etc. are removed when they are just containers, not meaningful domain terms

6. **Normalize overlapping words**
   - Avoid awkward duplication like `requested` + `request`

7. **Use explicit schema-local names for true API surface decisions**
   - When the API really wants a specific public name, that name is declared in the schema instead of in generator-side override maps

### Previously too long

| Before | After |
| --- | --- |
| `AccountGetQuotaResultQuotaSnapshotsValue` | `AccountQuotaSnapshot` |
| `SessionPermissionsHandlePendingPermissionRequestResult` | `PermissionRequestResult` |
| `SessionToolsHandlePendingToolCallResult` | `HandleToolCallResult` |
| `SessionUiElicitationRequestRequestedSchema` | `UiElicitationSchema` |
| `SessionUiHandlePendingElicitationRequestResult` | `UiElicitationResponse` |
| `SessionUiHandlePendingElicitationResult` | `UiElicitationResult` |
| `ToolExecutionCompleteDataResultContentsItemResourceLinkIconsItemTheme` | `ToolExecutionCompleteContentResourceLinkIconTheme` |
| `UserMessageDataAttachmentsItemSelectionSelectionEnd` | `UserMessageAttachmentSelectionDetailsEnd` |

### Previously too short / too generic

| Before | After |
| --- | --- |
| `Entry` | `SessionFsReaddirWithTypesEntry` / `SessionFSReaddirWithTypesEntry` |
| `End` | `UserMessageAttachmentSelectionDetailsEnd` |
| `Mode` | `SessionMode` |
| `SessionModeGetResultMode` | `SessionMode` |

## A few especially notable cases

- The worst "path explosion" example was the old icon theme helper name:
  - `ToolExecutionCompleteDataResultContentsItemResourceLinkIconsItemTheme`
  - now: `ToolExecutionCompleteContentResourceLinkIconTheme`

- The worst "too generic to understand in isolation" examples were names like:
  - `Entry`
  - `End`
  - `Mode`

  Those now carry enough domain context to be understandable in generated SDKs without needing to inspect the surrounding method/property path.

## API-specific naming that is now explicit

A few names are intentionally chosen as API surface names rather than just synthesized mechanically, for example:

- `UiElicitationResponse`
- `UiElicitationResult`
- `HandlePendingElicitationRequest`
- `PermissionDecision`
- `PermissionDecisionRequest`
- `PermissionRequestResult`
- `ModeGetResult`
- `ModeSetResult`
- `HandleToolCallResult`

That keeps the generator logic generic while still letting the schema declare clear public names where needed.

## TypeScript note

TypeScript still avoids exploding the surface area with every synthesized helper as a named export. Explicit public names are preserved where intended, while synthetic helper naming is mainly used to improve the non-TS SDKs and shared generation consistency.

### DIFF
--- a/dotnet/src/Generated/Rpc.cs
+++ b/dotnet/src/Generated/Rpc.cs
@@ -1081,11 +1081,11 @@ internal sealed class CommandsHandlePendingCommandRequest
 }
 
 /// <summary>The elicitation response (accept with form values, decline, or cancel).</summary>
-public sealed class UiElicitationResponse
+public sealed class UIElicitationResponse
 {
     /// <summary>The user's response: accept (submitted), decline (rejected), or cancel (dismissed).</summary>
     [JsonPropertyName("action")]
-    public ElicitationResponseAction Action { get; set; }
+    public UIElicitationResponseAction Action { get; set; }
 
     /// <summary>The form values submitted by the user (present when action is 'accept').</summary>
     [JsonPropertyName("content")]
@@ -1093,7 +1093,7 @@ public sealed class UiElicitationResponse
 }
 
 /// <summary>JSON Schema describing the form fields to present to the user.</summary>
-public sealed class UiElicitationSchema
+public sealed class UIElicitationSchema
 {
     /// <summary>Schema type indicator (always 'object').</summary>
     [JsonPropertyName("type")]
@@ -1108,8 +1108,8 @@ public sealed class UiElicitationSchema
     public IList<string>? Required { get; set; }
 }
 
-/// <summary>RPC data type for UiElicitation operations.</summary>
-internal sealed class UiElicitationRequest
+/// <summary>RPC data type for UIElicitation operations.</summary>
+internal sealed class UIElicitationRequest
 {
     /// <summary>Target session identifier.</summary>
     [JsonPropertyName("sessionId")]
@@ -1121,19 +1121,19 @@ internal sealed class UiElicitationRequest
 
     /// <summary>JSON Schema describing the form fields to present to the user.</summary>
     [JsonPropertyName("requestedSchema")]
-    public UiElicitationSchema RequestedSchema { get => field ??= new(); set; }
+    public UIElicitationSchema RequestedSchema { get => field ??= new(); set; }
 }
 
-/// <summary>RPC data type for UiElicitation operations.</summary>
-public sealed class UiElicitationResult
+/// <summary>RPC data type for UIElicitation operations.</summary>
+public sealed class UIElicitationResult
 {
     /// <summary>Whether the response was accepted. False if the request was already resolved by another client.</summary>
     [JsonPropertyName("success")]
     public bool Success { get; set; }
 }
 
-/// <summary>RPC data type for HandlePendingElicitation operations.</summary>
-internal sealed class HandlePendingElicitationRequest
+/// <summary>RPC data type for UIHandlePendingElicitation operations.</summary>
+internal sealed class UIHandlePendingElicitationRequest
 {
     /// <summary>Target session identifier.</summary>
     [JsonPropertyName("sessionId")]
@@ -1145,7 +1145,7 @@ internal sealed class HandlePendingElicitationRequest
 
     /// <summary>The elicitation response (accept with form values, decline, or cancel).</summary>
     [JsonPropertyName("result")]
-    public UiElicitationResponse Result { get => field ??= new(); set; }
+    public UIElicitationResponse Result { get => field ??= new(); set; }
 }
 
 /// <summary>RPC data type for PermissionRequest operations.</summary>
@@ -1808,8 +1808,8 @@ public enum ExtensionStatus
 
 
 /// <summary>The user's response: accept (submitted), decline (rejected), or cancel (dismissed).</summary>
-[JsonConverter(typeof(JsonStringEnumConverter<ElicitationResponseAction>))]
-public enum ElicitationResponseAction
+[JsonConverter(typeof(JsonStringEnumConverter<UIElicitationResponseAction>))]
+public enum UIElicitationResponseAction
 {
     /// <summary>The <c>accept</c> variant.</summary>
     [JsonStringEnumMemberName("accept")]
@@ -2490,17 +2490,17 @@ public sealed class UiApi
     }
 
     /// <summary>Calls "session.ui.elicitation".</summary>
-    public async Task<UiElicitationResponse> ElicitationAsync(string message, UiElicitationSchema requestedSchema, CancellationToken cancellationToken = default)
+    public async Task<UIElicitationResponse> ElicitationAsync(string message, UIElicitationSchema requestedSchema, CancellationToken cancellationToken = default)
     {
-        var request = new UiElicitationRequest { SessionId = _sessionId, Message = message, RequestedSchema = requestedSchema };
-        return await CopilotClient.InvokeRpcAsync<UiElicitationResponse>(_rpc, "session.ui.elicitation", [request], cancellationToken);
+        var request = new UIElicitationRequest { SessionId = _sessionId, Message = message, RequestedSchema = requestedSchema };
+        return await CopilotClient.InvokeRpcAsync<UIElicitationResponse>(_rpc, "session.ui.elicitation", [request], cancellationToken);
     }
 
     /// <summary>Calls "session.ui.handlePendingElicitation".</summary>
-    public async Task<UiElicitationResult> HandlePendingElicitationAsync(string requestId, UiElicitationResponse result, CancellationToken cancellationToken = default)
+    public async Task<UIElicitationResult> HandlePendingElicitationAsync(string requestId, UIElicitationResponse result, CancellationToken cancellationToken = default)
     {
-        var request = new HandlePendingElicitationRequest { SessionId = _sessionId, RequestId = requestId, Result = result };
-        return await CopilotClient.InvokeRpcAsync<UiElicitationResult>(_rpc, "session.ui.handlePendingElicitation", [request], cancellationToken);
+        var request = new UIHandlePendingElicitationRequest { SessionId = _sessionId, RequestId = requestId, Result = result };
+        return await CopilotClient.InvokeRpcAsync<UIElicitationResult>(_rpc, "session.ui.handlePendingElicitation", [request], cancellationToken);
     }
 }
 
@@ -2769,7 +2769,6 @@ public static class ClientSessionApiRegistration
 [JsonSerializable(typeof(ExtensionsEnableRequest))]
 [JsonSerializable(typeof(FleetStartRequest))]
 [JsonSerializable(typeof(FleetStartResult))]
-[JsonSerializable(typeof(HandlePendingElicitationRequest))]
 [JsonSerializable(typeof(HandleToolCallResult))]
 [JsonSerializable(typeof(HistoryCompact))]
 [JsonSerializable(typeof(HistoryCompactContextWindow))]
@@ -2858,10 +2857,11 @@ public static class ClientSessionApiRegistration
 [JsonSerializable(typeof(ToolList))]
 [JsonSerializable(typeof(ToolsHandlePendingToolCallRequest))]
 [JsonSerializable(typeof(ToolsListRequest))]
-[JsonSerializable(typeof(UiElicitationRequest))]
-[JsonSerializable(typeof(UiElicitationResponse))]
-[JsonSerializable(typeof(UiElicitationResult))]
-[JsonSerializable(typeof(UiElicitationSchema))]
+[JsonSerializable(typeof(UIElicitationRequest))]
+[JsonSerializable(typeof(UIElicitationResponse))]
+[JsonSerializable(typeof(UIElicitationResult))]
+[JsonSerializable(typeof(UIElicitationSchema))]
+[JsonSerializable(typeof(UIHandlePendingElicitationRequest))]
 [JsonSerializable(typeof(UsageMetrics))]
 [JsonSerializable(typeof(UsageMetricsCodeChanges))]
 [JsonSerializable(typeof(UsageMetricsModelMetric))]
--- a/dotnet/src/Generated/Rpc.cs
+++ b/dotnet/src/Generated/Rpc.cs
@@ -29,11 +29,11 @@ public class PingResult
 
     /// <summary>Server timestamp in milliseconds.</summary>
     [JsonPropertyName("timestamp")]
-    public double Timestamp { get; set; }
+    public long Timestamp { get; set; }
 
     /// <summary>Server protocol version number.</summary>
     [JsonPropertyName("protocolVersion")]
-    public double ProtocolVersion { get; set; }
+    public long ProtocolVersion { get; set; }
 }
 
 /// <summary>RPC data type for Ping operations.</summary>
@@ -64,28 +64,33 @@ public class ModelCapabilitiesLimitsVision
     public IList<string> SupportedMediaTypes { get => field ??= []; set; }
 
     /// <summary>Maximum number of images per prompt.</summary>
+    [Range(typeof(long), "1", "long.MaxValue")]
     [JsonPropertyName("max_prompt_images")]
-    public double MaxPromptImages { get; set; }
+    public long MaxPromptImages { get; set; }
 
     /// <summary>Maximum image size in bytes.</summary>
+    [Range(typeof(long), "0", "long.MaxValue")]
     [JsonPropertyName("max_prompt_image_size")]
-    public double MaxPromptImageSize { get; set; }
+    public long MaxPromptImageSize { get; set; }
 }
 
 /// <summary>Token limits for prompts, outputs, and context window.</summary>
 public class ModelCapabilitiesLimits
 {
     /// <summary>Maximum number of prompt/input tokens.</summary>
+    [Range(typeof(long), "0", "long.MaxValue")]
     [JsonPropertyName("max_prompt_tokens")]
-    public double? MaxPromptTokens { get; set; }
+    public long? MaxPromptTokens { get; set; }
 
     /// <summary>Maximum number of output/completion tokens.</summary>
+    [Range(typeof(long), "0", "long.MaxValue")]
     [JsonPropertyName("max_output_tokens")]
-    public double? MaxOutputTokens { get; set; }
+    public long? MaxOutputTokens { get; set; }
 
     /// <summary>Maximum total context window size in tokens.</summary>
+    [Range(typeof(long), "0", "long.MaxValue")]
     [JsonPropertyName("max_context_window_tokens")]
-    public double MaxContextWindowTokens { get; set; }
+    public long MaxContextWindowTokens { get; set; }
 
     /// <summary>Vision-specific limits.</summary>
     [JsonPropertyName("vision")]
@@ -156,12 +161,12 @@ public class Model
     public string? DefaultReasoningEffort { get; set; }
 }
 
-/// <summary>RPC data type for ModelsList operations.</summary>
-public class ModelsListResult
+/// <summary>RPC data type for ModelList operations.</summary>
+public class ModelList
 {
     /// <summary>List of available models with full metadata.</summary>
     [JsonPropertyName("models")]
-    public IList<Model> Models { get => field ??= []; set; }
+    public List<Model> Models { get; set; }
 }
 
 /// <summary>RPC data type for Tool operations.</summary>
@@ -188,12 +193,12 @@ public class Tool
     public string? Instructions { get; set; }
 }
 
-/// <summary>RPC data type for ToolsList operations.</summary>
-public class ToolsListResult
+/// <summary>RPC data type for ToolList operations.</summary>
+public class ToolList
 {
     /// <summary>List of available built-in tools with metadata.</summary>
     [JsonPropertyName("tools")]
-    public IList<Tool> Tools { get => field ??= []; set; }
+    public List<Tool> Tools { get; set; }
 }
 
 /// <summary>RPC data type for ToolsList operations.</summary>
@@ -204,24 +209,26 @@ internal class ToolsListRequest
     public string? Model { get; set; }
 }
 
-/// <summary>RPC data type for AccountGetQuotaResultQuotaSnapshotsValue operations.</summary>
-public class AccountGetQuotaResultQuotaSnapshotsValue
+/// <summary>RPC data type for AccountQuotaSnapshot operations.</summary>
+public class AccountQuotaSnapshot
 {
     /// <summary>Number of requests included in the entitlement.</summary>
     [JsonPropertyName("entitlementRequests")]
-    public double EntitlementRequests { get; set; }
+    public long EntitlementRequests { get; set; }
 
     /// <summary>Number of requests used so far this period.</summary>
+    [Range(typeof(long), "0", "long.MaxValue")]
     [JsonPropertyName("usedRequests")]
-    public double UsedRequests { get; set; }
+    public long UsedRequests { get; set; }
 
     /// <summary>Percentage of entitlement remaining.</summary>
     [JsonPropertyName("remainingPercentage")]
     public double RemainingPercentage { get; set; }
 
     /// <summary>Number of overage requests made this period.</summary>
+    [Range(typeof(long), "0", "long.MaxValue")]
     [JsonPropertyName("overage")]
-    public double Overage { get; set; }
+    public long Overage { get; set; }
 
     /// <summary>Whether pay-per-request usage is allowed when quota is exhausted.</summary>
     [JsonPropertyName("overageAllowedWithExhaustedQuota")]
@@ -229,27 +236,28 @@ public class AccountGetQuotaResultQuotaSnapshotsValue
 
     /// <summary>Date when the quota resets (ISO 8601).</summary>
     [JsonPropertyName("resetDate")]
-    public string? ResetDate { get; set; }
+    public DateTimeOffset? ResetDate { get; set; }
 }
 
-/// <summary>RPC data type for AccountGetQuota operations.</summary>
-public class AccountGetQuotaResult
+/// <summary>RPC data type for AccountQuota operations.</summary>
+public class AccountQuota
 {
     /// <summary>Quota snapshots keyed by type (e.g., chat, completions, premium_interactions).</summary>
     [JsonPropertyName("quotaSnapshots")]
-    public IDictionary<string, AccountGetQuotaResultQuotaSnapshotsValue> QuotaSnapshots { get => field ??= new Dictionary<string, AccountGetQuotaResultQuotaSnapshotsValue>(); set; }
+    public IDictionary<string, AccountQuotaSnapshot> QuotaSnapshots { get => field ??= new Dictionary<string, AccountQuotaSnapshot>(); set; }
 }
 
 /// <summary>RPC data type for DiscoveredMcpServer operations.</summary>
 public class DiscoveredMcpServer
 {
     /// <summary>Server name (config key).</summary>
+    [RegularExpression("^[0-9a-zA-Z_.@-]+(\\/[0-9a-zA-Z_.@-]+)*$")]
     [JsonPropertyName("name")]
     public string Name { get; set; } = string.Empty;
 
-    /// <summary>Server type: local, stdio, http, or sse.</summary>
+    /// <summary>Server transport type: stdio, http, sse, or memory (local configs are normalized to stdio).</summary>
     [JsonPropertyName("type")]
-    public string? Type { get; set; }
+    public DiscoveredMcpServerType? Type { get; set; }
 
     /// <summary>Configuration source.</summary>
     [JsonPropertyName("source")]
@@ -265,7 +273,7 @@ public class McpDiscoverResult
 {
     /// <summary>MCP servers discovered from all sources.</summary>
     [JsonPropertyName("servers")]
-    public IList<DiscoveredMcpServer> Servers { get => field ??= []; set; }
+    public List<DiscoveredMcpServer> Servers { get; set; }
 }
 
 /// <summary>RPC data type for McpDiscover operations.</summary>
@@ -297,7 +305,7 @@ internal class SessionFsSetProviderRequest
 
     /// <summary>Path conventions used by this filesystem.</summary>
     [JsonPropertyName("conventions")]
-    public SessionFsSetProviderRequestConventions Conventions { get; set; }
+    public SessionFsSetProviderConventions Conventions { get; set; }
 }
 
 /// <summary>RPC data type for SessionsFork operations.</summary>
@@ -322,16 +330,16 @@ internal class SessionsForkRequest
     public string? ToEventId { get; set; }
 }
 
-/// <summary>RPC data type for SessionLog operations.</summary>
-public class SessionLogResult
+/// <summary>RPC data type for Log operations.</summary>
+public class LogResult
 {
     /// <summary>The unique identifier of the emitted session event.</summary>
     [JsonPropertyName("eventId")]
     public Guid EventId { get; set; }
 }
 
-/// <summary>RPC data type for SessionLog operations.</summary>
-internal class SessionLogRequest
+/// <summary>RPC data type for Log operations.</summary>
+internal class LogRequest
 {
     /// <summary>Target session identifier.</summary>
     [JsonPropertyName("sessionId")]
@@ -343,7 +351,7 @@ internal class SessionLogRequest
 
     /// <summary>Log severity level. Determines how the message is displayed in the timeline. Defaults to "info".</summary>
     [JsonPropertyName("level")]
-    public SessionLogRequestLevel? Level { get; set; }
+    public SessionLogLevel? Level { get; set; }
 
     /// <summary>When true, the message is transient and not persisted to the session event log on disk.</summary>
     [JsonPropertyName("ephemeral")]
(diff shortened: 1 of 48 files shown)


go/rpc/generated_rpc.go
```diff
@@ -556,154 +556,155 @@ type Extension struct {
...
 	// The unique request ID from the elicitation.requested event
 	RequestID string `json:"requestId"`
 	// The elicitation response (accept with form values, decline, or cancel)
-	Result SessionUIHandlePendingElicitationParamsResult `json:"result"`
+	Result ResultClass `json:"result"`
 }
 
 // The elicitation response (accept with form values, decline, or cancel)
-type SessionUIHandlePendingElicitationParamsResult struct {
+type ResultClass struct {
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in add04dfe. The runtime schema already had `title: "UiElicitationResponse"` at both occurrences; the remaining bug was in the quicktype-based SDK codegen. I updated the Go/Python codegen to collapse placeholder `*Class` duplicates onto the explicit titled type when they are structurally identical, so `HandlePendingElicitationRequest.result` now uses `UIElicitationResponse` in Go.

python/copilot/generated/rpc.py
```diff
@@ -2363,62 +2361,62 @@ def to_dict(self) -> dict:
 
 
 @dataclass
-class SessionUIHandlePendingElicitationParamsResult:
+class ResultClass:
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in add04dfe. Same root cause as the Go case: the runtime schema already carried the explicit `UiElicitationResponse` title at both locations, but quicktype was still inventing a placeholder duplicate. The Python codegen now collapses that placeholder duplicate onto `UIElicitationResponse`, so `HandlePendingElicitationRequest.result` uses the same type.

nodejs/src/generated/rpc.ts
```diff
@@ -1194,71 +1196,74 @@ export interface SessionUiHandlePendingElicitationParams {
    * The unique request ID from the elicitation.requested event
    */
   requestId: string;
+  result: UiElicitationResponse;
+}
+/**
+ * The elicitation response (accept with form values, decline, or cancel)
+ */
+export interface UiElicitationResponse {
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in add04dfe. TypeScript now deduplicates repeated exported declarations emitted by per-method `json-schema-to-typescript` compiles, so this no longer produces a second `UiElicitationResponse` declaration.

nodejs/src/generated/rpc.ts
```diff
@@ -443,76 +440,79 @@ export interface SessionModelSwitchToParams {
...
+
+export interface ModeSetResult {
+  mode: SessionMode;
 }
 
-export interface SessionModeSetParams {
+/**
+ * The agent mode. Valid values: "interactive", "plan", "autopilot".
+ */
+export type SessionMode = "interactive" | "plan" | "autopilot";
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in add04dfe. The TS generator now deduplicates repeated named export blocks emitted by separate `json-schema-to-typescript` compiles, so `SessionMode` is only declared once in the generated file.

SteveSandersonMS:
Addressed the review findings in add04dfe. Specifically: (1) TS now deduplicates repeated named export blocks from per-method `json-schema-to-typescript` compiles, which removes the duplicate `SessionMode` and duplicate `UiElicitationResponse` declarations; and (2) the Go/Python quicktype-based generators now collapse placeholder `*Class` duplicates onto the explicit titled type when the structures are identical, so `HandlePendingElicitationRequest.result` now uses `UIElicitationResponse` instead of `ResultClass`.

dotnet/src/Generated/Rpc.cs
```diff
@@ -421,12 +457,12 @@ internal class SessionModelSwitchToRequest
     public ModelCapabilitiesOverride? ModelCapabilities { get; set; }
 }
 
-/// <summary>RPC data type for SessionModeGet operations.</summary>
-public class SessionModeGetResult
+/// <summary>RPC data type for ModeGet operations.</summary>
+public class ModeGetResult
```
stephentoub: Do we expect there to be additional data that's returned from this operation in the future? If not, could it just return a SessionMode instead of a wrapper around one?

dotnet/src/Generated/Rpc.cs
```diff
@@ -437,28 +473,28 @@ internal class SessionModeGetRequest
     public string SessionId { get; set; } = string.Empty;
 }
 
-/// <summary>RPC data type for SessionModeSet operations.</summary>
-public class SessionModeSetResult
+/// <summary>RPC data type for ModeSet operations.</summary>
+public class ModeSetResult
```
stephentoub: Is this needed (rather than void) because the set operation might fail?

dotnet/src/Generated/Rpc.cs
```diff
@@ -481,13 +517,13 @@ internal class SessionPlanReadRequest
     public string SessionId { get; set; } = string.Empty;
 }
 
-/// <summary>RPC data type for SessionPlanUpdate operations.</summary>
-public class SessionPlanUpdateResult
+/// <summary>RPC data type for PlanUpdate operations.</summary>
+public class PlanUpdateResult
```
stephentoub: Some of these namings sound a little odd to my ears, e.g. PlanUpdateResult vs UpdatePlanResult. But not a big deal.

dotnet/src/Generated/Rpc.cs
```diff
@@ -606,9 +642,9 @@ public class Agent
     public string Description { get; set; } = string.Empty;
 }
 
-/// <summary>RPC data type for SessionAgentList operations.</summary>
+/// <summary>RPC data type for AgentList operations.</summary>
 [Experimental(Diagnostics.Experimental)]
-public class SessionAgentListResult
+public class AgentList
```
stephentoub: For C#, we could consider nesting types, having the request/response types for an RPC API be nested within the type that API is defined on, e.g. have AgentList be nested inside of AgentApi. That way, we minimize chances of conflict by giving them a namespace, e.g. when we add an API for getting a list of all subagents, _if_ we were to just call them "agent", we'd have a conflict, but not if that listing method and the one for defined custom agents were on different types.

dotnet/src/Generated/Rpc.cs
```diff
@@ -640,13 +676,13 @@ public class SessionAgentGetCurrentResultAgent
     public string Description { get; set; } = string.Empty;
 }
 
-/// <summary>RPC data type for SessionAgentGetCurrent operations.</summary>
+/// <summary>RPC data type for AgentCurrent operations.</summary>
 [Experimental(Diagnostics.Experimental)]
-public class SessionAgentGetCurrentResult
+public class AgentCurrent
```
stephentoub: Another case where I'm wondering if the wrapper is adding any value. For cases where we already need to return multiple pieces of data, of course, and for cases where we think we will, makes sense... but in a case like this, I struggle to think what else we might need to return, in which case this needn't be in the public API surface area and the relevant method can just return an AgentCurrentAgent (which itself is a questionable name)

dotnet/src/Generated/Rpc.cs
```diff
@@ -711,13 +747,29 @@ internal class SessionAgentDeselectRequest
     public string SessionId { get; set; } = string.Empty;
 }
 
-/// <summary>RPC data type for SessionAgentReload operations.</summary>
+/// <summary>RPC data type for AgentReloadAgent operations.</summary>
+public class AgentReloadAgent
```
stephentoub: We seem to have multiple types that are identical other than in name, e.g. select agent, current agent, reload agent... can we do something to dedup those in the SDK? That would then also help with naming, as we wouldn't need to have multiple names for the exact same thing.

dotnet/src/Generated/Rpc.cs
```diff
@@ -885,15 +937,15 @@ internal class SessionMcpEnableRequest
     public string ServerName { get; set; } = string.Empty;
 }
 
-/// <summary>RPC data type for SessionMcpDisable operations.</summary>
+/// <summary>RPC data type for McpDisable operations.</summary>
 [Experimental(Diagnostics.Experimental)]
-public class SessionMcpDisableResult
+public class McpDisableResult
```
stephentoub: Unless we plan to add more to a type like this, can we model that in the SDK by just having the method return `void` / `Task` instead of an empty dummy object?

dotnet/src/Generated/Rpc.cs
```diff
@@ -1308,6 +1380,120 @@ internal class SessionHistoryTruncateRequest
...
+    [JsonPropertyName("sessionStartTime")]
+    public double SessionStartTime { get; set; }
+
+    /// <summary>Aggregated code change metrics.</summary>
+    [JsonPropertyName("codeChanges")]
+    public UsageMetricsCodeChanges CodeChanges { get => field ??= new(); set; }
+
+    /// <summary>Per-model token and request metrics, keyed by model identifier.</summary>
+    [JsonPropertyName("modelMetrics")]
+    public Dictionary<string, UsageMetricsModelMetric> ModelMetrics { get => field ??= []; set; }
```
stephentoub: Separate from the naming, but seeing here makes me think of it, I think we'll want to update the generator and the manually-written models to use interfaces rather than these concrete types. It gives us more flexibility for what we choose to implement in the future (neither Dictionary or List allows for derivation, overriding, etc., which means we can't do things like modification tracking, throwing on things we consider misuse, etc.) and it allows consumers when setting these to use other types without needing to do a translation.

dotnet/src/Generated/Rpc.cs
```diff
@@ -1464,28 +1650,28 @@ public class SessionFsReaddirParams
...
+public class SessionFsReaddirWithTypesEntry
 {
     /// <summary>Entry name.</summary>
     [JsonPropertyName("name")]
     public string Name { get; set; } = string.Empty;
 
     /// <summary>Entry type.</summary>
     [JsonPropertyName("type")]
-    public EntryType Type { get; set; }
+    public SessionFsReaddirWithTypesEntryType Type { get; set; }
```
stephentoub, line 1615: For .NET, I expect we'd want "Readdir" to be "ReadDir" or "ReadDirectory".

dotnet/src/Generated/Rpc.cs
```diff
@@ -1532,9 +1718,28 @@ public class SessionFsRenameParams
...
+    User,
+    /// <summary>The <c>workspace</c> variant.</summary>
+    [JsonStringEnumMemberName("workspace")]
+    Workspace,
+    /// <summary>The <c>plugin</c> variant.</summary>
+    [JsonStringEnumMemberName("plugin")]
+    Plugin,
+    /// <summary>The <c>builtin</c> variant.</summary>
+    [JsonStringEnumMemberName("builtin")]
+    Builtin,
```
stephentoub, line 1708: Builtin or BuiltIn ?

edburns:
Full disclosure 01: using my human brain Java subject matter expertise, Opus 4.6, and existing context, I arrived at these comments, in light of the comments Stephen [posted to Slack](https://github.slack.com/archives/C09TX0PUJ3Z/p1775847367181649?thread_ts=1775832592.979839&cid=C09TX0PUJ3Z).

Full disclosure 02: The current agentic workflow derives the Java SDK directly from the post-code-generated parts of the reference implementation. I have an outstanding work item ( https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2751554 ) to make it so the Java SDK also uses the Zod schema, in addition to the agentic workflow to reduce toil in keeping the Java SDK in synch with the .NET and TypeScript reference implementations (See [agentic workflow](https://github.com/github/copilot-sdk-java/blob/c05b0a2ab757f1c0b232b441274768d15a1061b7/.github/prompts/agentic-merge-upstream.prompt.md#L4)). Therefore, my comments are to be taken in light of what I would like to see before I undertake the work to implement [dd-2751554-do-code-gen-for-java](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2751554).

Disclosures aside, I stand by the following comments.

### 1. Reinforce @stephentoub 's point on empty types — with a Java-specific angle

> **+1 on eliminating empty result types.** From the Java SDK perspective, the 13 empty types (`PlanUpdateResult`, `AgentDeselect`, `SkillsEnableResult`, `McpReload`, etc.) present an awkward ergonomics issue. Java doesn't have `Task` vs `Task<T>` — we use `CompletableFuture<Void>` for void results. Generating an empty record like `public record McpReload() {}` would feel unnatural to Java developers and force consumers to handle a meaningless return type. If the schema expressed these as null/void results, each language generator could map to its natural void representation (`Task` in C#, `CompletableFuture<Void>` in Java, `None` in Python, empty struct or no return in Go).
>
> **Question:** Would it make sense for empty result schemas to be represented as `null` in the API schema rather than `{ "type": "object", "properties": {} }`? That way generators could choose `void`/`Void` without needing per-language special-casing. Happy to discuss if there's a reason the current shape is preferred.

### 2. Reinforce the duplicate types point — with specific examples

> **+1 on consolidating identical structures.** The Agent group stood out to me: `Agent`, `AgentCurrentAgent`, `AgentSelectAgent`, and `AgentReloadAgent` are structurally identical (all have `name`, `displayName`, `description`). Similarly, `ModelCurrent` and `ModelSwitchToResult` both wrap a single `modelId: string?`. And four types (`UiElicitationResult`, `PermissionRequestResult`, `HandleToolCallResult`, `CommandsHandlePendingCommandResult`) are all `{ success: boolean }`.
>
> In Java, each of these becomes a separate `record` class. The naming machinery in PR 5908 does a nice job producing clean names for them, but I wonder if we could go a step further: would it be feasible to define a shared `Agent` schema (with `$ref`) and have `agent.select`, `agent.current`, and `agent.reload` all reference it? I noticed that `hoistTitledSchemas()` already deduplicates schemas that share the same title — so maybe it's just a matter of giving these inline definitions a shared title? I might be missing a reason they need to be separate.

### 3. Namespacing: Java has a different approach than C# nested types

> **On the namespacing suggestion:** For context on how this lands in Java — our natural namespacing mechanism is **packages**, not nested classes. Where C# would nest `Agent.SelectResult` inside a `class Agent`, Java would typically use `com.github.copilot.sdk.rpc.agents.SelectResult`. This tends to work well in practice because:
> - It avoids the `Agent.Agent` pattern (outer class and inner class sharing a name, which is a Java code smell).
> - It aligns with how Jackson deserialization resolves types.
> - It enables per-package `package-info.java` for group documentation.
>
> **One thing I'd love to explore:** package-based namespacing would benefit from the schema carrying some grouping information so generators know which package/namespace a type belongs to. Today, the schema titles are flat strings (`AgentSelectResult`, `McpConfigServerLocal`). I'm wondering whether it might be worth considering an optional `group` or `namespace` annotation alongside `title` — that way each language generator could map to its natural grouping mechanism (packages in Java, nested types in C#, module prefixes in Python) without parsing the group from the type name string. This may be out of scope for this PR, but wanted to plant the seed.

### 4. Java-specific observation: the MCP filter mapping type explosion

> **From the Java SDK perspective, the MCP config types are a notable hotspot.** There are 18 `FilterMapping` variant types across 6 contexts (server local, server http, add local, add http, update local, update http) — structurally identical in groups of 3. In Java, each would become a separate sealed interface + record hierarchy, with identical Jackson `@JsonTypeInfo`/`@JsonSubTypes` annotations multiplied 6×.
>
> This hasn't been an issue in practice because the Java SDK hasn't implemented MCP config types yet (it was easy to skip when hand-writing). With codegen, they'd all materialize automatically. I'm curious — is there a reason these can't share a single `McpConfigFilterMapping` definition with `$ref`? There may be a forward-compatibility reason I'm not seeing for keeping them separate.

### 5. Stability guarantee + Java sealed types interaction

> **The stability guarantee from PR 5908 (new schemas never rename existing types) is very welcome.** One subtlety for Java: with sealed interfaces/classes, adding a new event type requires updating the `permits` clause on the base type. This means the generated base type changes on every new event — which is fine since it's generated. But it also means **consumers who do exhaustive `switch` on event types will get compile warnings when new events are added.** This is actually desirable behavior (it's the point of sealed types), but it's worth documenting as an expected contract: adding a new event type in the runtime is a source-compatible but not pattern-exhaustiveness-compatible change.

### 6. Single-property wrappers: a different perspective from the Java side

> **On single-property wrapper types:** I wanted to offer a slightly different perspective from the Java side. Stephen's suggestion to return the raw value directly (e.g., `string` instead of `ShellExecResult { processId: string }`) makes a lot of sense for .NET ergonomics. In Java, there's a tension: `CompletableFuture<String>` is less self-documenting than `CompletableFuture<ShellExecResult>`, and unwrapping removes the ability to add fields later without a breaking change.
>
> **One possible approach:** Keep the wrapper types in the schema (for extensibility), but mark single-property types with a schema annotation (e.g., `"unwrappable": true`). Generators that prefer unwrapping (C#) could do so; generators that prefer wrapper types (Java, for future-proofing) could ignore it. This would avoid baking an unwrapping decision into the shared schema that all languages must follow. That said, I recognize this adds complexity to the schema — I'm open to simpler alternatives if others have ideas.

dotnet/src/Generated/Rpc.cs
MackinnonBuck: I know this isn't naming-related, but lots of APIs use `double` when they would be better represented as `int`, `TimeSpan`, etc. It would be nice if we could support a richer set of numeric types in generated code.

dotnet/src/Generated/Rpc.cs
```diff
@@ -155,8 +155,8 @@ public class Model
     public string? DefaultReasoningEffort { get; set; }
 }
 
-/// <summary>RPC data type for ModelsList operations.</summary>
-public class ModelsListResult
+/// <summary>RPC data type for ModelList operations.</summary>
+public class ModelList
```
MackinnonBuck: Would this be a good time to make these classes `sealed`?

dotnet/src/Generated/Rpc.cs
```diff
@@ -2297,17 +2513,17 @@ internal UiApi(JsonRpc rpc, string sessionId)
     }
 
     /// <summary>Calls "session.ui.elicitation".</summary>
-    public async Task<SessionUiElicitationResult> ElicitationAsync(string message, SessionUiElicitationRequestRequestedSchema requestedSchema, CancellationToken cancellationToken = default)
+    public async Task<UiElicitationResponse> ElicitationAsync(string message, UiElicitationSchema requestedSchema, CancellationToken cancellationToken = default)
```
MackinnonBuck: I think `ElicitAsync` might be a better name for this.

dotnet/src/Generated/Rpc.cs
MackinnonBuck: Seems like this is a wrapper type that probably doesn't have to exist (could we just use `List<Skill>` directly)?

MackinnonBuck: I see similar patterns elsewhere (like `McpList`)

dotnet/src/Generated/Rpc.cs
```diff
@@ -848,13 +900,13 @@ public class Server
...
+/// <summary>RPC data type for McpList operations.</summary>
 [Experimental(Diagnostics.Experimental)]
-public class SessionMcpListResult
+public class McpList
 {
     /// <summary>Configured MCP servers.</summary>
     [JsonPropertyName("servers")]
-    public List<Server> Servers { get => field ??= []; set; }
+    public List<McpServer> Servers { get => field ??= []; set; }
 }
```
MackinnonBuck: Even if we decide to keep these wrapper types, it seems like this one should be `McpServerList`.

dotnet/src/Generated/Rpc.cs
```diff
@@ -1104,20 +1156,20 @@ internal class SessionCommandsHandlePendingCommandRequest
     public string? Error { get; set; }
 }
 
-/// <summary>RPC data type for SessionUiElicitation operations.</summary>
-public class SessionUiElicitationResult
+/// <summary>The elicitation response (accept with form values, decline, or cancel).</summary>
+public class UiElicitationResponse
```
MackinnonBuck: Maybe a nit, but I think `UI` feels more natural than `Ui`.

dotnet/src/Generated/Rpc.cs
```diff
@@ -1308,6 +1380,120 @@ internal class SessionHistoryTruncateRequest
...
+    [JsonPropertyName("count")]
+    public double Count { get; set; }
+
+    /// <summary>User-initiated premium request cost (with multiplier applied).</summary>
+    [JsonPropertyName("cost")]
+    public double Cost { get; set; }
+}
+
+/// <summary>Token usage metrics for this model.</summary>
+public class UsageMetricsModelMetricUsage
```
MackinnonBuck: Some of these names still seem a bit unwieldy (though for cases like this I anticipate we could just use `.withTypeName` and override it).

dotnet/src/Generated/Rpc.cs
```diff
@@ -1308,6 +1380,120 @@ internal class SessionHistoryTruncateRequest
...
+public class UsageMetricsModelMetric
+{
+    /// <summary>Request count and cost metrics for this model.</summary>
+    [JsonPropertyName("requests")]
+    public UsageMetricsModelMetricRequests Requests { get => field ??= new(); set; }
+
+    /// <summary>Token usage metrics for this model.</summary>
+    [JsonPropertyName("usage")]
+    public UsageMetricsModelMetricUsage Usage { get => field ??= new(); set; }
+}
```
MackinnonBuck: I wonder if some of these would make sense as nested types.

dotnet/src/Generated/Rpc.cs
stephentoub: https://github.com/github/copilot-sdk/pull/1062 is adding $ref support. And https://github.com/github/copilot-agent-runtime/pull/6059 takes advantage of that to avoid a bazillion copies of the same data type. If we like this, we can use it to dedup in the existing schema.

dotnet/src/Generated/Rpc.cs
stephentoub: This is now done.

dotnet/src/Generated/Rpc.cs
stephentoub: I've updated the C# generator to use long for integer, and I've merged an update into copilot-agent-runtime that fixes a bunch of properties typed as number to instead be typed as integer. If we want to take this a step further, we could update the schema in runtime to use min/max in the schema, and then in the generators use those ranges to further constrain the type, e.g. if max is <= int.MaxValue, use int instead of long.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Good point. Previously we didn't support returning non-object results from RPC methods, but I've updated the codegenerators to support that now (e.g., for enum, string, or other primitive results values) and on the runtime side have unwrapped `SessionMode` in the get/set methods. This shows we can now avoid the unwanted wrapper classes.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: I don't know if it's strictly needed but I can certainly imagine some future scenario where you ask for mode X but we decide that's not valid right now and return some other mode that we fell back on.

In any case we should make sure we can have `void`-returning RPC methods in case we want that. I've updated the codegenerator so that it does support that for all languages.

Note that the Go variant of this is weird. Unlike the other languages, changing from void to non-void would be a breaking change because we'd go from:

```go
func (a *ModeApi) Set(ctx context.Context, params *ModeSetRequest) error { ... }
```

to:

```go
func (a *ModeApi) Set(ctx context.Context, params *ModeSetRequest) (*SessionMode, error) { ... }
```

... and so caller code that operates on `error` (i.e., all well-written code) would have to be updated.

To handle this I'm updating the Go codegen specifically so that it doesn't use `error` on its own as a result type and always produces synthetic/empty structs so in this case it returns `(*SessionModeSetResult, error)` even though `SessionModeSetResult` is an empty struct.

According to Copilot this is not completely weird in Go - gRPC/protobuf does the same thing in its own codegenerator.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Reversing the noun-verb order probably requires us to set names at each point in the schema on the runtime side. We can consider doing that when we do a full API review but would be most practical to keep that separate from the current goal to improve mechanical naming.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Agreed that might be better for C#, but there'll be limited value in avoiding conflicts for C# if the same types still conflict in other languages. Definitely open to doing something like this when we do a full API review if we conclude there's a pattern that's going to work across languages.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Note that this still constrains us in the future. We **cannot** later change mode.set to return `SessionMode`. We would have to change it to return a wrapper object, `SessionModeSetResult`, otherwise it would be a breaking change for Go.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Agreed, but would it be OK to limit this PR to changes that affect the codegen machinery rather than being an API review in general? For any changes that force us to update the codegenerator I'll definitely try to handle them here, but for ones that are purely schema updates on the runtime side it would be good to handle them separately as there may be a pretty long tail.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: That's great, thanks!

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: I know that would be more conventional for .NET but you could equally say that for Node it should be `readdir` to match the `fs` APIs. Let's later consider whether we want to add annotations for per-language names. Trying to make truly optimal APIs across all languages might be counterproductive.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Let's come back to that when we do an API review

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: We could make that change but it doesn't give us total freedom in the future. If we later want to make it non-void, we will be restricted to making it return `McpDisableResult` (and not a primitive or any other type name) otherwise it will be a breaking change for Go.

SteveSandersonMS:
About empty result types: there were about 15 of these, and I've removed them all.

Well, the reality is a bit more complex: for Go we need to retain them for forward compat, which in turn means we also need to retain the name generator output from the runtime so we wouldn't break them name in the future if we changed from void to nonvoid.

I've updated things on both sides so that Go is able to produce empty result structs with the desired names, and the other languages use their own versions of `void`.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Potentially, but let's separate that out into future runtime-side schema changes.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Agreed - seems like one for the API review.

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Agreed - seems like one for the API review.

SteveSandersonMS:
> 1. Reinforce @stephentoub 's point on empty types — with a Java-specific angle
> +1 on eliminating empty result types. 

This is done, though it's not a perfect solution given that in Go it creates an extra forward-compat concern. The middle ground we're in now is:

 * For Go, it continues to produce empty result types so the return value can be `(*SomeType, error)`
 * For other languages it now uses their version of void/async-void

In the future, if we want to change these methods to be nonvoid, we will be constrained to give them the type `SomeType` and not anything else (not a primitive, not some other name) for compatibility with existing Go code.

> 2. Reinforce the duplicate types point — with specific examples

Valid but we must do this through runtime schema changes (e.g., using `$ref`). We can't collapse types just because they are structurally equivalent as that would be breaking if they stop being equivalent later. So likely we should keep that out of scope until we add `$ref` in @stephentoub's PR.

> 3. Namespacing: Java has a different approach than C# nested types

This PR is already quite hard to land so would be good to get in what we have before we think about systems for nesting types. Given the large differences between languages (e.g., how TypeScript doesn't have any global namespace) it might or might not be productive to try to have a unified model here.

> 4. Java-specific observation: the MCP filter mapping type explosion

We can change whatever we think is best on the runtime schema side. For this PR I'm trying to focus on the machinery that generates the JSON schema and then renders that out into per-language code. 

> 5. Stability guarantee + Java sealed types interaction
> adding a new event type in the runtime is a source-compatible but not pattern-exhaustiveness-compatible change.

Understood, thanks.

> 7. Single-property wrappers: a different perspective from the Java side

With the example given (`CompletableFuture<String>` vs `CompletableFuture<ShellExecResult>`), it's no different from in .NET (`Task<string>` vs `Task<ShellExecResult>`). So I'm not sure there's any Java-vs-.NET reason to vary the representation. We can't say that the ability to add more properties in the future is a reason to have some but not all languages use wrappers - either they must all do it, or none of them can, since changing a primitive return to an object return would be breaking.

The way the machinery works now is that the runtime schema decides whether a given method's return value is a raw primitive or is wrapped in some object. The person authoring that schema has to decide on the likelihood of needing to add extra properties in the future.

I think our options are:

 * Risk averse - say we always think more might be added, and hence require every RPC method to return a wrapper `*Result` type (even things that would otherwise be void). That's what we had before all this work but @stephentoub was recommending not to do that.
 * Let the runtime schema decide - we have a case-by-case decision about how we want to represent the RPC result and whether it might be expandable with other properties in the future. The ergonomics are better, at the cost that it's more painful if our guess turns out to be wrong.

Can we decide which of these two we're going for? I've implemented the second option and am happy to stick with that unless there are big concerns.

stephentoub:
> Let the runtime schema decide - we have a case-by-case decision about how we want to represent the RPC result and whether it might be expandable with other properties in the future.

I think this is the right answer. The runtime here is the source-of-truth for the shape of the API, and the developer of that API needs to say whether there's any possibility it could be expanded in the future.

dotnet/src/Generated/Rpc.cs
```diff
@@ -156,12 +161,12 @@ public class Model
...
 
-/// <summary>RPC data type for ModelsList operations.</summary>
-public class ModelsListResult
+/// <summary>RPC data type for ModelList operations.</summary>
+public class ModelList
 {
     /// <summary>List of available models with full metadata.</summary>
     [JsonPropertyName("models")]
-    public IList<Model> Models { get => field ??= []; set; }
+    public List<Model> Models { get; set; }
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed to IList

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Good, fixed

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Trying to keep the UI prefix for disambiguation with other kinds of elicitation

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Good idea - done

scripts/codegen/python.ts
```diff
@@ -332,7 +434,19 @@ def _timeout_kwargs(timeout: float | None) -> dict:
...
+    // Patch models.list to normalize capabilities before deserialization
+    let finalCode = lines.join("\n");
+    finalCode = finalCode.replace(
+        `ModelList.from_dict(await self._client.request("models.list"`,
+        `ModelList.from_dict(_patch_model_capabilities(await self._client.request("models.list"`,
+    );
+    // Close the extra paren opened by _patch_model_capabilities(
+    finalCode = finalCode.replace(
+        /(_patch_model_capabilities\(await self\._client\.request\("models\.list",\s*\{[^)]*\)[^)]*\))/,
+        "$1)",
```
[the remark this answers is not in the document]
SteveSandersonMS: Yes, we'll remove this once we update the schema in runtime (PR is already out there)

PR #1048, opened 2026-04-08
Export ProviderConfig from Node.js and Python SDKs

### LINKED ISSUE #968: Export ProviderConfig to better support Responses API
Looking at exported types, ProviderConfig defined in [nodejs/src/types.ts#L1226](https://github.com/github/copilot-sdk/blob/346a38ecad61d38ca3a0f72c93cae755dbb0f706/nodejs/src/types.ts#L1226) is not one of them.

This appears to be the only way to access 'responses' API - by setting your `ProviderConfig.wireApi := 'responses'`. Currently, I have to duplicate the type locally. I suspect it's an oversight that it isn't exported.

stephentoub:
`ProviderConfig` is the only way to configure the Responses API (`wireApi: 'responses'`) and BYOK providers, but it wasn't re-exported from the public API of the Node.js or Python SDKs. Consumers had to duplicate the type locally to reference it in their own code.

This adds `ProviderConfig` to the public exports of both SDKs:

- **Node.js** — added to the `export type { ... }` block in `src/index.ts`
- **Python** — imported from `session` and added to `__all__` in `copilot/__init__.py`

Go and .NET were already fine — Go exports via capitalized naming and .NET via `public class`.

Fixes https://github.com/github/copilot-sdk/issues/968


SteveSandersonMS:
Seems worth doing to simplify things for consumers.

I mentioned to Jakub earlier this week that it's also possible to reference the type without it being exported (it's  `SessionConfig["provider"]`), so it's not really necessary to duplicate it. Jakub thought that was sufficient, but it's certainly not as obvious, so exporting it with a name makes sense.

SteveSandersonMS (approved):
Thanks!

PR #1051, opened 2026-04-09
Refactor MCP server config types across all SDK languages

### LINKED ISSUE #245: improve typing of `McpServers` in C# sdk.
it's possible to set `McpServers` on the following types

SessionConfig - https://github.com/github/copilot-sdk/blob/c34e9e714cbd0d5d6a56f66f738323202f75643d/dotnet/src/Types.cs#L365
ResumeSessionConfig - https://github.com/github/copilot-sdk/blob/c34e9e714cbd0d5d6a56f66f738323202f75643d/dotnet/src/Types.cs#L411

the docs contain the following:

    Keys are server names, values are server configurations (McpLocalServerConfig or McpRemoteServerConfig).

I think it can be better by introducing a base-class/interface and having `McpLocalServerConfig` & `McpRemoteServerConfig` extend/implement that base type.

stephentoub:
## Motivation

The C# SDK typed `McpServers` dictionaries as `Dictionary<string, object>`, losing all compile-time safety and forcing consumers to know which concrete config type to use without IDE guidance. The type names (`McpLocalServerConfig` / `McpRemoteServerConfig`) also didn't align with MCP protocol terminology. Fixes #245.

## Approach

**C# type hierarchy** — Introduces an abstract `McpServerConfig` base class with:
- A `private protected` constructor (no external subclassing)
- `sealed` derived types `McpStdioServerConfig` and `McpHttpServerConfig`
- Shared properties (`Tools`, `Type`, `Timeout`) deduplicated into the base
- `[JsonPolymorphic]` discriminator on `"type"` following the same pattern as `SessionEvent`
- `[JsonIgnore]` virtual `Type` property overridden in each derived type

**Strongly-typed dictionaries** — All `Dictionary<string, object>` McpServers properties (`SessionConfig`, `ResumeSessionConfig`, `CustomAgentConfig`, internal RPC records) are now `Dictionary<string, McpServerConfig>`.

**Cross-language rename** — Renamed Local/Remote → Stdio/Http across all four SDKs to match MCP protocol terminology:

| Language | Stdio | HTTP |
|----------|-------|------|
| C# | `McpStdioServerConfig` | `McpHttpServerConfig` |
| TypeScript | `MCPStdioServerConfig` | `MCPHttpServerConfig` |
| Go | `MCPStdioServerConfig` | `MCPHTTPServerConfig` |
| Python | `MCPStdioServerConfig` | `MCPHttpServerConfig` |

**Docs** — Updated MCP feature guide and troubleshooting guide with new type names and removed `Type = "local"` from C# examples (now handled by the type discriminator).

## Validation

- .NET: builds clean, all 13 clone/unit tests pass
- Go: builds clean
- TypeScript: no rename-related type errors (pre-existing unrelated errors from missing dev deps in this environment)
- Grep sweep: zero remaining references to old type names across all files


stephentoub:
Replying to [review comment on Types.cs line 1422](https://github.com/github/copilot-sdk/pull/1051/files#r213456):

The suggestion won't compile — `JsonDerivedType` doesn't allow registering the same type with multiple discriminator values. Attempting it would throw `InvalidOperationException` at runtime.

`IgnoreUnrecognizedTypeDiscriminators = true` is kept for forward-compatibility (matching the existing pattern on `SessionEvent`). The primary use case here is serialization (SDK to CLI), where the types always emit the canonical `stdio` and `http` discriminators. Legacy values like `local` and `sse` are runtime-only synonyms that don't need SDK-side deserialization support.

stephentoub:
@SteveSandersonMS, mind taking a peek at this one, just to make sure you agree with the API changes?

SteveSandersonMS (approved):
Great - the stdio/http terminology is way clearer than local/remote.

PR #1063, opened 2026-04-11
Generate dedicated Python session event types

stephentoub:
## Summary
Python session events still used quicktype and collapsed payload members into one merged `Data` model, unlike the newer Go implementation. That made the language bindings inconsistent and increased the chance of naming conflicts as the runtime schema grows.

This change switches Python session-event generation to dedicated per-event payload dataclasses, matching the newer Go-style approach. `SessionEvent.from_dict()` now dispatches to typed payload classes, unknown events still round-trip through `RawSessionEventData`, and the Python runtime/tests were updated to handle the typed payload union.

The one non-obvious part is compatibility: the generator also restores legacy top-level helper exports and keeps arbitrary nested mappings in the `Data` shim as plain `dict`s so older callers do not break while the generated model becomes more structured.

## Validation
- `python -m pytest test_event_forward_compatibility.py test_commands_and_elicitation.py`
- `python -m pytest --ignore=e2e`


scripts/codegen/python.ts
```diff
@@ -144,59 +144,1145 @@
...
+    const valuesKey = [...values].sort().join("|");
+    const existing = ctx.enumsByValues.get(valuesKey);
+    if (existing) {
+        return existing;
+    }
+
+    const lines: string[] = [];
+    if (description) {
+        lines.push(`class ${enumName}(Enum):`);
+        lines.push(`    """${description.replace(/"/g, '\\"')}"""`);
```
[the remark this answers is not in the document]
stephentoub: Implemented in a5e3ec9. The generator now emits docstrings via a safe Python string literal, so backslashes and quotes are escaped in the generated source instead of being embedded raw.

scripts/codegen/python.ts
```diff
@@ -144,59 +144,1145 @@
...
+            resolved,
+        };
     });
 
-    let code = result.lines.join("\n");
+    const lines: string[] = [];
+    lines.push(`@dataclass`);
+    lines.push(`class ${typeName}:`);
+    if (description || schema.description) {
+        lines.push(`    """${(description || schema.description || "").replace(/"/g, '\\"')}"""`);
```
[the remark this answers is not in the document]
stephentoub: Implemented in a5e3ec9. The generator now emits docstrings via a safe Python string literal, so backslashes and quotes are escaped in the generated source instead of being embedded raw.

scripts/codegen/python.ts
```diff
@@ -144,59 +144,1145 @@
...
+            resolved,
+        };
+    });
 
-`;
+    const lines: string[] = [];
+    lines.push(`@dataclass`);
+    lines.push(`class ${typeName}:`);
+    if (description) {
+        lines.push(`    """${description.replace(/"/g, '\\"')}"""`);
```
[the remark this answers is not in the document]
stephentoub: Implemented in a5e3ec9. The generator now emits docstrings via a safe Python string literal, so backslashes and quotes are escaped in the generated source instead of being embedded raw.

scripts/codegen/python.ts
```diff
@@ -144,59 +144,1145 @@ function collectRpcMethods(node: Record<string, unknown>): RpcMethod[] {
...
+    out.push(``);
+    out.push(`    def to_dict(self) -> Any:`);
+    out.push(`        return self.raw`);
+    out.push(``);
+    out.push(``);
+    out.push(`def _compat_to_python_key(name: str) -> str:`);
+    out.push(`    result: list[str] = []`);
+    out.push(`    for index, char in enumerate(name.replace(".", "_")):`);
+    out.push(
+        `        if char.isupper() and index > 0 and (not name[index - 1].isupper() or (index + 1 < len(name) and name[index + 1].islower())):`
```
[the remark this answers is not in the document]
stephentoub: Good catch. The generator now normalizes 
ame.replace(".", "_") once and uses that normalized string for both the loop and the neighbor checks, so dotted keys like oo.Bar now become oo_bar consistently.

scripts/codegen/python.ts
```diff
@@ -144,59 +144,1145 @@ function collectRpcMethods(node: Record<string, unknown>): RpcMethod[] {
...
+    const lines: string[] = [];
+    if (description) {
+        lines.push(`class ${enumName}(Enum):`);
+        lines.push(`    """${description.replace(/"/g, '\\"')}"""`);
+    } else {
+        lines.push(`class ${enumName}(Enum):`);
+    }
+    for (const value of values) {
+        lines.push(`    ${toEnumMemberName(value)} = ${JSON.stringify(value)}`);
+    }
```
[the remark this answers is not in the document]
stephentoub: I am not applying this in this PR. This refactor intentionally preserves forward compatibility only for top-level event types, where we also preserve the raw event type for round-tripping. Adding generic UNKNOWN fallbacks to every generated enum without preserving each raw enum value would change existing behavior and could break payload round-tripping, so that broader cross-SDK design change is better handled separately.

python/copilot/session.py
```diff
@@ -1171,11 +1178,13 @@ def on(self, handler: Callable[[SessionEvent], None]) -> Callable[[], None]:
...
+            >>> from copilot.generated.session_events import AssistantMessageData, SessionErrorData
             >>> def handle_event(event):
-            ...     if event.type == "assistant.message":
-            ...         print(f"Assistant: {event.data.content}")
-            ...     elif event.type == "session.error":
-            ...         print(f"Error: {event.data.message}")
+            ...     if event.type.value == "assistant.message":
+            ...         print(f"Assistant: {cast(AssistantMessageData, event.data).content}")
+            ...     elif event.type.value == "session.error":
+            ...         print(f"Error: {cast(SessionErrorData, event.data).message}")
```
SteveSandersonMS: Rather than the cast, can we recommand a pattern like:

```py
if isinstance(event.data, AssistantMessageData):
     print(f"Assistant: {event.data.content}")  # type-checker narrows automatically
 elif isinstance(event.data, SessionErrorData):
     print(f"Error: {event.data.message}")
```

SteveSandersonMS (approved):
Generally looks good to me, would be interesting to run it by @brettcannon if you want more authoritative Python feedback.

python/copilot/session.py
```diff
@@ -1123,9 +1128,11 @@ async def send_and_wait(
             Exception: If the session has been disconnected or the connection fails.
 
         Example:
+            >>> from typing import cast
+            >>> from copilot.generated.session_events import AssistantMessageData
             >>> response = await session.send_and_wait("What is 2+2?")
             >>> if response:
-            ...     print(response.data.content)
+            ...     print(cast(AssistantMessageData, response.data).content)
```
brettcannon: Why the cast when you're just printing something to stdout?

python/copilot/session.py
```diff
@@ -1171,11 +1178,13 @@ def on(self, handler: Callable[[SessionEvent], None]) -> Callable[[], None]:
...
 
         Example:
+            >>> from typing import cast
+            >>> from copilot.generated.session_events import AssistantMessageData, SessionErrorData
             >>> def handle_event(event):
-            ...     if event.type == "assistant.message":
-            ...         print(f"Assistant: {event.data.content}")
-            ...     elif event.type == "session.error":
-            ...         print(f"Error: {event.data.message}")
+            ...     if event.type.value == "assistant.message":
```
brettcannon: This can use `match` to be more readable.

python/copilot/session.py
```diff
@@ -1222,32 +1231,34 @@ def _handle_broadcast_event(self, event: SessionEvent) -> None:
         are broadcast as session events to all clients.
         """
         if event.type == SessionEventType.EXTERNAL_TOOL_REQUESTED:
-            request_id = event.data.request_id
-            tool_name = event.data.tool_name
+            data = cast(ExternalToolRequestedData, event.data)
```
brettcannon: The perpetual casting has me concerned. Why is the type wrong based on all the other typing in the code such that you have to override it everywhere? That suggests to me that either something about the type of `event.data` needs to change (e.g. be a protocol), this needs to be a different attribute, or the attribute can be multiple types and people need to test which type it is. But using `cast()` should be a rare thing and not something used at nearly every attribute access.

stephentoub: > or the attribute can be multiple types and people need to test which type it is

To clarify, you'd be fine if these were instead runtime checks with isinstance, it's just the cast you object to?

python/copilot/session.py
stephentoub: Good call. I updated the Python doc examples to use isinstance-based narrowing instead of casts so the examples stay simpler and match what type checkers understand automatically.

PR #1128, opened 2026-04-23
docs: fix Python examples passing dict to session.send() instead of positional str + keyword args

### LINKED ISSUE #1127: [Bug] `session.error` with `t.asString is not a function` when using `session.send()` with BYOK provider
**## Description**

When using the Python SDK with a BYOK provider (tested with Azure OpenAI and Anthropic) and sending a prompt via `session.send({"prompt": "..."})`, every query results in a `session.error` event with the message `t.asString is not a function`. No `assistant.message` event is ever received — the session silently goes idle.

This makes the SDK completely unusable with BYOK when following the documented `session.send()` pattern.

**## Environment**

- ***SDK version:**** `github-copilot-sdk==0.2.2`
- ***CLI binary version:**** `copilot.exe` v1.0.21 (bundled)
- ***Python version:**** 3.13
- ***OS:**** Windows 11
- ***Providers tested:**** Azure OpenAI (`type: "azure"`), Anthropic (`type: "anthropic"`)

**## Steps to Reproduce**

1. Install the SDK:

```bash

pip install github-copilot-sdk==0.2.2

```

2. Create and run the following script with a valid BYOK provider and API key:

```python

import asyncio

import os

from dotenv import load_dotenv

from copilot.session import PermissionHandler

load_dotenv()

FOUNDRY_MODEL_URL = "https://<your-resource>.openai.azure.com/"

async def main():

client = CopilotClient()

await client.start()

session = await client.create_session(

on_permission_request=PermissionHandler.approve_all,

model="<your-deployment-name>",

provider={

"type": "azure",

"base_url": FOUNDRY_MODEL_URL,

"api_key": os.environ["FOUNDRY_API_KEY"],

},

available_tools=[],

infinite_sessions={"enabled": False},

)

done = asyncio.Event()

def on_event(event):

print(f"Received event: {event.type.value}")

if event.type.value == "assistant.message":

print(event.data.content)

elif event.type.value == "session.error":

print(f"error_type={event.data.error_type!r}  message={event.data.message!r}")

done.set()

elif event.type.value == "session.idle":

done.set()

session.on(on_event)

await session.send({"prompt": "What is 2+2?"})

await done.wait()

await session.disconnect()

await client.stop()

asyncio.run(main())

```

**## Actual Behaviour**

The `session.error` event fires with `t.asString is not a function`. No `assistant.message` event is received. The session then goes idle.

```

Received event: pending_messages.modified

Received event: pending_messages.modified

Received event: session.info

Received event: session.tools_updated

Received event: session.error

error_type='query'  message='t.asString is not a function'

Received event: session.idle

```

**## Expected Behaviour**

A `session.error` event should not be fired. A `assistant.message` event should be received containing the model's response content.

**## Additional Notes**

- Verified the BYOK endpoint itself is healthy — a direct HTTP call to the Azure OpenAI API returns a valid `200` response:

```json

{

"choices": [

{

"finish_reason": "stop",

"message": { "content": "2 + 2 equals **4**.", "role": "assistant" }

}

],

"model": "gpt-4o-2024-11-20",

"usage": {

"completion_tokens": 11,

"prompt_tokens": 14,

"total_tokens": 25

}

}

```

- The error occurs regardless of provider type (`"azure"`, `"openai"`, `"anthropic"`).
- The error occurs regardless of whether `available_tools=[]` is set.
- The reproduction code was written by following the example in `docs/auth/byok.md` in this repository.

Akhand-99:
Closes #1127 

## What does this change do?
 
Fixes broken Python code examples in two docs files where `session.send()` is called with a `dict` instead of the correct positional `str` and keyword arguments.
 
**`docs/auth/byok.md`** — 1 line changed:
 
```diff
- await session.send({"prompt": "What is 2+2?"})
+ await session.send("What is 2+2?")
```
 
**`docs/features/steering-and-queueing.md`** — 3 Python blocks, 8 lines changed:
 
```diff
- msg_id = await session.send({
-     "prompt": "Refactor the authentication module to use sessions",
- })
+ msg_id = await session.send("Refactor the authentication module to use sessions")
 
- await session.send({
-     "prompt": "Actually, use JWT tokens instead of sessions",
-     "mode": "immediate",
- })
+ await session.send(
+     "Actually, use JWT tokens instead of sessions",
+     mode="immediate",
+ )
 
# (same pattern repeated for the queueing and combining sections)
```
 
## Why is this needed?
 
`session.send()` has this signature:
 
```python
async def send(self, prompt: str, *, mode: Literal["enqueue", "immediate"] | None = None) -> str:
```
 
Python does not enforce type hints at runtime. When a dict is passed as `prompt`, two things silently break:
 
**1. `prompt` bug (both files):** The dict is forwarded verbatim as the `prompt` value in the JSON-RPC payload to the CLI binary:
 
```json
{ "sessionId": "...", "prompt": { "prompt": "What is 2+2?" } }
```
 
The CLI receives `prompt` as a JSON object instead of a string and calls `.asString()` on it, throwing `TypeError: t.asString is not a function`. This surfaces as a `session.error` event — no Python exception is raised, making it extremely difficult to diagnose:
 
```
Received event: session.info
Received event: session.tools_updated
Received event: session.error
error_type='query'  message='t.asString is not a function'
Received event: session.idle
```
 
**2. `mode` bug (`steering-and-queueing.md` only):** The `*` in the signature makes `mode` a keyword-only argument. When `mode` is buried inside a dict, it is never actually passed to `send()` — the steering and queueing behaviour silently does nothing. The agent receives no delivery mode instruction at all.
 
Both issues are caused by the same issue in the docs: using a dict where the Python SDK expects positional and keyword arguments.
 
## Notes for maintainers
 
- All other language tabs (Node.js, Go, .NET, Java) use the correct calling convention and are unaffected.
- The `byok.md` fix is a single line. In that example I kept the existing `asyncio.Event` / `on_event` handler pattern unchanged to keep the diff minimal. If you prefer to align it with the other language tabs by switching to `send_and_wait("...")`, I am happy to update.
- A related improvement worth considering (out of scope for this PR): adding `if not isinstance(prompt, str): raise TypeError(...)` in `session.py` `send()` would surface this class of mistake immediately at the Python level rather than as a cryptic CLI error. Happy to follow up with a separate issue or PR if that would be useful.


Akhand-99:
I kindly request a review of this PR.

SteveSandersonMS:
Thanks for submitting this, @Akhand-99! As it turns out these docs have already been fixed in the same way in the meantime.

PR #1164, opened 2026-04-29
Add Rust SDK (technical preview)

tclem:
Adds a Rust SDK alongside the existing Node, Python, Go, and .NET SDKs in this repo. Same JSON-RPC client model, same protocol, same session lifecycle — just in Rust.

> [!IMPORTANT]
> **Technical preview.** This is published as `github-copilot-sdk = "0.1"` (pre-1.0) and the public API is **subject to breaking changes** as we iterate. Pin to an exact version, expect churn, and please file issues for friction or missing parity.

See [`rust/README.md`](./rust/README.md) for the full overview, examples, and the build/test commands. Generated types follow the same schema-driven flow used by the other SDKs (`scripts/codegen/rust.ts`).

CI for the new crate runs in [`.github/workflows/rust-sdk-tests.yml`](./.github/workflows/rust-sdk-tests.yml).

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>


.github/skills/rust-coding-skill/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+  (1) `std`/`core`/`alloc`, (2) external crates, (3)
+  `crate::`/`super::`/`self::`. Enforced by nightly `cargo fmt` via
+  `rust/.rustfmt.nightly.toml`.
+- **`pub(crate)` vs `pub`** — most modules in `lib.rs` are private (`mod`), so
+  `pub` items inside them are already crate-private. Use `pub(crate)` only when
+  you want to be explicit that an item must not become part of the public API.
+
+## Testing
+
+- **No mock testing.** Depend on real implementations, spin up lightweight
```
stephentoub, line 173: What's the reasoning for this? For this kind of thing, it seems desirable to have a relatively consistent policy across all the SDKs.

.github/skills/rust-coding-skill/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+## Build speed
+
+Specify Tokio features explicitly — never `features = ["full"]`. Iterate with
+`cargo check`; reach for `cargo build` only when you need the binary. Audit
+new dependency feature flags with `cargo tree` before committing.
+
+## Comments
+
+Explain **why**, never **what**. No comments that restate code. No decorative
+banners (`// ── Section ────────`).
```
stephentoub, line 210: This isn't really Rust specific, is it?

.github/copilot-instructions.md
```diff
@@ -48,7 +48,13 @@
...
-- SDK code: `nodejs/src`, `python/copilot`, `go`, `dotnet/src`
-- Unit tests: `nodejs/test`, `python/*`, `go/*`, `dotnet/test`
+- SDK code: `nodejs/src`, `python/copilot`, `go`, `dotnet/src`, `rust/src`
+- Unit tests: `nodejs/test`, `python/*`, `go/*`, `dotnet/test`, `rust/tests`
 - E2E tests: `*/e2e/` folders that use the shared replay proxy and `test/snapshots/`
 - Generated types: update schema in `@github/copilot` then run `cd nodejs && npm run generate:session-types` and commit generated files in `src/generated` or language generated location.
+
+## Skills 🛠️
+
+Repo-scoped skills live under `.github/skills/<skill-name>/` and are auto-discovered by Copilot CLI. Load the relevant skill before editing the matching file types.
```
stephentoub: "and are auto-discovered by Copilot CLI"... is this needed? Other agents like CCA would also auto-discover it.

rust/src/session.rs
```diff
@@ -0,0 +1,1813 @@
...
+    event_tx: tokio::sync::broadcast::Sender<SessionEvent>,
+) -> JoinHandle<()> {
+    let crate::router::SessionChannels {
+        mut notifications,
+        mut requests,
+    } = channels;
+
+    tokio::spawn(async move {
+        loop {
+            tokio::select! {
```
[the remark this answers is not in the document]
tclem: Accepted. Wrapped the spawned future with `error_span!("session_event_loop", session_id = %session_id).instrument(...)` and pulled `Instrument` into scope so warn/error events from the loop correlate to a session.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/build.rs
```diff
@@ -0,0 +1,282 @@
+use std::io::Read;
+use std::path::Path;
+
+use sha2::Digest;
+
+fn main() {
+    println!("cargo:rerun-if-env-changed=COPILOT_CLI_VERSION");
+    println!("cargo:rerun-if-env-changed=BUNDLED_CLI_CACHE_DIR");
+    println!("cargo::rustc-check-cfg=cfg(has_bundled_cli)");
```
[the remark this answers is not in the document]
tclem: Declining: `cargo::rustc-check-cfg` (double colon) is the modern stable directive syntax accepted by Cargo since 1.77 — both `cargo:` (legacy) and `cargo::` (current) forms work, and the cfg is registered correctly. The crate already pins to Rust 1.94 via `rust-toolchain.toml`, so this is well within range.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/README.md
```diff
@@ -0,0 +1,544 @@
...
+
+```bash
+COPILOT_CLI_VERSION=1.0.15 cargo build
+```
+
+### How it works
+
+1. **Build time:** The SDK's `build.rs` detects `COPILOT_CLI_VERSION`, downloads the platform-appropriate binary from npm (`@github/copilot-{platform}`), verifies the tarball's SHA-512 integrity hash against npm's registry metadata, compresses with zstd, and embeds via `include_bytes!()`. No extra steps or tools needed — just the env var.
+
+2. **Runtime:** On the first call to `copilot::resolve::copilot_binary()`, the embedded binary is lazily extracted to `~/.cache/copilot-sdk/copilot_{version}`, SHA-256 verified, and cached. Subsequent calls return the cached path.
```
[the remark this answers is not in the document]
tclem: Accepted, good catch. The README was inherited from an earlier npm-based design; updated to match what `build.rs` and `embeddedcli.rs` actually do — GitHub Releases archives, SHA-256 against `SHA256SUMS.txt`, runtime cache at `~/.cache/copilot-sdk-{version}/copilot`. Also synced the stale doc comment on `embeddedcli::path`.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

test/scenarios/sessions/streaming/verify.sh
```diff
@@ -111,9 +111,13 @@ check "TypeScript (build)"   bash -c "cd '$SCRIPT_DIR/typescript' && npm run bui
 check "Python (install)" bash -c "python3 -c 'import copilot' 2>/dev/null || (cd '$SCRIPT_DIR/python' && pip3 install -r requirements.txt --quiet 2>&1)"
 check "Python (syntax)"  bash -c "python3 -c \"import ast; ast.parse(open('$SCRIPT_DIR/python/main.py').read()); print('Syntax OK')\""
 
+# Go: build
 # Go: build
 check "Go (build)" bash -c "cd '$SCRIPT_DIR/go' && go build -o streaming-go . 2>&1"
```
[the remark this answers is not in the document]
tclem: Accepted, dropped the duplicate.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

.github/workflows/scenario-builds.yml
```diff
@@ -185,3 +187,46 @@ jobs:
...
           fi
+
+  # ── Rust ────────────────────────────────────────────────────────────
+  build-rust:
+    name: "Rust scenarios"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@1.94.0
```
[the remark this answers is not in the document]
tclem: Declining: `dtolnay/rust-toolchain@1.94.0` is a documented and supported pinning form — the action's [README](https://github.com/dtolnay/rust-toolchain) explicitly states the toolchain defaults to match the `@rev`, so `@1.94.0` installs Rust 1.94.0. The other workflows that use `@stable` rely on `rust-toolchain.toml` for the pin instead, which is a different style but achieves the same effect. Either pattern is fine; I'd rather not rev workflows just for stylistic consistency.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/session.rs
```diff
@@ -0,0 +1,1817 @@
...
+/// Created via [`Client::create_session`] or [`Client::resume_session`].
+/// Owns an internal event loop that dispatches events to the [`SessionHandler`].
+///
+/// Protocol methods (`send_message`, `get_messages`, `abort`, etc.) automatically
+/// inject the session ID into RPC params.
+///
+/// Call [`destroy`](Self::destroy) for graceful cleanup (RPC + local). If dropped
+/// without calling `destroy`, the `Drop` impl aborts the event loop and
+/// unregisters from the router as a best-effort safety net.
+pub struct Session {
```
stephentoub, line 76: These are called CopilotClient and CopilotSession in the other SDKs. Just "Client" is very generic and would seem to introduce ambiguity, if not in actual source then in documentation and the like.

rust/src/lib.rs
```diff
@@ -0,0 +1,1170 @@
...
+    /// GitHub token for authentication. When set, the SDK passes the token
+    /// to the CLI via `--auth-token-env COPILOT_SDK_AUTH_TOKEN` and exports
+    /// the token in that env var. When set, the CLI defaults to *not*
+    /// using the logged-in user (override with [`Self::use_logged_in_user`]).
+    pub github_token: Option<String>,
+    /// Whether the CLI should fall back to the logged-in `gh` user when no
+    /// token is provided. `None` means use the runtime default (true unless
+    /// [`Self::github_token`] is set, in which case false).
+    pub use_logged_in_user: Option<bool>,
+}
```
stephentoub, line 617: This looks like it's missing a variety of things the other SDKs have, e.g. LogLevel, AutoStart, Telemetry, SessionFS, SessionIdleTimeoutSeconds, OnListModels, etc.

rust/src/lib.rs
```diff
@@ -0,0 +1,1170 @@
...
+    /// match tokio::time::timeout(
+    ///     std::time::Duration::from_secs(5),
+    ///     client.stop(),
+    /// ).await {
+    ///     Ok(_) => {}
+    ///     Err(_) => client.force_stop(),
+    /// }
+    /// # }
+    /// ```
+    pub fn force_stop(&self) {
```
stephentoub, line 1786: In the C# SDK this is asynchronous because cleanup may need to wait for a connection created by StartAsync to be established. That highlights that the construction mechanisms also differ here: C# has `new CopilotClient()` and then separately `StartAsync`, whereas this PR for rust has Client::start as a single async associated function. @SteveSandersonMS, what led us to separate construction + starting/resuming in that manner? Is there interaction with a client instance (e.g. setting properties) expected prior to starting/resuming it?

rust/src/types.rs
```diff
@@ -0,0 +1,1985 @@
...
+/// ```no_run
+/// # use copilot::session::Session;
+/// # use copilot::types::SendOptions;
+/// # async fn run(session: Session) -> Result<(), copilot::Error> {
+/// session.send_message(SendOptions::new("hello")).await?;
+/// # Ok(()) }
+/// ```
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct SendOptions {
```
stephentoub: The other SDKs call this MessageOptions.

rust/src/types.rs
```diff
@@ -0,0 +1,1985 @@
...
+    /// The user prompt to send.
+    pub prompt: String,
+    /// Optional permission mode for this turn (e.g. `"agent"`, `"autopilot"`).
+    pub mode: Option<String>,
+    /// Optional attachments to include with the message.
+    pub attachments: Option<Vec<Attachment>>,
+    /// Maximum time to wait for the session to go idle. Honored only by
+    /// `send_and_wait`. Defaults to 60 seconds when unset.
+    pub wait_timeout: Option<Duration>,
```
stephentoub, line 2457: Why is this on SendOptions rather than an argument to send_and_wait?

rust/src/types.rs
```diff
@@ -0,0 +1,1985 @@
...
+}
```
stephentoub, line 2473: This is missing the equivalent of RequestHeaders

rust/src/session.rs
```diff
@@ -0,0 +1,1817 @@
...
+            )
+            .await?;
+        let elicitation: ElicitationResult = serde_json::from_value(result)?;
+        Ok(elicitation)
+    }
+
+    /// Ask the user a yes/no confirmation question.
+    ///
+    /// Returns `true` if the user accepted and confirmed, `false` otherwise.
+    pub async fn confirm(&self, message: &str) -> Result<bool, Error> {
```
stephentoub, line 572: The C# SDK puts this into a UI group, e.g. session.UI.ConfirmAsync, rather than having it directly on session. That seems desirable from a growth perspective.

rust/src/types.rs
```diff
@@ -0,0 +1,1985 @@
...
+#[serde(rename_all = "lowercase")]
+pub enum ConnectionState {
+    /// No CLI process is attached or the process has exited cleanly.
+    Disconnected,
+    /// The client is starting up (spawning the CLI, negotiating protocol).
+    Connecting,
+    /// The client is connected and ready to handle RPC traffic.
+    Connected,
+    /// Startup failed or the connection encountered an unrecoverable error.
+    Errored,
```
stephentoub: Is this going to result in the wire data being "errored"? Doesn't it need to be "error"?

rust/src/session.rs
```diff
@@ -0,0 +1,1817 @@
...
+    /// `wait_timeout` field on `SendOptions` is ignored here (use
+    /// [`send_and_wait`](Self::send_and_wait) if you need to wait).
+    ///
+    /// Returns the assigned message ID, which can be used to correlate the
+    /// send with later [`SessionEvent`]s emitted in
+    /// response (assistant messages, tool requests, etc.).
+    ///
+    /// Returns an error if a [`send_and_wait`](Self::send_and_wait) call is
+    /// currently in flight, since the plain send would race with the waiter.
+    pub async fn send_message(&self, opts: impl Into<SendOptions>) -> Result<String, Error> {
```
stephentoub: .NET has a CancellationToken type, and the C# SDK does the idiomatic thing of accepting one into most methods in order to support canceling that individual operation. Does Rust / tokio not have something similar? For someone to cancel e.g. an in-flight send-message, would they need to kill the whole connection?

rust/src/types.rs
```diff
@@ -0,0 +1,1985 @@
...
+    /// `hooks` flag. Use [`with_hooks`](Self::with_hooks) to install one.
+    #[serde(skip)]
+    pub hooks_handler: Option<Arc<dyn SessionHooks>>,
+    /// System-message transform. When set, the SDK injects the matching
+    /// `action: "transform"` sections into the system message and routes
+    /// `systemMessage.transform` RPC callbacks to it during the session.
+    /// Use [`with_transform`](Self::with_transform) to install one.
+    #[serde(skip)]
+    pub transform: Option<Arc<dyn SystemMessageTransform>>,
+}
```
stephentoub, line 1112: Looks like this is missing GithubToken for per-session auth?

rust/Cargo.toml
```diff
@@ -0,0 +1,67 @@
+[package]
+name = "copilot-sdk"
```
stephentoub: There are unfortunately many different copilots now. Do we want to make it clear which one this is? The C# nuget package, for example, is GitHub.Copilot.SDK.

rust/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+    "release-plz.toml",
+    "rust-toolchain.toml",
+    ".rustfmt.toml",
+    ".rustfmt.nightly.toml",
+    "clippy.toml",
+    ".gitignore",
+]
+
+[lib]
+name = "copilot"
```
stephentoub: Similarly here... it's nice and concise, but might it not conflict with other copilots at MS?

stephentoub:
Thanks for getting this up.

It'd be helpful to run an agent over the other SDKs and this new one to look for any inconsistencies / gaps / divergences, so that we can then evaluate each and decide whether it's acceptable or should be addressed. The more consistent we can be across the SDKs, the easier it'll be to maintain them moving forward, the more information / docs about one will translate to consumption of the others, the better we'll be able to evolve with reduced concerns for how something we want to add may not fit well in a particular SDK, etc.

rust/src/session.rs
tclem: In Rust, "a future, on being dropped, is cancelled", but cancellation is a little complex and different in Rust compared to other languages. For example, while `send_message` is an async function, there are other places where we use `spawn` and `spawn_blocking` which return join handles that you can abort.

Some good reading: https://rfd.shared.oxide.computer/rfd/0400

rust/src/lib.rs
SteveSandersonMS: >  what led us to separate construction + starting/resuming in that manner?

Mainly just that:

 * Start/connect needs to be async
 * Constructors are a little more idiomatic than static factory functions in C#/TypeScript - it feels more approachable in basic samples
 * Start can be implicit anyway, so there's no need for devs to think about it in most cases. We implicitly await start in `sendAsync`

We could change this if we don't like it, but nobody's expressed dissatisfaction about it to my knowledge.

rust/Cargo.toml
tclem: Good callout. We're going to do `github_copilot_sdk` here as the module to import. Updated already.

rust/Cargo.toml
tclem: Good callout. We're going to do `github-copilot-sdk` here. Updated already.

rust/src/session.rs
tclem: This is canonical and idiomatic in Rust. Do you feel strongly about this? Here's one example for `reqwest`: https://docs.rs/reqwest/latest/reqwest/struct.Client.html

You don't generally repeat the namespace in the struct name. Kinda like Go avoids stuttering names.

.github/skills/rust-coding-skill/SKILL.md
tclem: It's not, brought this over from an internal project. Many of the models seem to be trained to write poor comments that are noisy and eventually confuse future models/agents.

I'm open to elevating it if you think it's broadly applicable, but maybe let's leave it here for the first iteration of this landing.

rust/src/types.rs
tclem: Fixed.

rust/src/lib.rs
tclem: Yes, and good eyes. I've got an agent working through a few parity gaps and it also identified this.

.github/copilot-instructions.md
tclem: Good catch -- broadened to "auto-discovered by Copilot" so it covers CCA and other Copilot agent tooling, not just the CLI. Pushed in cd8d6bb.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

tclem: is it needed though? can we delete the line?

rust/src/types.rs
tclem: Good catch, accepted. Go's source-of-truth at `go/types.go:17` uses `"error"` (not `"errored"`); the Rust `rename_all = "lowercase"` was producing `"errored"` and silently diverging. Added `#[serde(rename = "error")]` on the variant plus two unit tests locking the wire format. Variant name stays `Errored` to avoid shadowing `std::error::Error` and the crate's own `Error` type. Pushed in 3109b77.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

tclem: Per tclem follow-up: renamed the variant to `ConnectionState::Error` and dropped the `#[serde(rename = "error")]` attribute. Cleaner enum surface, and the `Error` name doesn't collide with anything in scope (no `use crate::Error` in `types.rs`; the variant lives inside the typed enum so no shadowing). Pushed in 9062771.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/lib.rs
tclem: All landed in this stack:

- `c4132c2` (Bucket B.2): `log_level`, `session_idle_timeout_seconds`, `auto_start` on `ClientOptions`
- `3b30d5a` (Bucket B.2): `on_list_models` BYOK callback
- `95a2ece` (Phase 4 § 4.2): `on_session_fs` (`SessionFsProvider` trait)
- `aefb108` (Phase 4 § 4.4): `telemetry: Option<TelemetryConfig>` env-var passthrough

`ClientOptions` surface now matches Node / Go / .NET for these knobs.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/types.rs
tclem: Landed as `8c00cc0` (Phase 4 § 4.5): `MessageOptions::request_headers: Option<HashMap<String, String>>` with `with_request_headers(...)` and `with_request_header(k, v)` builders. Mirrors Node `MessageOptions.requestHeaders` and Go `MessageOptions.RequestHeaders`.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/types.rs
tclem: `SessionConfig::github_token` exists at `rust/src/types.rs:759` and `ResumeSessionConfig::github_token` at `:1019`. Per-session auth is supported; both predate this stack.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/types.rs
```diff
@@ -0,0 +1,1969 @@
...
```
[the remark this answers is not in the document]
tclem: Landed as `2a8f4e8` (Phase 4 § 4.1): `SessionConfig::commands: Option<Vec<Command>>` plus `Command` type and `with_commands(...)` builder. Slash command registration now matches Node / Python / Go / .NET.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/types.rs
```diff
@@ -0,0 +1,1969 @@
...
+    /// start/end, etc.). When set, the SDK auto-enables the wire-level
+    /// `hooks` flag. Use [`with_hooks`](Self::with_hooks) to install one.
+    #[serde(skip)]
+    pub hooks_handler: Option<Arc<dyn SessionHooks>>,
+    /// System-message transform. When set, the SDK injects the matching
+    /// `action: "transform"` sections into the system message and routes
+    /// `systemMessage.transform` RPC callbacks to it during the session.
+    /// Use [`with_transform`](Self::with_transform) to install one.
+    #[serde(skip)]
+    pub transform: Option<Arc<dyn SystemMessageTransform>>,
```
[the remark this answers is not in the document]
tclem: Landed as `2a8f4e8` (Phase 4 § 4.1) - same fix as #3162816813. `SessionConfig::commands` + `Command` type are now supported.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/types.rs
tclem: Declining. The placement mirrors Node's `MessageOptions.waitTimeout` and Go's `MessageOptions.WaitTimeout`, and the doc-comment explicitly notes "Honored only by `send_and_wait`. Defaults to 60 seconds when unset." Keeping cross-SDK shape parity is more valuable than a single-SDK refactor here.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/session.rs
tclem: Done in `f4aa8d9`. New `SessionUi<'a>` view returned by `Session::ui()`; `confirm` / `select` / `input` / `elicitation` moved onto it. Now `session.ui().confirm(...)` matches .NET `session.UI.ConfirmAsync`, Python `session.ui.confirm`, Go `session.UI().Confirm()`. Wire-method names unchanged.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/session.rs
```diff
@@ -0,0 +1,1746 @@
...
+            )
+            .await?;
+        let elicitation: ElicitationResult = serde_json::from_value(result)?;
+        Ok(elicitation)
+    }
+
+    /// Ask the user a yes/no confirmation question.
+    ///
+    /// Returns `true` if the user accepted and confirmed, `false` otherwise.
+    pub async fn confirm(&self, message: &str) -> Result<bool, Error> {
```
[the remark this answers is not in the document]
tclem: Done in `f4aa8d9` - same fix as #3161986035. UI methods now grouped under `Session::ui()` returning `SessionUi<'a>`, mirroring .NET / Python / Go.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/lib.rs
```diff
@@ -0,0 +1,1178 @@
...
+    }
+
+    /// Get the CLI server status.
+    pub async fn get_status(&self) -> Result<serde_json::Value, Error> {
+        self.call("getStatus", Some(serde_json::json!({}))).await
+    }
+
+    /// Get authentication status.
+    pub async fn get_auth_status(&self) -> Result<serde_json::Value, Error> {
+        self.call("getAuthStatus", Some(serde_json::json!({})))
```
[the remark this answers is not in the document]
tclem: Done in `f4aa8d9`. `Client::get_status` now returns `GetStatusResponse { version, protocol_version }` and `Client::get_auth_status` returns `GetAuthStatusResponse { is_authenticated, auth_type, host, login, status_message }`. Both `#[non_exhaustive]`. Matches Node / Python / Go shapes.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/session.rs
```diff
@@ -0,0 +1,1824 @@
...
+            )
+            .await?;
+        Ok(())
+    }
+
+    /// Switch to a different model.
+    ///
+    /// Pass `None` for `opts` if no extra configuration is needed. Mirrors
+    /// Go's `Session.SetModel(ctx, model, *SetModelOptions)`.
+    pub async fn set_model(
```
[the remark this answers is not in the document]
tclem: Done in `f4aa8d9`. `Session::set_model` now returns `Result<(), Error>` to match Node / Python / Go / .NET void-on-success semantics.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

.github/skills/rust-coding-skill/SKILL.md
tclem: Cross-SDK consistency on test philosophy is a good thing in general, but this particular guidance reflects a real Rust-specific cost: in async Rust, mocking is more invasive than in C#/TS/Python because trait-based mocks (`mockall`, `mock_it`) erase concrete types behind `dyn Trait`, force `Pin<Box<dyn Future>>` returns, and break `&self`-borrow patterns the rest of the SDK relies on. The result is mock-aware code shapes that diverge from production code shapes — and that's the failure mode I want to avoid.

The SDK already exposes the seam this rule depends on: `Client::from_streams(reader, writer)` lets tests substitute any `AsyncRead`/`AsyncWrite` pair. Plus `MockServer` in `test/harness/` for the wire-level loop. Real implementations + real wire = real coverage, and `from_streams` keeps it cheap.

Happy to align cross-SDK if there's a place we want to centralize testing philosophy, but I'd want the policy to remain compatible with "we don't mock our own traits" on the Rust side specifically. Resolving as-is for 0.1.0.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

.github/skills/rust-coding-skill/SKILL.md
tclem: Agreed it's not Rust-specific. Keeping it here for the first iteration as tclem flagged; we can promote to a top-level repo doc + reference back from this skill once we have a place to land it. Resolving for 0.1.0.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/session.rs
tclem: Following up on tclem's reply: this is canonical Rust style and the prior art is broad. `reqwest::Client`, `redis::Client`, `tonic::client::Grpc`, `aws-sdk-s3::Client`, `octocrab::Octocrab` (one of the few that does prefix, and is widely cited as an outlier). The Rust convention is "the namespace IS the type qualifier" — `use github_copilot_sdk::Client` already conveys "Copilot CLI client" at the call site.

Stutter-rename precedent (`CopilotClient`, `CopilotSession`) is a Go-/C#-style hedge against `import` semantics where the namespace fades after the import line. Rust paths stay visible at every usage, so the prefix becomes redundant noise rather than a disambiguator.

Module re-exports keep the qualified path one tab away too: `github_copilot_sdk::Client` ↔ `github_copilot_sdk::session::Session`. Resolving as `Client`/`Session`.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/lib.rs
tclem: For the Rust side: the construction-vs-start split doesn't transfer cleanly because Rust doesn't have constructors as a language feature — every "constructor" is just an associated function (`new`, `with_*`, `start`, etc.). So the choice isn't `new() + start()` vs `start()` — it's "two associated fns" vs "one associated fn", and that ergonomic delta isn't worth a public API doubling.

A few specifics:

1. **`Client::start` is async because the work it does is async** — spawning a child process, writing the JSON-RPC handshake, awaiting the protocol-version response. There's no observable state to inspect or configure between "construct" and "start"; everything that influences startup goes through `ClientOptions`, which is a plain `pub struct` consumers fill in beforehand.
2. **No "interact with client before starting" expectation.** The `Client` value isn't usable until the underlying transport is connected, so a hypothetical pre-start `Client` would be a typestate that just delays compile errors to the first method call. Type-state is fine pattern but feels over-engineered for a binary "did I call start?" gate.
3. **Async drop is the dual concern.** I considered an async `disconnect`/`stop` for parity with C#'s async `Dispose`. Landed on sync `Drop` for the common case (best-effort kill on drop) plus an explicit `Client::stop().await` for callers that want to await reaper completion. `force_stop` is the synchronous escape hatch when the runtime is going away — this is the line stephentoub flagged. It's deliberately sync because it's used at shutdown when there's no runtime to await on.

If we want cross-SDK uniformity to be a stronger pull, the cheapest convergence is to add `Client::new(opts) -> Client` + `Client::connect(self) -> impl Future<...>` and have `start` be a one-shot convenience. Happy to do that as a 0.2.0 follow-up if it removes friction for cross-SDK readers, but I'd rather not split the public API at 0.1.0 without seeing demand for the sequencing (configure-then-connect) it enables.

Resolving for 0.1.0; happy to revisit if the constructor-vs-start asymmetry produces real friction.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

test/scenarios/RUST_COVERAGE.md
```diff
@@ -0,0 +1,60 @@
...
+| `tools/`       | `tool-filtering`        | ✅     |
+| `tools/`       | `custom-agents`         | ✅     |
+| `tools/`       | `tool-overrides`        | ✅     |
+| `tools/`       | `virtual-filesystem`    | ❌ deferred (needs `VirtualFilesystem` hook port) |
+| `callbacks/`   | `hooks`                 | ✅     |
+| `callbacks/`   | `permissions`           | ✅     |
+| `callbacks/`   | `user-input`            | ✅     |
+| `prompts/`     | `system-message`        | ✅     |
+| `prompts/`     | `reasoning-effort`      | ✅     |
+| `bundling/`    | *                       | ❌ app-level concern, not an SDK gap |
```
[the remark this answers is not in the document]
tclem: Good catch — accidental omission, not a deliberate deferral. Added the `prompts/attachments/rust/` scenario in the next commit, mirroring the TypeScript / Python / Go / C# implementations:

```rust
let response = session.send_and_wait(
    MessageOptions::new("What languages are listed in the attached file?")
        .with_attachments(vec![Attachment::File {
            path: sample_file,
            display_name: None,
            line_range: None,
        }]),
).await?;
```

Updated `RUST_COVERAGE.md` to ✅, added the Rust syntax row to the README's File/Blob attachment tables, and wired the Rust build/run steps into `verify.sh`. Local `cargo build` is clean.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/README.md
```diff
@@ -0,0 +1,663 @@
...
+    BinaryNotFound { name, hint }, // CLI binary not found
+}
+
+// Check if the transport is broken (caller should discard the client)
+if err.is_transport_failure() {
+    client = Client::start(options).await?;
+}
+```
+
+## Differences From Other SDKs
```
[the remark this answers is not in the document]
tclem: Good catch on cross-SDK README parity. Added six new brief sections to the API Reference between "Capabilities & Elicitation" and "Progress Reporting", each with a runnable code example:

- **Slash Commands** — `CommandHandler` + `CommandDefinition::new(name, handler).with_description(...)`
- **User Input Requests** — `SessionHandler::on_user_input` example
- **Streaming** — `streaming: true` + delta-event match loop reading `event.data.get("delta")`
- **Infinite Sessions** — `InfiniteSessionConfig::workspace_path`, compaction events, resume-via-`Client::resume_session`
- **Custom Providers (BYOK)** — `ProviderConfig.provider_type` / `base_url` / `bearer_token` / `headers`, with `openai`/`azure`/`anthropic` kinds
- **Telemetry** — `TelemetryConfig` + `OtelExporterType::OtlpHttp`/`File` with the env-var injection note

All shapes verified against the actual API (caught one bug while writing — `ProviderConfig.provider_type: Option<String>` not `kind`, `base_url: String` not `Option`). Resolving with the next commit pushed.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/README.md
```diff
@@ -0,0 +1,663 @@
...
+none of them are scheduled for removal.
+
+- **`Client::get_quota`** — fetch account-level request quota snapshots.
+  The underlying `account.getQuota` JSON-RPC endpoint is exposed only by
+  the Rust SDK in this release.
+- **First-class `Session` convenience methods** — `set_mode` / `get_mode`,
+  `set_name` / `get_name`, `read_plan` / `update_plan` / `delete_plan`,
+  `start_fleet`, `list_workspace_files` / `read_workspace_file` /
+  `create_workspace_file`. The other SDKs require the consumer to drive
+  the typed JSON-RPC namespace directly for these.
```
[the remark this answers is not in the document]
tclem: Both confirmed real (`Session::get_model` at `session.rs:336`, `Client::send_telemetry` at `lib.rs:1335`, `Session::send_telemetry` at `session.rs:484`) and now in the inventory:

- `get_model` folded into the existing "First-class Session convenience methods" bullet.
- `send_telemetry` (both Client and Session variants) added as a new dedicated bullet — these are higher-divergence than the other Rust-only methods because no other SDK exposes the underlying RPC endpoints in their public API at all (not even via typed namespaces), so they get their own line.

Pushed in `9c055aa`. Resolving.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/lib.rs
```diff
@@ -0,0 +1,1930 @@
...
+            "session.setForeground",
+            Some(serde_json::json!({ "sessionId": session_id })),
+        )
+        .await?;
+        Ok(())
+    }
+
+    /// Get the CLI server status.
+    pub async fn get_status(&self) -> Result<GetStatusResponse, Error> {
+        let result = self.call("getStatus", Some(serde_json::json!({}))).await?;
```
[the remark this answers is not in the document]
tclem: Confirmed and fixed in `0f6a2ed`. Verified the CLI runtime registration in `@github/copilot@1.0.39` app.js — both `"status.get"` and `"auth.getStatus"` are registered; `"getStatus"` and `"getAuthStatus"` are not. Schema doesn't currently define these top-level RPCs typed (only `session.auth.getStatus` exists in the schema, a different namespaced endpoint), so codegen didn't catch the hand-authored layer's drift.

Fix shape mirrors the elicitation (`c58e2f2`) and list_sessions (`64541af`) wire-fixes:

- `lib.rs:1301`: `"getStatus"` → `"status.get"`
- Test now asserts canonical name AND `assert_ne!(request["method"], "getStatus")` to prevent the same regression silently
- CHANGELOG entry under `### Fixed`

Same root cause as the prior two: hand-authored RPC wrapper + mock-server test asserting on the implementation's wrong-name string. Brings the wire-shape-audit save count to 4. The regression-prevention asserts are in place across all four sites now.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/lib.rs
```diff
@@ -0,0 +1,1930 @@
...
+        Ok(serde_json::from_value(result)?)
+    }
+
+    /// Get authentication status.
+    pub async fn get_auth_status(&self) -> Result<GetAuthStatusResponse, Error> {
+        let result = self
+            .call("getAuthStatus", Some(serde_json::json!({})))
```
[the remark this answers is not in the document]
tclem: Confirmed and fixed alongside the `getStatus` sibling in `0f6a2ed`. Same root cause, same fix shape: `lib.rs:1308`: `"getAuthStatus"` → `"auth.getStatus"`, regression-prevention assert added to the shared test, CHANGELOG entry covers both methods together.

Both methods had been silently dead since the hand-authored layer landed; the mock-server test round-tripped through the wrong names on both ends, so the bug class is identical to the elicitation `requestedSchema` and `list_sessions` filter-wrapping wire fixes earlier in this stack.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/lib.rs
```diff
@@ -0,0 +1,1962 @@
...
+            }
+            Err(error) => Err(error),
+        }
+    }
+
+    /// Fetch account-level quota snapshots (request-based usage).
+    ///
+    /// **Rust-only as of 0.1.0.** This method has no equivalent in the Node,
+    /// Python, Go, or .NET SDKs at the time of this release; the underlying
+    /// `account.getQuota` JSON-RPC endpoint is exposed only by the Rust SDK.
```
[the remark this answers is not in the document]
tclem: Good catch, accepting. You are right that the `account.getQuota` RPC is generated and exposed in all five SDKs via their typed `rpc()` namespace; what is Rust-only is the top-level `Client::get_quota` convenience wrapper.

Verified cross-SDK exposure:
- Node: `nodejs/src/generated/rpc.ts:2203` (`account.getQuota` in the `rpc()` namespace)
- Python: `python/copilot/generated/rpc.py:5647` (`rpc().account.get_quota`)
- Go: `go/rpc/generated_rpc.go:2252` (`RPC.Account.GetQuota`)
- .NET: `dotnet/src/Generated/Rpc.cs:3140` (`Rpc().Account().GetQuotaAsync`)
- Rust: `rust/src/generated/rpc.rs:90` (`rpc().account().get_quota`)

Fixed in e04357c: updated the rustdoc on `Client::get_quota` in `rust/src/lib.rs` and the "Rust-only API" bullet in `rust/README.md` to make clear the wrapper is Rust-only while the underlying endpoint is cross-SDK.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

.github/skills/rust-coding-skill/examples.md
```diff
@@ -0,0 +1,170 @@
...
+        info!(event_type = ?event.kind, "session event");
+    }
+}.instrument(span));
+```
+
+## Concurrent permission handlers
+
+`HandlerEvent::PermissionRequest` and `HandlerEvent::ExternalTool` are dispatched
+on spawned tasks (see `rust/src/session.rs:973` and `:1022`). Implementations
+must be safe for concurrent invocation.
```
stephentoub, line 86: Is this expected? I assume this means we've buried some unsafe usage and we're then putting it on the developer to know when they're using shared state and the borrowed checker won't help them?

.github/skills/rust-coding-skill/SKILL.md
```diff
@@ -0,0 +1,254 @@
...
+tokio::spawn(async move { run_loop().await }.instrument(span));
+```
+
+Log with structured fields: `info!(session_id = %id, "Session created")`.
+Static messages stay greppable; dynamic data goes in named fields, not
+interpolated into the message string.
+
+## Idioms that don't port from Go or Node
+
+The most common pitfall when adapting code from the Node and Go SDKs is the
```
stephentoub: Curious the lack of C# mentioned

.github/copilot-instructions.md
```diff
@@ -48,7 +48,13 @@
...
+- SDK code: `nodejs/src`, `python/copilot`, `go`, `dotnet/src`, `rust/src`
+- Unit tests: `nodejs/test`, `python/*`, `go/*`, `dotnet/test`, `rust/tests`
 - E2E tests: `*/e2e/` folders that use the shared replay proxy and `test/snapshots/`
 - Generated types: update schema in `@github/copilot` then run `cd nodejs && npm run generate:session-types` and commit generated files in `src/generated` or language generated location.
+
+## Skills 🛠️
+
+Repo-scoped skills live under `.github/skills/<skill-name>/` and are auto-discovered by Copilot. Load the relevant skill before editing the matching file types.
+
+- **`rust-coding-skill`** (`.github/skills/rust-coding-skill/SKILL.md`) — load before editing any `*.rs` file in `rust/`. Covers error handling, async/concurrency, tracing, codegen workflow, and Rust SDK-specific trait patterns.
```
stephentoub: I'm surprised this is necessary or helpful. Shouldn't these skills be discovered automatically, and doesn't the frontmatter in the skill provide the necessary guidance on when to use it? Having this here seems like unnecessary pollution of the system prompt?

rust/src/generated/rpc.rs
```diff
@@ -0,0 +1,1359 @@
...
+
+/// `session.agent.*` RPCs.
+#[derive(Clone, Copy)]
+pub struct SessionRpcAgent<'a> {
+    pub(crate) session: &'a Session,
+}
+
+impl<'a> SessionRpcAgent<'a> {
+    /// Wire method: `session.agent.list`.
+    /// Stability: `experimental`.
```
stephentoub: Is this just documentation, or does this get recognized by the tooling in some way to let someone know they're using something experimental?

rust/src/generated/session_events.rs
```diff
@@ -0,0 +1,2767 @@
...
+/// A session event with typed data payload.
+///
+/// The common event fields (id, timestamp, parentId, ephemeral) are
+/// available directly. The event-specific data is in the `payload` field
+/// as a [`SessionEventData`] enum.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TypedSessionEvent {
+    /// Unique event identifier (UUID v4).
+    pub id: String,
```
stephentoub, line 346: Is there a uuid/guid type that this should map to instead of string? The schema includes a format specifier that would let you determine that, if applicable.

rust/src/generated/session_events.rs
```diff
@@ -0,0 +1,2767 @@
...
+    /// ISO 8601 timestamp when the event was created.
+    pub timestamp: String,
```
stephentoub, line 348: Similarly here: is there a datetime type that could be used rather than string? The schema includes a format specifier to indicate when that'd be appropriate.

.github/copilot-instructions.md
tclem: I agree, I'm not convinced that it is! Let me remove this entire "Skills" section.

rust/src/generated/session_events.rs
tclem: There is a [`uuid`](https://docs.rs/uuid/latest/uuid/) crate we could bring in as a dependency. I would probably skip it if this is just an opaque identifier - but if uuid v4 has meaning and we want the strong typing it's a well known and trusted dependency.

What do you think?

rust/src/generated/session_events.rs
tclem: Same as `uuid`, we could bring in a dependency... if we think it's worth it. `std::time` is intentionally minimal and free of calendar complexity, timezones, etc.

Is this important for the sdk?

rust/src/generated/session_events.rs
stephentoub: Meh, probably not worth it. For C#, Guid is the canonical answer and is in the core library, but with Rust requiring a separate crate, string should be fine.

rust/src/generated/session_events.rs
stephentoub: I think these timestamp ones are more worthwhile, so I'd go for this one if it's easy enough. Folks are more likely to actually use this data, compare time stamps, etc.

rust/src/generated/rpc.rs
tclem: I'll have my agent respond too. It's just docs, there are some options to make it stronger, but it sounds like this sort of annotation isn't done consistently in the other languages right now...

tclem: Today, just documentation -- the schema's `"stability": "experimental"` flag is read by codegen and emitted as a rustdoc comment.

I just upgraded the rendering: instead of a single-line `Stability: experimental` note, codegen now emits a styled rustdoc warning admonition (`<div class="warning">...</div>`) so docs.rs renders a yellow callout box on every experimental method. Matches the visibility Python's Sphinx output already gets.

Stable Rust doesn't have a first-class `#[experimental]` attribute (`#[unstable]` is rustc-internal, std-only). If we want compile-time opt-in friction, the ecosystem patterns are:

- **Cargo feature gate** (e.g. `tokio_unstable`-style `features = ["experimental"]`) -- consumers can't call experimental methods without explicitly enabling the feature. Strongest signal.
- **`unstable` module path** -- e.g. `session.rpc().unstable().agent().list()`. Visible in the import.

Both are bigger codegen changes that affect every experimental RPC across all SDKs' surface. Punting on those for 0.1.0; happy to revisit in a follow-up if we want the harder opt-in.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

.github/skills/rust-coding-skill/examples.md
tclem: No, this is confusing.

Not `unsafe`, that has a different meaning... this section is confusing... asking copilot to rewrite now.

tclem: Good catch on the framing -- it's misleading. The compiler does enforce this; nothing buried.

`SessionHandler` declares `Send + Sync + 'static` (`rust/src/handler.rs:280`), and `Arc<dyn SessionHandler>` is cloned into `tokio::spawn(async move { ... })` for the concurrent dispatch paths. The `RefCell` example in the doc literally won't compile -- `RefCell<T>` is `!Sync`, so the `impl SessionHandler for MyHandler` line fails the trait bound. The borrow checker is doing the work; the doc is just teaching "use `parking_lot::Mutex` here so you don't waste time reaching for `RefCell` first."

Updated `examples.md` to make this explicit -- renamed the anti-pattern to "Won't compile", showed the actual trait-bound error, and added a one-paragraph note saying the compiler is what enforces it. The rejection should be obvious from the doc now without a reader having to mentally trace `spawn -> Send + Sync -> RefCell !Sync`.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

.github/skills/rust-coding-skill/SKILL.md
tclem: This is probably keying off of my guidance and copilot giving up after looking at only a couple of examples. No python mentioned either.

Honestly, I'm not sure how helpful this is, but I do know that copilot's first pass at this was to write some very non-idiomatic rust. You have to prompt it to look at what the community projects do and what they avoid. This language can probably be tightened up...

.github/copilot-instructions.md
tclem: Removed.

rust/src/generated/session_events.rs
tclem: We'll need `chrono` for that.. which is a pretty big dependency to bring in for this capability... Let me audit in a bit more detail how many timestamps we're dealing with here and keep in might usage (like comparison) and see if I can propose something a bit more lightweight.

Ideally, we push a few 3rd party deps as possible on consumers.

tclem: https://crates.io/crates/time-format is the other options... though I kinda wonder if serde already has a dependency on one of these...

stephentoub: If it's going to be a problem, we can skip. C#, Go, and Python do all special-case this, though (TypeScript doesn't).

tclem: OK. If you feel strongly that we should make this a real type and not a string. My recommendation is [time-format](https://crates.io/crates/time-format) - zero dependencies, tiny (575 SLoC). [chrono](https://crates.io/crates/chrono) is 20K SLoC with a big list of dependencies.

.github/skills/rust-coding-skill/SKILL.md
tclem: Good catch — and your hunch that the framing was lopsided is right. Let me come back with a thorough rewrite.

Doing a proper research pass across the section's four claims (event subscription, cancellation, optional fields, serde) and auditing the SDK's current code against them. The cancellation guidance in particular is too absolute — `tokio_util::sync::CancellationToken` is real and idiomatic for SDK-internal task coordination (it's what `tonic` does). Will follow up here once the doc rewrite + matching SDK refactor land.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/generated/session_events.rs
stephentoub: I don't feel strongly about it. Up to you whether you think it's worth it.

.github/skills/rust-coding-skill/SKILL.md
tclem: Done. Two commits landed addressing this and your original "lopsided framing" callout:

- [`e96aea6`](https://github.com/github/copilot-sdk/commit/e96aea6) -- rewrites the SKILL "Idioms that don't port" section.
- [`7202dfd`](https://github.com/github/copilot-sdk/commit/7202dfd) -- refactors `Session.shutdown` from `Arc<Notify>` to `tokio_util::sync::CancellationToken` and exposes `Session::cancellation_token()`.

**Section is now language-agnostic** -- title is "Idioms that don't port from other languages", lists Node / Python / C# / Go equivalents inline where pattern divergence is notable.

**Cancellation guidance rewritten.** The previous framing ("drop the future or call abort() -- there is no `ctx.Done()` analogue") was wrong. Drop-to-cancel is the primitive, but `tokio_util::sync::CancellationToken` IS the Rust analog to `ctx.Done()` / .NET's `CancellationToken`, just optional and explicit per-API. The right framing is two cases:

1. **Caller-owned futures** (`send_message`, `send_and_wait`, subscription streams): drop the future. `select!` / `timeout` / drop give the caller everything they need; adding a token parameter just duplicates that. This is what `reqwest` / `sqlx` / `aws-sdk-*` / `tonic`'s client side do.
2. **SDK-internal task coordination** (event loop, subprocess reader, anything `tokio::spawn`ed): use `CancellationToken`. This is what `tonic` does for client-disconnect propagation. `tokio-graceful-shutdown` builds an entire framework on it.

The SDK now follows pattern 2 internally: `Session.shutdown` is a `CancellationToken` (refactored from the previous `Arc<Notify>`), and there's a public `Session::cancellation_token()` returning a child token so power users can bind their own work to the session lifetime via `select!`. Cancelling the child does NOT shut down the session -- isolation by design.

**Other three idioms** (event subscription, optional fields, serde JSON) -- guidance was correct but incomplete; doc enrichment landed in `e96aea6` with the full primitive matrix, `#[non_exhaustive]` + builder pairing, and `skip_serializing_if`/`default` notes. No code refactor needed for those -- already idiomatic.

**Citations for the cancellation framing**:
- [tokio-util CancellationToken docs](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html)
- [tonic cancellation example](https://github.com/hyperium/tonic/blob/master/examples/src/cancellation/server.rs)
- [withoutboats: Asynchronous clean-up](https://without.boats/blog/asynchronous-clean-up/)
- [Cybernetist: Rust tokio task cancellation patterns](https://cybernetist.com/2024/04/19/rust-tokio-task-cancellation-patterns/)

Validation green: 237 tests (was 233; +4 from new builder + cancellation_token tests), full clippy / fmt / doc -D warnings clean.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/types.rs
```diff
@@ -0,0 +1,3560 @@
...
+    /// Session-level event handler. See [`SessionConfig::handler`].
+    #[serde(skip)]
+    pub handler: Option<Arc<dyn SessionHandler>>,
+    /// Session hook handler. See [`SessionConfig::hooks_handler`].
+    #[serde(skip)]
+    pub hooks_handler: Option<Arc<dyn SessionHooks>>,
+    /// System-message transform. See [`SessionConfig::transform`].
+    #[serde(skip)]
+    pub transform: Option<Arc<dyn SystemMessageTransform>>,
+}
```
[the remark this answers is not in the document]
tclem: Real gap, fixed in [`59557da`](https://github.com/github/copilot-sdk/commit/59557da). Added `pub continue_pending_work: Option<bool>` to `ResumeSessionConfig` with a `with_continue_pending_work` builder, plus a wire-shape test asserting the field serializes to `continuePendingWork` and is omitted when unset.

One small note on the suggested diff: the manual `if let Some(v) ...` payload-construction code in `Client::resume_session` isn't needed -- `resume_session` builds the wire payload via `serde_json::to_value(&config)`, so the container `rename_all = "camelCase"` plus `skip_serializing_if = "Option::is_none"` get us the right wire shape automatically. Verified with the new wire test.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

stephentoub:
> request_auto_mode_switch + on_auto_mode_switch handler —

Were these recently added in runtime? We haven't added anything for them yet in the other SDKs. Could we remove them from rust for now?

> disabled_mcp_servers 

Same here. This isn't exposed in any of the other SDKs.

If the desire ends up being to have this, we can add it across all of them, including Rust, in one fell swoop.

> env_value_mode 

Ditto. I'm not even sure what this is :)

stephentoub (approved):
Thanks!

stephentoub:
(We will follow-up in a subsequent PR addressing the remaining feedback.)

PR #1181, opened 2026-04-30
fix(dotnet): add AgentId to SessionEvent base class

### LINKED ISSUE #1169: ParentToolCallId obsolete
Sorry @SteveSandersonMS creating a new one for this #1148 as i cant reopen it and dont want it getting lost.
______

Upgrading from 0.2.2 to 0.3.0 and im getting a warning that `ParentToolCallId` is obsolete - `This member is deprecated and will be removed in a future version.`.

Image

There isnt anything in the docs/ release notes about it, but what is the recommended way to link message/ and tool calls to subagent?

As in my UI I currently link messages and tool calls to their parent to create a live running tree - 

Image

_____

As mentioned on the other issue a new property has been added called `AgentId` but it doesn't look like that's been propagated down to anything.

Currently, I'm using `ParentToolCallId` in 2 places -
- From [`AssistantMessageEvent`](https://github.com/github/copilot-sdk/blob/c63feb2794786342d57936c13d28c250e723c676/dotnet/src/Generated/SessionEvents.cs#L512) where the `ParentToolCallId` is within its data property [`AssistantMessageData`](https://github.com/github/copilot-sdk/blob/c63feb2794786342d57936c13d28c250e723c676/dotnet/src/Generated/SessionEvents.cs#L1821)
- From [`ToolExecutionStartEvent`](https://github.com/github/copilot-sdk/blob/c63feb2794786342d57936c13d28c250e723c676/dotnet/src/Generated/SessionEvents.cs#L590) where the `ParentToolCallId` is within its data property [`ToolExecutionStartData`](https://github.com/github/copilot-sdk/blob/c63feb2794786342d57936c13d28c250e723c676/dotnet/src/Generated/SessionEvents.cs#L2011)

Neither of the above has an `AgentId` property

Looking at the source code looks like `AgentId` has only been added to `SystemNotificationAgentCompleted` and `SystemNotificationAgentIdle` 
____

Im using the .NET SDK

IeuanWalker:
Add the AgentId property to the SessionEvent base class in the .NET SDK, matching the agentId field present on all event interfaces in the Node.js generated types. This allows consumers to identify which sub-agent emitted a given event, replacing the deprecated ParentToolCallId pattern.

Fixes #1169

### DIFF
--- a/dotnet/src/Generated/SessionEvents.cs
+++ b/dotnet/src/Generated/SessionEvents.cs
@@ -119,6 +119,11 @@ public partial class SessionEvent
     [JsonPropertyName("ephemeral")]
     public bool? Ephemeral { get; set; }
 
+    /// <summary>Sub-agent instance identifier. Absent for events from the root/main agent and session-level events.</summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    [JsonPropertyName("agentId")]
+    public string? AgentId { get; set; }
+
     /// <summary>
     /// The event type discriminator.
     /// </summary>
--- a/dotnet/test/ForwardCompatibilityTests.cs
+++ b/dotnet/test/ForwardCompatibilityTests.cs
@@ -97,4 +97,41 @@ public class ForwardCompatibilityTests
 
         Assert.Equal("unknown", evt.Type);
     }
+
+        [Fact]
+    public void FromJson_WithAgentId_DeserializesAgentId()
+    {
+        var json = """
+        {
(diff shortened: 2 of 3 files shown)


IeuanWalker:
fyi, i've tested this change in my app and the agentId/ nesting is working as expected - 
image

stephentoub:
Thanks for calling out the issue. I've opened https://github.com/github/copilot-sdk/pull/1184 to fix the generator instead so that this doesn't reoccur the next time a new shared session event property is added.

PR #1184, opened 2026-05-01
Derive session event envelopes from schema

stephentoub:
Session-event wrappers in C#, Go, and Python were relying on hard-coded sets of top-level envelope properties, so schema-level fields like `agentId` could be omitted from language base event models. This derives shared envelope fields from the `SessionEvent` schema variants and shares that schema discovery across generators, keeping SDKs aligned as the envelope evolves.

- Add shared codegen helpers to resolve `SessionEvent` variants and collect envelope properties present on every variant.
- Update C#, Go, and Python generated session-event models to include and preserve top-level `agentId` while keeping language-specific emission local.
- Add round-trip and forward-compatibility coverage for top-level `agentId` on known and unknown event types.
- Leave Node/TypeScript unchanged because its generated event interfaces already include top-level `agentId`.

### DIFF
--- a/dotnet/src/Generated/SessionEvents.cs
+++ b/dotnet/src/Generated/SessionEvents.cs
@@ -103,22 +103,27 @@ namespace GitHub.Copilot.SDK;
 [JsonDerivedType(typeof(UserMessageEvent), "user.message")]
 public partial class SessionEvent
 {
-    /// <summary>Unique event identifier (UUID v4), generated when the event is emitted.</summary>
-    [JsonPropertyName("id")]
-    public Guid Id { get; set; }
-
-    /// <summary>ISO 8601 timestamp when the event was created.</summary>
-    [JsonPropertyName("timestamp")]
-    public DateTimeOffset Timestamp { get; set; }
-
-    /// <summary>ID of the chronologically preceding event in the session, forming a linked chain. Null for the first event.</summary>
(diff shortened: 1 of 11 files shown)


SteveSandersonMS:
@stephentoub This looks great!

Since Copilot mentions that a similar fix would help Go/Python codegen, I wonder if it's worth moving some of this "shared property" utility code into the shared utils file and consuming it from the other codegenerators too.

Longer term - probably after May - I'm super keep to rip up most of our per-language codegen logic and replace the whole thing with something a lot more coherent. Right now it's so expensive to have to keep extending codegen with more and more logic on a per-language basis. Longer term I would love to pick whichever one of our codegenerators is best (probably the C# one) and generalize it into something that maps the JSON schema an intermediate representation of all the types/properties/methods, and then have very simple per-language generators that render that out into the syntax for each language. As far as I can tell, there isn't *that* much per-language semantic difference to worry about so it seems like it should work.

stephentoub:
> Since Copilot mentions that a similar fix would help Go/Python codegen, I wonder if it's worth moving some of this "shared property" utility code into the shared utils file and consuming it from the other codegenerators too.

Yeah, I'm cleaning that up now. I hadn't realized the others were missing it as well until I saw copilot's review.

> replace the whole thing with something a lot more coherent.

Sounds good.

PR #1201, opened 2026-05-05
Make agent reload test runtime-compatible

stephentoub:
The agent reload E2E tests were brittle against live runtime behavior because they assumed fixed custom agent names and, in some SDKs, a specific post-reload treatment of session-configured agents. In copilot-agent-runtime, the live runtime can preserve the session-configured agent after reload, while the replay runtime may return a different post-reload set.

This updates the C#, Go, Python, and Node.js reload coverage to use unique reload agent names per run, verifies the session-configured agent is present before reload, and then validates the portable contract: `session.agent.reload` succeeds and returns the same agent set exposed by a subsequent `session.agent.list` call.

Validation:

```powershell
dotnet test dotnet\test\GitHub.Copilot.SDK.Test.csproj --filter "FullyQualifiedName=GitHub.Copilot.SDK.Test.E2E.RpcAgentE2ETests.Should_Call_Agent_Reload" --no-restore --logger "console;verbosity=minimal"
```

```powershell
Set-Location go; go test ./internal/e2e -run "TestAgentSelectionRpcE2E/should_call_agent_reload"
```

```powershell
Set-Location python; python -m pytest e2e\test_agent_and_compact_rpc_e2e.py -k test_should_call_agent_reload
```

```powershell
Set-Location nodejs; npm test -- test\e2e\agent_and_compact_rpc.e2e.test.ts -t "should call agent reload"
```

### DIFF
--- a/dotnet/test/E2E/RpcAgentE2ETests.cs
+++ b/dotnet/test/E2E/RpcAgentE2ETests.cs
@@ -2,6 +2,7 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------------------------------------------*/
 
+using GitHub.Copilot.SDK.Rpc;
 using GitHub.Copilot.SDK.Test.Harness;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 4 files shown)


stephentoub:
> The same issue exists in the Go and Python E2E counterparts - they should get the same treatment.

Applied this across Go and Python as well: both reload tests now use unique agent names and assert the reload response matches a subsequent agent list result.

stephentoub:
> The equivalent test is missing from the Node.js SDK.

Added the Node.js reload E2E coverage too. It now follows the same pattern: unique reload agent name, pre-reload metadata assertion, and reload/list consistency check.

PR #1214, opened 2026-05-06
Add enableSessionTelemetry option to all SDKs

Davsterl:
Adds an optional enableSessionTelemetry field to SessionConfig and ResumeSessionConfig across Node.js, Python, Go, and .NET SDKs.

When omitted (default), session telemetry remains enabled — no extra parameter is needed. When set to false, internal session telemetry is disabled.

Includes unit tests for serialization, cloning, and wire forwarding in all four languages.


dotnet/src/Types.cs
```diff
@@ -1930,6 +1931,11 @@ protected SessionConfig(SessionConfig? other)
     /// </summary>
     public ProviderConfig? Provider { get; set; }
 
+    /// <summary>
+    /// When false, disables internal session telemetry for this session.
```
stephentoub: Can we say anything more about what this is / when it applies / what its default is / etc.? It'd also be good to clarify that this is unrelated to the otel settings in Telemetry.

dotnet/test/Unit/CloneTests.cs
```diff
@@ -337,4 +352,24 @@ public void ResumeSessionConfig_Clone_PreservesContinuePendingWorkDefault()
...
+    [Fact]
+    public void ResumeSessionConfig_Clone_PreservesEnableSessionTelemetryDefault()
+    {
+        var original = new ResumeSessionConfig();
+
+        var clone = original.Clone();
+
+        Assert.Null(clone.EnableSessionTelemetry);
+    }
 }
```
stephentoub, line 375: Is there any kind of E2E test we can put in place that validates it's not sending telemetry when false, when BYOK is used, etc.?

Davsterl: Not the BYOK aspect in the SDK. It's covered well in the CLI unit tests.

dotnet/test/E2E/ClientOptionsE2ETests.cs
```diff
@@ -159,6 +159,63 @@ public async Task Should_Propagate_Process_Options_To_Spawned_Cli()
...
+        {
+            OnPermissionRequest = PermissionHandler.ApproveAll,
+        });
+
+        using var capture = JsonDocument.Parse(await File.ReadAllTextAsync(capturePath));
+        var createRequest = GetCapturedRequestParams(capture.RootElement, "session.create");
+        Assert.False(createRequest.TryGetProperty("enableSessionTelemetry", out _));
+
+        await session.DisposeAsync();
+    }
```
stephentoub, line 217: Rather than a test that validates the setting is sent to the CLI, I was wondering if there would be any way we could validate that it actually had the desired impact. But maybe there's no good way for us to observe that?

nickfyson:
👋 Hello!

Is it likely to be possible to get this merged today, and then into a published version of the SDK? 😄

(Context – we're waiting on this functionality for our work on TS Autofind, and it's one of the blockers to starting our dark ship of the re-platforming onto CCA.)

dotnet/test/E2E/ClientOptionsE2ETests.cs
Davsterl: There's no easy way :(

stephentoub:
@Davsterl, could you make the docs clarifying change for the other languages as well? Then we can get this merged. Thanks.

stephentoub:
I'm not sure why the node tests are failing; it looks like infrastructure and not being able to get an auth token. I'm going to try submitting your commits in a new PR...

stephentoub:
@Davsterl, I cherry-picked your commits into https://github.com/github/copilot-sdk/pull/1224, which is passing CI. I'm going to close this PR and go with that one, which also addresses the code review that highlighted because of things racing in CI, the Rust SDK was missing the new members.

PR #1212, opened 2026-05-06
Align Rust SDK public surface

stephentoub:
This follow-up from the cross-language SDK surface audit keeps Rust aligned with the established C#, Go, Python, and TypeScript public APIs before the new SDK surface settles.

## Summary

- Remove Rust-only high-level `autoModeSwitch`, `exitPlanMode`, and `disabled_mcp_servers` surfaces while leaving generated protocol event types untouched.
- Add Rust `ResumeSessionConfig` parity for `available_tools` and `disabled_skills`.
- Expand Rust tool results with binary payloads, tool telemetry, and an MCP `CallToolResult` conversion helper to match the richer tool-result support in the other SDKs.
- Update Rust README examples to use only valid high-level methods or generated `session.rpc()` calls, and fix Go permission handler comments.

PR #1215, opened 2026-05-06
Internalize env_value_mode (cross-SDK parity)

tclem:
Follow-up to #1212. Internalizes the last Rust-only knob on `SessionConfig` / `ResumeSessionConfig` that hadn't already been cleaned up by the cross-SDK surface audit.

## What changes

`env_value_mode` was a `pub Option<String>` on both configs, exposed via a `with_env_value_mode` builder. All four other SDKs unconditionally send `envValueMode: "direct"` on every `session.create` and `session.resume` and never expose it as a knob:

- Node `nodejs/src/client.ts:813`, `:959`
- Python `python/copilot/client.py:1431`, `:1740`
- Go `go/client.go:636`, `:826`
- .NET `dotnet/src/Client.cs:578`, `:707`

Now Rust matches: the field is `pub(crate) String` on both configs with `#[serde(default = "default_env_value_mode", skip_deserializing)]`, so serde always emits `"direct"` on the wire and any inbound JSON value is forced back to `"direct"` on round-trip. The public `with_env_value_mode` builder is gone.

## Other touch-ups

- Simplified the `McpStdioServerConfig::env` doc, which used to describe `"direct"` vs `"indirect"` interpretation — now stale since consumers can no longer flip the mode.
- Wire-shape test `env_value_mode_hardcoded_direct_on_create_and_resume` in `rust/tests/session_test.rs` covering both `session.create` and `session.resume` (including the resume-side `session.skills.reload` follow-up).

## Validation

```
cargo +1.94.0 test --all-features
cargo +nightly-2026-04-14 fmt --all -- --config-path .rustfmt.nightly.toml --check
cargo clippy --all-features --all-targets -- -D warnings
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features
```

All green.

## Release impact

Doesn't touch `rust/Cargo.toml`, `rust/Cargo.lock`, or `rust/release-plz.toml`, so `rust-publish-release.yml` won't fire on merge. release-plz will pick this up as a breaking change on the next manual `rust-release-pr.yml` run, alongside the other 0.1.0 follow-ups in #1212.

### DIFF
--- a/dotnet/src/Generated/SessionEvents.cs
+++ b/dotnet/src/Generated/SessionEvents.cs
@@ -74,6 +74,8 @@ namespace GitHub.Copilot.SDK;
 [JsonDerivedType(typeof(SessionPlanChangedEvent), "session.plan_changed")]
 [JsonDerivedType(typeof(SessionRemoteSteerableChangedEvent), "session.remote_steerable_changed")]
 [JsonDerivedType(typeof(SessionResumeEvent), "session.resume")]
+[JsonDerivedType(typeof(SessionScheduleCancelledEvent), "session.schedule_cancelled")]
+[JsonDerivedType(typeof(SessionScheduleCreatedEvent), "session.schedule_created")]
 [JsonDerivedType(typeof(SessionShutdownEvent), "session.shutdown")]
 [JsonDerivedType(typeof(SessionSkillsLoadedEvent), "session.skills_loaded")]
 [JsonDerivedType(typeof(SessionSnapshotRewindEvent), "session.snapshot_rewind")]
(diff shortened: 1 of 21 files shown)


stephentoub:
https://github.com/github/copilot-sdk/pull/1212 already covers most of this and a bit more. Could we merge that and just trim this back to env_value_mode, which I neglected in that other PR?

tclem:
> #1212 already covers most of this and a bit more. Could we merge that and just trim this back to env_value_mode, which I neglected in that other PR?

Yes, that's fine. Go ahead and merge that and this one can sync up if there is anything left to do.

stephentoub:
> and this one can sync up if there is anything left to do.

I _think_ it'll just be the env_value_mode, which I missed in the other one. Thanks for following up on that.

tclem:
OK @stephentoub - down to just `env_value_mode` changes now. I'll let you approve/merge.

stephentoub (approved):

PR #1222, opened 2026-05-07
feat(rust): support binary tool results

cschleiden:
Rust tool results need to preserve binary payloads from MCP tool calls so callers can return images and binary resources to the LLM without collapsing structured results into text.

This updates the Rust SDK coverage around binary tool results by ensuring resource blobs default to `application/octet-stream` when `mimeType` is missing or empty, and by adding focused tests for the public `binaryResultsForLlm` wire shape plus MCP image, resource blob, mixed text/binary, and no-binary conversion behavior.

Validation:
- `cd rust && cargo test --all-features`


stephentoub:
Thanks, @cschleiden. I _think_ the Python and Node.js SDKs have the same bug. Would you mind fixing those in this PR as well?

PR #1234, opened 2026-05-08
Replace Go RPC quicktype generation

qmuntal:
## Motivation

Go RPC type generation currently depends on quicktype plus several post-processing passes. That breaks down when distinct schema definitions have the same structural shape: quicktype can merge those definitions into one Go type, which caused separate protocol concepts such as `MCPServerSource` and `DiscoveredMCPServerSource` to be deduplicated. Quicktype does not expose a suitable Go option to disable that same-layout merging, so the generated API cannot reliably preserve schema-level type identity.

The quicktype output also made casing harder to keep consistent with this repository's Go naming policy. Owning the Go generator lets us apply one explicit initialism list and avoid a mix of incidental quicktype casing plus local post-processing fixes.

This change follows the custom-generator approach used by other SDK codegen and replaces quicktype for Go RPC types with a schema-aware Go generator. The generator preserves named definitions, keeps generated declarations sorted, emits quicktype-like wrapped comments for RPC output, preserves enum ordering, and flattens object unions so the generated Go surface stays close to the existing quicktype shape where that behavior was desirable.

Dropping quicktype also makes the Go codegen easier to reason about: the generator now owns the schema-to-Go mapping directly instead of generating quicktype output and then repairing it with a pile of post-processing hacks. The end result is better-controlled codegen with an equivalent number of lines, but with the behavior expressed in the generator instead of scattered after-the-fact rewrites.

Python codegen still uses quicktype; this only removes quicktype from the Go RPC generation path.

## Breaking changes

This changes the generated Go API surface. The breaking changes are intentionally limited to generated Go types, constants, fields, and method signatures that reference those generated types.

The main breaking-change patterns are:

- Generated names now follow the Go generator's current initialism list exactly. Since `mcp`, `fs`, and `sse` are intentionally not initialisms in this PR, generated `MCP...` / `SessionFS...` names become `Mcp...` / `SessionFs...`. The same casing rule also applies to generated field names such as `PID` and `MS` suffixes.
- The synthetic `RPCTypes` root type is removed. It existed only as a quicktype input artifact and was not a real RPC schema concept.
- Distinct schema definitions that happen to have identical structure now remain distinct Go types instead of being collapsed into shared quicktype helper types. As a result, some previously shared helper types become schema-specific names, especially around models/plugins, workspace metadata, task status/execution mode, permission approval kinds, and filter mapping values.
- Discriminated object unions are flattened into Go structs, matching the useful quicktype shape while preserving schema identity. This can produce schema-specific discriminator fields/types instead of reusing broad merged enum names.
- Non-discriminated object unions are also flattened where quicktype previously produced a useful value struct shape, such as MCP server config and UI elicitation schema property unions.
- Fields that quicktype previously emitted as `any` or broad primitive types are now generated as typed structs, typed maps, or schema-specific enum/string aliases when the schema provides that information. This affects areas such as embedded resources, UI elicitation values, and session-event payload details.

Call sites and tests in the Go SDK were updated for the renamed generated RPC types, renamed fields, schema-specific helper types, and typed generated fields.

## Why the breaks are done

The generated RPC package is derived from the protocol schema, so preserving the schema's named concepts is more important than preserving quicktype's incidental same-layout merging. Keeping separate named types avoids accidental aliasing between unrelated protocol concepts, makes future schema evolution safer, and removes fragile post-generation rewrites that were compensating for quicktype behavior.

The casing changes are done so generated names consistently follow the generator's explicit initialism list. The `Mcp` and `SessionFs` casing is intentional for this PR: it keeps the current initialism policy unchanged, so adding `mcp`, `fs`, or `sse` can be reviewed separately if desired.

## Validation

- `npm run generate:go`
- `go test ./... -run '^$'`


go/rpc/generated_rpc.go
```diff
@@ -12,257 +12,6 @@ import (
 	"time"
 )
 
-type RPCTypes struct {
```
SteveSandersonMS, line 15: It's great to see this gone. It made every update produce an enormous, unreadable diff.

go/rpc/generated_rpc.go
```diff
@@ -464,79 +260,33 @@ type ExternalToolTextResultForLlm struct {
...
-// Terminal/shell output content block with optional exit code and working directory
-//
-// # Image content block with base64-encoded data
-//
-// # Audio content block with base64-encoded data
-//
-// # Resource link content block referencing an external resource
-//
-// Embedded resource content block with inline text or binary data
 type ExternalToolTextResultForLlmContent struct {
```
SteveSandersonMS, line 263: Am I misreading or is this collapsing the discriminated union into a single struct?

If so that seems potentially dangerous, because what if more than one of the `ExternalToolTextResultForLlmContent` subtypes declared fields with clashing names but different types? We did have this problem in practice before and had to change how we codegenned the discriminated unions to avoid collapsing multiple types into a single struct.

go/rpc/generated_rpc.go
```diff
@@ -1125,43 +941,44 @@ type PermissionDecisionApproveForSession struct {
 
 // The approval to add as a session-scoped rule
 type PermissionDecisionApproveForSessionApproval struct {
-	CommandIdentifiers []string     `json:"commandIdentifiers,omitempty"`
-	Kind               ApprovalKind `json:"kind"`
-	ServerName         *string      `json:"serverName,omitempty"`
-	ToolName           *string      `json:"toolName,omitempty"`
+	// Kind discriminator
+	Kind               PermissionDecisionApproveForSessionApprovalKind `json:"kind"`
+	CommandIdentifiers []string                                        `json:"commandIdentifiers,omitempty"`
```
SteveSandersonMS: I notice the fields appear out of alphabetical order here - maybe it's just that `Kind` is special cased, but if not, would it be possible to ensure that all members (types, fields, etc) appear alphabetically relative to their peers? 

Just looking for a deterministic ordering that's easy to agree on so we can minimize diff sizes and retain the ordering no matter what codegen logic changes we do in the future.

SteveSandersonMS:
Thanks @qmuntal! This looks like a great improvements. Just had a couple of comments but I'm sure we can get this merged soon.

stephentoub (approved):

go/rpc/generated_rpc.go
qmuntal: Good call. I've made updated the generator to emit alphabetical order symbols wherever possible.

go/rpc/generated_rpc.go
qmuntal: Ye, that's what's happening here. But this was already collapsed before this PR. I'll fix it in a follow-up.

qmuntal:
@stephentoub this can be merged now, thanks!

stomde (commented):
Gd

PR #1309, opened 2026-05-15
feat: add model field to CustomAgentConfig across all SDKs

patniko:
## Summary

Add an optional `model` property to `CustomAgentConfig` in all five SDKs (Node/TS, Python, Go, .NET, Rust). This closes the gap between the runtime's `SweCustomAgent.model` support and the SDK's public API.

## Motivation

The runtime already supports per-subagent model selection via YAML frontmatter in `.github/agents/*.md` files, but the programmatic SDK API (`customAgents` array in session config) did not expose this capability. Users who define agents programmatically could not specify which model an agent should use.

## Changes

Each SDK gets its own commit for easy review:

- **Node/TS** (`nodejs/src/types.ts`): `model?: string` on `CustomAgentConfig` interface — passed through directly to wire format
- **Python** (`python/copilot/session.py` + `client.py`): `model: NotRequired[str]` on `CustomAgentConfig` TypedDict + wire format conversion
- **Go** (`go/types.go`): `Model string` field with `json:"model,omitempty"` tag
- **.NET** (`dotnet/src/Types.cs`): `string? Model` property with `[JsonPropertyName("model")]`
- **Rust** (`rust/src/types.rs`): `model: Option<String>` field + `with_model()` builder method

## Behavior

When set, the runtime will attempt to use the specified model for the agent, falling back to the parent session model if unavailable. When omitted/null, the agent inherits the session's model (existing behavior, no breaking change).

## Validation

All five SDKs compile successfully: `npx tsc --noEmit`, `python3 -c "import copilot"`, `go build ./...`, `dotnet build`, `cargo check --all-features`.

### DIFF
--- a/dotnet/src/Types.cs
+++ b/dotnet/src/Types.cs
@@ -1933,6 +1933,14 @@ public class CustomAgentConfig
     /// </summary>
     [JsonPropertyName("skills")]
     public IList<string>? Skills { get; set; }
+
+    /// <summary>
+    /// Model identifier for this agent (e.g. "claude-haiku-4.5").
+    /// When set, the runtime will attempt to use this model for the agent,
+    /// falling back to the parent session model if unavailable.
+    /// </summary>
+    [JsonPropertyName("model")]
+    public string? Model { get; set; }
 }
 
 /// <summary>
--- a/dotnet/test/Unit/CloneTests.cs
+++ b/dotnet/test/Unit/CloneTests.cs
@@ -95,7 +95,7 @@ public class CloneTests
             EnableSessionTelemetry = false,
             IncludeSubAgentStreamingEvents = false,
             McpServers = new Dictionary<string, McpServerConfig> { ["server1"] = new McpStdioServerConfig { Command = "echo" } },
-            CustomAgents = [new CustomAgentConfig { Name = "agent1" }],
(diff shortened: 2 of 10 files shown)


dotnet/src/Types.cs
```diff
@@ -1933,6 +1933,14 @@ public class CustomAgentConfig
     /// </summary>
     [JsonPropertyName("skills")]
     public IList<string>? Skills { get; set; }
+
+    /// <summary>
+    /// Model identifier for this agent (e.g. "claude-haiku-4.5").
+    /// When set, the runtime will attempt to use this model for the agent,
+    /// falling back to the parent session model if unavailable.
```
stephentoub, line 1940: > falling back to the parent session model if unavailable.

I'm torn on this. So if I have a typo in the model name, I don't get a warning or failure or anything, it just silently reverts to a different model?

Not saying we shouldn't do that... just thinking it through.

stephentoub (approved):

dotnet/src/Types.cs
patniko: I am not a big fan of how we handle these as well, but it is the norm for custom agents across companies. Side effect of most only supporting their single family of models and the hyper competitive market. If this wasn't the standard name I would have preferred `default_model`, `preferred_model` or even supported a clear set so we could match at least 1.

PR #1313, opened 2026-05-16
Fix Python from_dict() round-trip for optional fields with schema defaults

### LINKED ISSUE #1139: SessionTaskCompleteData round-trip turns None summary into empty string
## Summary

`SessionTaskCompleteData.from_dict()` reads a missing/null `summary` field as the empty string `""` instead of `None`, so the round-trip `from_dict(to_dict(x))` is not the identity for `SessionTaskCompleteData(success=None, summary=None)`. Any code that compares the parsed event to the original, or that distinguishes "no summary provided" from "empty summary", sees incorrect data.

## Affected versions

`main` at commit `dd2dcbc439256acfb9feb2cff07c0b9c820091b8`. The class is generated; the same shape ships in every Python release built from this codegen.

## Affected source

- `python/copilot/generated/session_events.py:2807-2820` — `SessionTaskCompleteData` definition and `from_dict`
- The bug is on line **2816**:
  ```python
  summary = from_union([from_none, from_str], obj.get("summary", ""))
  ```
  Compare line 2815 (correct):
  ```python
  success = from_union([from_none, from_bool], obj.get("success"))
  ```

The `obj.get("summary", "")` default coerces a missing key to `""`, which `from_union([from_none, from_str], ...)` then accepts as a valid `str`. Without the default, `obj.get("summary")` returns `None`, and `from_union` correctly resolves through `from_none`.

## Reproduction

```python
from copilot.generated.session_events import SessionTaskCompleteData

original = SessionTaskCompleteData(success=None, summary=None)
print("original:  ", original)
print("to_dict(): ", original.to_dict())
roundtrip = SessionTaskCompleteData.from_dict(original.to_dict())
print("roundtrip: ", roundtrip)
print("equal?     ", roundtrip == original)
```

## Expected vs actual output

Expected:
```
original:   SessionTaskCompleteData(success=None, summary=None)
to_dict():  {}
roundtrip:  SessionTaskCompleteData(success=None, summary=None)
equal?      True
```

Actual:
```
original:   SessionTaskCompleteData(success=None, summary=None)
to_dict():  {}
roundtrip:  SessionTaskCompleteData(success=None, summary='')
equal?      False
```

## Root cause

The `obj.get("summary", "")` default on line 2816 is inconsistent with both:

1. **The dataclass field declaration** (line 2810): `summary: str | None = None` — the in-memory default is `None`, so the wire-default should also be `None`.
2. **The sibling field's parser** (line 2815): `success` correctly uses `obj.get("success")` with no default and round-trips cleanly.
3. **The serializer** (lines 2822-2826, omits `summary` when `None`): the asymmetry between `to_dict()` (omit on None) and `from_dict()` (default to "") is what breaks the round-trip.

This is the only occurrence of the `obj.get("...", "")` pattern in `session_events.py` (`grep -c 'obj.get("[a-zA-Z]*", "")'` returns 1), so the fix is local to one line.

## Impact

- The dataclass advertises `summary: str | None = None`, so callers may pattern-match on `None` to mean "no summary." After a round-trip they see `""` instead, breaking that branch.
- Any code caching or comparing events (e.g., session replay, audit logs, deduplication) sees `original != round_tripped` for events with no summary.
- `SessionTaskCompleteData` is dispatched at `session_events.py:4480` from the `SESSION_TASK_COMPLETE` event type — so any session that emits a task-complete event without a summary triggers the bug end-to-end.

### LINKED ISSUE #1140: PermissionPromptRequest.action round-trip turns None into PermissionPromptRequestMemoryAction.STORE
## Summary

`PermissionPromptRequest.from_dict()` reads a missing `action` field as the literal string `"store"`, which then gets parsed as the enum value `PermissionPromptRequestMemoryAction.STORE` instead of `None`. Round-tripping `from_dict(to_dict(x))` for any `PermissionPromptRequest` with `action=None` silently turns it into a `STORE` action.

This is the same shape as already-fixed bug #1139 (`SessionTaskCompleteData.summary`), in a different class.

## Affected versions

`main` at commit `dd2dcbc439256acfb9feb2cff07c0b9c820091b8`. Generated file; same logic ships in every Python release built from this codegen.

## Affected source

- `python/copilot/generated/session_events.py:1695` — the misdefault on `action`:
  ```python
  action = from_union([from_none, lambda x: parse_enum(PermissionPromptRequestMemoryAction, x)], obj.get("action", "store"))
  ```

Compare the immediately-adjacent sibling field on **line 1694**, which is correct (no default):
```python
access_kind = from_union([from_none, lambda x: parse_enum(PermissionPromptRequestPathAccessKind, x)], obj.get("accessKind"))
```

The dataclass declaration (line 1664) has the right in-memory default:
```python
action: PermissionPromptRequestMemoryAction | None = None
```

And `to_dict()` correctly omits `action` when it is `None` (line 1755):
```python
if self.action is not None:
    result["action"] = ...
```

The asymmetry is in `from_dict` only.

## Reproduction

```python
from copilot.generated.session_events import PermissionPromptRequest, PermissionPromptRequestKind

original = PermissionPromptRequest(kind=PermissionPromptRequestKind.MEMORY)
print("original.action:  ", original.action)
print("to_dict() has 'action' key?", "action" in original.to_dict())
roundtrip = PermissionPromptRequest.from_dict(original.to_dict())
print("roundtrip.action: ", roundtrip.action)
print("equal?            ", roundtrip == original)
```

## Expected vs actual output

Expected:
```
original.action:   None
to_dict() has 'action' key? False
roundtrip.action:  None
equal?             True
```

Actual:
```
original.action:   None
to_dict() has 'action' key? False
roundtrip.action:  PermissionPromptRequestMemoryAction.STORE
equal?             False
```

## Root cause

The `obj.get("action", "store")` default coerces a missing key to the string `"store"`. `from_union([from_none, parse_enum], "store")` then succeeds via `parse_enum`, producing `PermissionPromptRequestMemoryAction.STORE` — the enum value whose wire representation is `"store"`.

The sibling field on line 1694 (`access_kind`) does it correctly with `obj.get("accessKind")` and no default, so missing keys flow through `from_none` and produce `None`.

This is the same misdefault pattern as already-fixed bug #1139 (`SessionTaskCompleteData.summary`), where `obj.get("summary", "")` was changed to `obj.get("summary")` in line 2816.

## Impact

- `PermissionPromptRequest` is dispatched by the permission-handling flow in `python/copilot/session.py`. Any permission request without an `action` field becomes a `STORE` action after parsing.
- Code that pattern-matches on `action == None` to mean "no memory action requested" sees `STORE` instead and may take an unintended branch.
- Round-tripping (caching, replay, audit logs) silently mutates the field.

### LINKED ISSUE #1141: PermissionRequest.action round-trip turns None into PermissionRequestMemoryAction.STORE
## Summary

`PermissionRequest.from_dict()` reads a missing `action` field as the literal string `"store"`, which then gets parsed as the enum value `PermissionRequestMemoryAction.STORE` instead of `None`. Round-tripping `from_dict(to_dict(x))` for any `PermissionRequest` with `action=None` silently turns it into a `STORE` action.

This is the same shape as already-fixed bug #1139 (`SessionTaskCompleteData.summary`), in a different class.

## Affected versions

`main` at commit `dd2dcbc439256acfb9feb2cff07c0b9c820091b8`. Generated file; same logic ships in every Python release built from this codegen.

## Affected source

- `python/copilot/generated/session_events.py:1845` — the misdefault on `action`:
  ```python
  action = from_union([from_none, lambda x: parse_enum(PermissionRequestMemoryAction, x)], obj.get("action", "store"))
  ```

Compare the immediately-adjacent sibling field on **line 1846**, which is correct (no default):
```python
args = obj.get("args")
```

The dataclass declaration (line 1812) has the right in-memory default:
```python
action: PermissionRequestMemoryAction | None = None
```

`to_dict()` correctly omits `action` when it is `None`. The asymmetry is in `from_dict` only.

## Reproduction

```python
from copilot.generated.session_events import PermissionRequest, PermissionRequestKind

original = PermissionRequest(kind=PermissionRequestKind.MEMORY)
print("original.action:  ", original.action)
print("to_dict() has 'action' key?", "action" in original.to_dict())
roundtrip = PermissionRequest.from_dict(original.to_dict())
print("roundtrip.action: ", roundtrip.action)
print("equal?            ", roundtrip == original)
```

## Expected vs actual output

Expected:
```
original.action:   None
to_dict() has 'action' key? False
roundtrip.action:  None
equal?             True
```

Actual:
```
original.action:   None
to_dict() has 'action' key? False
roundtrip.action:  PermissionRequestMemoryAction.STORE
equal?             False
```

## Root cause

The `obj.get("action", "store")` default coerces a missing key to the string `"store"`. `from_union([from_none, parse_enum], "store")` then succeeds via `parse_enum`, producing `PermissionRequestMemoryAction.STORE` — the enum value whose wire representation is `"store"`.

The sibling field on line 1846 (`args`) does it correctly with `obj.get("args")` and no default, so missing keys flow through correctly and produce `None`.

This is the same misdefault pattern as already-fixed bug #1139 (`SessionTaskCompleteData.summary`), where `obj.get("summary", "")` was changed to `obj.get("summary")` in line 2816.

## Impact

- `PermissionRequest` is part of the permission-handling flow in `python/copilot/session.py`. Any permission request without an `action` field becomes a `STORE` action after parsing.
- Code that pattern-matches on `action == None` to mean "no memory action requested" sees `STORE` instead and may take an unintended branch.
- Round-tripping (caching, replay, audit logs) silently mutates the field.

stephentoub:
Fixes #1139.
Fixes #1140.
Fixes #1141.

## Why

`from_dict(to_dict(x))` was not the identity for three generated Python dataclasses whenever the affected optional field was `None`. The dataclass declares the field as `T | None = None` and `to_dict()` omits it when `None`, but `from_dict()` was substituting the JSON-Schema `default` value when the key was missing, silently mutating unset fields:

- `SessionTaskCompleteData.summary`: `None` → `""`
- `PermissionPromptRequest.action`: `None` → `PermissionPromptRequestMemoryAction.STORE`
- `PermissionRequest.action`: `None` → `PermissionRequestMemoryAction.STORE`

Any code that round-trips events (caching, replay, audit logs, dedup) or pattern-matches on `None` to mean "absent" was getting the wrong answer end-to-end. The bug was reproducible on `main` as of today.

## What changed

Root cause is in `scripts/codegen/python.ts`: for optional fields with a schema `default`, the codegen was emitting `obj.get("key", "<default>")`, then passing that into `from_union([from_none, parse_X], ...)`, which happily parsed the default and returned it. The dataclass default and the wire default were inconsistent, so the round-trip broke.

- Dropped the `defaultLiteral` mechanism from both `emitPyClass` and `emitPyFlatDiscriminatedUnion`. `from_dict()` now uniformly emits `obj.get(key)`, so missing/null keys flow through `from_none` and land at the dataclass-declared `None`. Removed the now-unused `toPythonLiteral` helper.
- Regenerated `python/copilot/generated/session_events.py` via `npm run generate:python`. Exactly three call sites changed, all matching the bug reports.
- Flipped the two codegen-level assertions in `nodejs/test/python-codegen.test.ts` that previously locked in the buggy output, and added negative assertions to prevent regression at the generator level.
- Replaced `test_schema_defaults_are_applied_for_missing_optional_fields` in `python/test_event_forward_compatibility.py` (which asserted the bug as expected behavior) with two regression tests covering missing-key parsing, explicit-null parsing, and the full `from_dict(to_dict(x))` round-trip for all three affected classes.

## Surface area / non-obvious notes

- Other languages (Go, .NET, Rust, TypeScript) are not affected. Their generators never read `propSchema.default` to construct deserialization fallbacks; they use real nullable types end-to-end.
- Grep on the post-fix generated `session_events.py` and `rpc.py` confirms zero remaining `obj.get("...", "...")` patterns, so no other class was silently relying on baked-in schema defaults.
- The `test_schema_defaults_are_applied_for_missing_optional_fields` test was the only consumer in the repo of the old behavior. Nothing in runtime code depended on `request.action == STORE` or `task_complete.summary == ""` for unset fields.
- JSON Schema `default` is an annotation, not validation behavior, so applying it only during deserialization while ignoring it in the dataclass field and in `to_dict()` was internally inconsistent. This change makes Python deserialization symmetric with serialization and with the in-memory default.

## Validation

- `python -m pytest test_event_forward_compatibility.py` -> 10/10 pass (including new regression tests).
- Broader Python suite: 136/136 non-E2E tests pass.
- `npx vitest run` on `python-codegen.test.ts` and other non-CLI test files: pass.
- `ruff check`, `ruff format --check`, `ty check copilot`, `tsc --noEmit`, `eslint`: clean.
- Manual reproduction of each issue's repro script on unmodified `main` (changes stashed) confirms the bug; with the fix applied, all three round-trip with `equal? True`.

Independent review by the `code-review` agent on `claude-opus-4.7-xhigh` found no significant issues and confirmed both codegen paths and the test coverage.

### DIFF
--- a/nodejs/test/python-codegen.test.ts
+++ b/nodejs/test/python-codegen.test.ts
@@ -74,11 +74,13 @@ describe("python session event codegen", () => {
         );
         expect(code).toContain("def to_timedelta_int(x: timedelta) -> int:");
         expect(code).toContain(
-            'action = from_union([from_none, lambda x: parse_enum(SessionSyntheticDataAction, x)], obj.get("action", "store"))'
+            'action = from_union([from_none, lambda x: parse_enum(SessionSyntheticDataAction, x)], obj.get("action"))'
         );
         expect(code).toContain(
-            'summary = from_union([from_none, from_str], obj.get("summary", ""))'
+            'summary = from_union([from_none, from_str], obj.get("summary"))'
         );
+        expect(code).not.toContain('obj.get("action", "store")');
+        expect(code).not.toContain('obj.get("summary", "")');
         expect(code).toContain("uri: str");
         expect(code).toContain("pattern: str");
         expect(code).toContain("payload: str");
--- a/nodejs/test/python-codegen.test.ts
+++ b/nodejs/test/python-codegen.test.ts
@@ -76,9 +76,7 @@ describe("python session event codegen", () => {
         expect(code).toContain(
             'action = from_union([from_none, lambda x: parse_enum(SessionSyntheticDataAction, x)], obj.get("action"))'
         );
-        expect(code).toContain(
-            'summary = from_union([from_none, from_str], obj.get("summary"))'
-        );
+        expect(code).toContain('summary = from_union([from_none, from_str], obj.get("summary"))');
         expect(code).not.toContain('obj.get("action", "store")');
         expect(code).not.toContain('obj.get("summary", "")');
         expect(code).toContain("uri: str");
--- a/python/copilot/generated/session_events.py
+++ b/python/copilot/generated/session_events.py
@@ -5245,3 +5245,4 @@ def session_event_from_dict(s: Any) -> SessionEvent:
 
 def session_event_to_dict(x: SessionEvent) -> Any:
     return x.to_dict()
+
--- a/python/copilot/generated/session_events.py
+++ b/python/copilot/generated/session_events.py
@@ -1879,7 +1879,7 @@ class PermissionPromptRequest:
         assert isinstance(obj, dict)
         kind = parse_enum(PermissionPromptRequestKind, obj.get("kind"))
         access_kind = from_union([from_none, lambda x: parse_enum(PermissionPromptRequestPathAccessKind, x)], obj.get("accessKind"))
-        action = from_union([from_none, lambda x: parse_enum(PermissionRequestMemoryAction, x)], obj.get("action", "store"))
+        action = from_union([from_none, lambda x: parse_enum(PermissionRequestMemoryAction, x)], obj.get("action"))
         args = from_union([from_none, lambda x: x], obj.get("args"))
         can_offer_session_approval = from_union([from_none, from_bool], obj.get("canOfferSessionApproval"))
         capabilities = from_union([from_none, lambda x: from_list(from_str, x)], obj.get("capabilities"))
@@ -2044,7 +2044,7 @@ class PermissionRequest:
     def from_dict(obj: Any) -> "PermissionRequest":
         assert isinstance(obj, dict)
         kind = parse_enum(PermissionRequestKind, obj.get("kind"))
-        action = from_union([from_none, lambda x: parse_enum(PermissionRequestMemoryAction, x)], obj.get("action", "store"))
+        action = from_union([from_none, lambda x: parse_enum(PermissionRequestMemoryAction, x)], obj.get("action"))
         args = obj.get("args")
         can_offer_session_approval = from_union([from_none, from_bool], obj.get("canOfferSessionApproval"))
         capabilities = from_union([from_none, lambda x: from_list(from_str, x)], obj.get("capabilities"))
@@ -3280,7 +3280,7 @@ class SessionTaskCompleteData:
     def from_dict(obj: Any) -> "SessionTaskCompleteData":
         assert isinstance(obj, dict)
         success = from_union([from_none, from_bool], obj.get("success"))
-        summary = from_union([from_none, from_str], obj.get("summary", ""))
+        summary = from_union([from_none, from_str], obj.get("summary"))
         return SessionTaskCompleteData(
             success=success,
             summary=summary,
@@ -5245,4 +5245,3 @@ def session_event_from_dict(s: Any) -> SessionEvent:
 
 def session_event_to_dict(x: SessionEvent) -> Any:
     return x.to_dict()
-
--- a/python/test_event_forward_compatibility.py
+++ b/python/test_event_forward_compatibility.py
@@ -17,7 +17,10 @@ from copilot.generated.session_events import (
     ElicitationCompletedAction,
     ElicitationRequestedMode,
     ElicitationRequestedSchema,
+    PermissionPromptRequest,
+    PermissionPromptRequestKind,
     PermissionRequest,
+    PermissionRequestKind,
     PermissionRequestMemoryAction,
     SessionEventType,
     SessionTaskCompleteData,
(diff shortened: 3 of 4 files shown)


stephentoub:
@brettcannon would you mind looking at this? Is it the right fix?

brettcannon (approved):

brettcannon:
> Is it the right fix?

Yep, it's the right fix for the problem.

PR #1323, opened 2026-05-18
Expose enableOnDemandInstructionDiscovery across all SDK SessionConfig types

examon:
Exposes the new `enableOnDemandInstructionDiscovery` configuration flag across all five in-repo SDKs (Node, Python, Go, .NET, Rust). Mirrors the precedent set by [#1044 (`enableConfigDiscovery`)](https://github.com/github/copilot-sdk/pull/1044) and [#1295 (`remoteSession`)](https://github.com/github/copilot-sdk/pull/1295).

The runtime change shipped in [github/copilot#7759](https://github.com/github/copilot-agent-runtime/pull/7759) and is available in `@github/copilot@^1.0.49-1`.

## Cross-language consistency matrix

| SDK    | Public field                                                       | Wire key                                | Create | Resume | Unset behavior | Explicit-`false` behavior |
| ------ | ------------------------------------------------------------------ | --------------------------------------- | ------ | ------ | -------------- | ------------------------- |
| Node   | `enableOnDemandInstructionDiscovery?: boolean`                     | `enableOnDemandInstructionDiscovery`    | ✓      | ✓      | omitted        | wire `false`              |
| Python | `enable_on_demand_instruction_discovery: bool \| None = None`      | `enableOnDemandInstructionDiscovery`    | ✓      | ✓      | omitted        | wire `false`              |
| Go     | `EnableOnDemandInstructionDiscovery *bool`                         | `enableOnDemandInstructionDiscovery`    | ✓      | ✓      | omitted (`nil`)| wire `false`              |
| .NET   | `bool? EnableOnDemandInstructionDiscovery`                         | `enableOnDemandInstructionDiscovery`    | ✓      | ✓      | omitted (null) | wire `false`              |
| Rust   | `enable_on_demand_instruction_discovery: Option<bool>`             | `enableOnDemandInstructionDiscovery`    | ✓      | ✓      | omitted (None) | wire `false`              |

## Runtime-gated availability

When set to `true`, the SDK asks the runtime to discover custom instruction files on demand after the agent successfully reads or views files. The option is runtime-gated: it takes effect only when custom instructions are enabled and the connected runtime supports and enables on-demand custom instruction discovery. Otherwise the runtime accepts the option but performs no on-demand discovery (silent no-op, no error or warning).

The SDK does not attempt to detect runtime gating; it forwards the option unconditionally. On `session.resume`, omitting the option leaves the existing session setting unchanged, and passing an explicit `false` disables future on-demand discovery for that resumed session.

## Trust and security

Enable only for trusted repositories or workspaces. Discovered instruction files are treated as model instructions, may affect agent behavior, and may be stored or replayed with session history. Do not enable for untrusted content, CI jobs processing untrusted forks, or directories writable by untrusted users or processes.

## Go shape

Uses `*bool` (not `bool`) on both `SessionConfig` and `ResumeSessionConfig` so callers can disable a previously-enabled session on resume. Reuses the precedent already set by `EnableSessionTelemetry *bool` and `IncludeSubAgentStreamingEvents *bool`. This PR does not retrofit the existing `EnableConfigDiscovery bool` field — that would be a separate breaking source change with broader blast radius.

## Tests

Each SDK adds tests for the new field on both `session.create` and `session.resume`, asserting that explicit `false` is serialized on the wire and that omission drops the key entirely. Mirrors the test patterns already in place for `enable_session_telemetry`, `include_sub_agent_streaming_events`, and `enable_config_discovery`.

- Node: extends `nodejs/test/e2e/client_options.e2e.test.ts` happy-path; adds 2 unit tests in `nodejs/test/client.test.ts` (create + resume, explicit `false`).
- Python: extends `python/e2e/test_client_options_e2e.py` happy-path; adds 2 unit tests in `python/test_client.py` (create + resume, explicit `False`).
- Go: extends `go/internal/e2e/client_options_e2e_test.go` happy-path; adds 6 wire-format unit tests in `go/client_test.go` (create + resume; `Bool(true)`, `Bool(false)`, and omitted).
- .NET: adds 2 serialization tests in `dotnet/test/Unit/SerializationTests.cs` (create + resume; true, false, omitted) and 4 clone tests in `dotnet/test/Unit/CloneTests.cs` covering the copy-ctor.
- Rust: adds 2 wire-format tests in `rust/tests/session_test.rs` (create + resume; `Some(true)`, `Some(false)`, omitted) and extends the existing builder unit-tests in `rust/src/types.rs` to call the new `.with_enable_on_demand_instruction_discovery(...)` method on both `SessionConfig` and `ResumeSessionConfig`.

## Documentation

The new field is documented inline in each SDK's typed config surface using the language-appropriate doc-comment style (JSDoc, Google-style, godoc, XML, Rustdoc). README and CHANGELOG are intentionally not updated, matching the precedent of #1044 and #1295.

## Notes for reviewers

- `go/types.go` shows ~170 line changes; the bulk is gofmt-driven struct-tag column re-alignment because the new field name is longer than any existing field. The only semantic changes are the additions of `EnableOnDemandInstructionDiscovery` to `SessionConfig`, `ResumeSessionConfig`, and the two wire-request structs.
- The existing `enableConfigDiscovery` docstring statement that custom instruction files "are always loaded from the working directory regardless of this setting" remains accurate — the new option adds on-demand discovery on top of the up-front load; it does not replace it.
- `python/copilot/session.py` `TypedDict`s do not include the new field. They are also missing existing fields (`enable_config_discovery`, `remote_session`); closing those gaps is best handled in a follow-up so this PR stays narrow.


dotnet/src/Types.cs
```diff
@@ -2138,6 +2139,25 @@ protected SessionConfig(SessionConfig? other)
     /// </summary>
     public bool? EnableConfigDiscovery { get; set; }
 
+    /// <summary>
+    /// When <see langword="true"/>, requests on-demand discovery of custom instruction
```
stephentoub: What does "on-demand discovery" actually mean? Demanded by what / when, discovered where, etc.

examon: @stephentoub thanks for feedback. I improved the docs so they explain how the on-demand discovery works and how the customer instructions are injected into the context https://github.com/github/copilot-sdk/pull/1323/changes/8586e6a6a7336238d4e358958be3765e88de7fbe

examon:
Resolved the merge conflicts by merging latest `main`.

Heads up for reviewers: the `enableOnDemandInstructionDiscovery` implementation already landed independently in #1474 ("granular per-session flags for multitenancy hardening") across all five SDKs (public field + wire serialization on create/resume). I resolved every conflict by taking `main`'s implementation, so all production source files are now byte-identical to `main`.

As a result, the remaining diff of this branch against `main` is test-only regression coverage:
- Node/Python/Go/.NET: added create/resume serialization + clone tests for the flag (kept alongside the tests `main` added at the same spots).
- Rust: `main` made `SessionConfig`/`ResumeSessionConfig` non-`Serialize` and moved wire conversion to the crate-private `into_wire`, so the original integration tests that called `serde_json::to_value(&config)` no longer compile. I relocated that wire-format coverage into the in-crate `*_into_wire_serializes_bucket_b_fields` unit tests and kept the builder-method assertions.

Verified locally: Rust (lib + session integration), Node (field tests + full typecheck), Python (field tests), and .NET (Clone/Serialization suites) all pass; `go vet` and the Go unit tests pass.

Since the implementation is now superseded by #1474, this PR is effectively test-only. Maintainers may want to either merge it for the added regression coverage or close it as superseded.

stephentoub (approved):
Thanks!

PR #1324, opened 2026-05-18
Add filtering RPC call to all SDK clients

Chuxel:
Adds a new filtering RPC method across all SDK language bindings, enabling callers to dynamically register values they want to be filtered.


SteveSandersonMS:
Thanks for submitting this, @Chuxel.

It's possible there's some reason it has to be done this way, but if instead the APIs could be added to the `sharedApi` Zod schemas on the runtime side, most of (or maybe all) the per-language SDK code will appear automatically through codegen.

Chuxel:
@SteveSandersonMS - Will take a look. Thanks for the tip!

Chuxel:
We sync'd offline and will follow the automated path here, so closing this PR.

PR #1335, opened 2026-05-19
feat: add MCP Apps (SEP-1865) support

mattdholloway:
Adds opt-in MCP Apps (SEP-1865) support across all SDKs: a new `enableMcpApps` session flag and regenerated RPC/session-event types.

## Changes

- **`enableMcpApps` opt-in** (`SessionConfig` + `ResumeSessionConfig` in all SDKs) — plumbed to wire field `requestMcpApps`. Defaults to `false`; hosts without an iframe renderer are unaffected.
- **Capability detection** — consumers inspect `capabilities.ui.mcpApps` on the create/resume response to detect whether the runtime honored the opt-in

Closes https://github.com/github/copilot-mcp-core/issues/1715


mattdholloway:
@copilot+claude-opus-4.7 resolve the merge conflicts in this pull request

nodejs/src/types.ts
```diff
@@ -1406,6 +1406,24 @@ export interface SessionConfig {
...
+     * setHostContext,getHostContext}` JSON-RPC methods.
+     *
+     * SDK consumers MUST set this to `true` only when they have an iframe
+     * renderer that can display `ui://` MCP App bundles. Setting it without a
+     * renderer will cause MCP servers to register UI-enabled tool variants
+     * the consumer cannot display.
+     *
+     * @default false
+     */
+    enableMcpApps?: boolean;
```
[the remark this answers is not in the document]
mattdholloway: @copilot+claude-opus-4.7 Complete the implementation of the SDK surfaces across all implementations

mattdholloway:
@copilot+claude-opus-4.7 resolve the merge conflicts in this pull request

mattdholloway:
@copilot+claude-opus-4.7 resolve the merge conflicts in this pull request

mattdholloway:
@copilot+claude-opus-4.7 resolve the merge conflicts in this pull request

nodejs/src/client.ts
```diff
@@ -74,6 +74,31 @@ import { defaultJoinSessionPermissionHandler } from "./types.js";
...
+ * Apps would see no error — just tools that never expose
+ * `_meta.ui.resourceUri`.
+ *
+ * Using `process.emitWarning` (rather than `console.warn`) means consumers
+ * can route the warning via `process.on("warning", …)`, suppress it with
+ * `--no-warnings`, or filter by the `name` field below.
+ */
+function warnIfMcpAppsDropped(
+    requested: boolean | undefined,
+    capabilities: { ui?: { mcpApps?: boolean } } | undefined
```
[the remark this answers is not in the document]
mattdholloway: @copilot resolve this

mattdholloway:
@copilot+claude-opus-4.7 resolve the merge conflicts in this pull request

nodejs/src/mcpAppsSandbox.ts
```diff
@@ -0,0 +1,193 @@
+/**
```
SteveSandersonMS: This file seems to have no equivalent in the other languages, plus it's not used directly by any of the other SDK code. So my guess is this logic could actually go into the app that's using the MCP apps feature and then we wouldn't need to document or maintain it in the SDK.

I'm not sure if this code is even usable from the SDK since quite a bit of it is browser-specific. One of the `@example` blocks below refers to `iframe` which isn't a concept in Node.

Could we simply remove this file, and let app developers supply their own implementation for what it does?

dotnet/src/Client.cs
```diff
@@ -1991,6 +1995,23 @@ private void RemoveSession(string sessionId)
...
+    /// <c>_meta.ui.resourceUri</c>.
+    /// </summary>
+    private void WarnIfMcpAppsDropped(bool requested, SessionCapabilities? capabilities, string sessionId)
+    {
+        if (!requested) return;
+        if (capabilities?.Ui?.McpApps == true) return;
+        _logger?.LogWarning(
+            "Session {SessionId}: EnableMcpApps was requested but the runtime did not advertise capabilities.ui.mcpApps. The runtime's MCP_APPS feature flag or COPILOT_MCP_APPS=true environment override is likely unset; the MCP Apps surface is unavailable for this session.",
+            sessionId);
+    }
```
SteveSandersonMS: Instead of logging warnings, could we just throw an exception (or return an error in Go/Rust) if the runtime doesn't honor the request? If the app developer is trying to use MCP apps then they generally do so because they need it to work.

My guess is the developer effectively controls whether or not the runtime supports `capabilities.ui.mcpApps`, and so it's always correct for them to enable that if they want to use it.

mattdholloway: I think the right move might be to remove this warning altogether, it doesn't make much sense to have it as it seems to reductive to alert for the disablement of a feature that might have been intentionally disabled.

SteveSandersonMS (approved):
Thanks @mattdholloway! Will merge when checks pass.

PR #1343, opened 2026-05-20
C# API review fixes

SteveSandersonMS:
## Summary

API review fixes for the **.NET SDK only**. Sources combined into this PR:

- ApiView comments classified into "SDK-side fixes" (the runtime-side fixes landed via [copilot-agent-runtime#8283](https://github.com/github/copilot-agent-runtime/pull/8283)).
- The `api_review_csharp.md` review doc.
- Follow-on cleanups discovered while implementing the above.

Other SDKs are not touched here; once we are happy with the .NET shape we will port what makes sense.

## What changed (by phase)

| Phase | Theme |
|---|---|
| 3 | Sealed every hand-written public type that does not need a derived class. |
| 4a | Simple property/method renames (e.g. spelling, casing, plurals). |
| 4b | Expanded abbreviations in `SessionFsProvider`. |
| 4c | Shape changes: `Streaming` flag, tools list shape, `ToolBinaryResultType` cleanup, misc. |
| 4d/4e | Removed the redundant `CopilotClient.State` enum; retyped `LogLevel` to a strongly-typed `CopilotLogLevel` value type. |
| 4f | Generic `client.OnLifecycle<T>` / `session.On<T>`; lifecycle events are now a typed polymorphic hierarchy (`SessionCreatedEvent`, etc.) instead of stringly-typed discriminators. |
| 4g | `DateTimeOffset` timestamps; `PermissionRequestResult.Feedback` cleanup. |
| 5 | Extracted `SessionConfigBase` between `SessionConfig` / `ResumeSessionConfig`; sealed all config classes. |
| 6 | Removed every hand-written named delegate type in favour of plain `Func<...>`. |
| 7 | `SendAsync(string)` / `SendAndWaitAsync(string)` convenience overloads. |
| 8 | README updates to match. |
| 9 | New `RuntimeConnection` discriminated config: replaces the flat `CliPath`/`CliArgs`/`Port`/`UseStdio`/`CliUrl`/`TcpConnectionToken` properties on `CopilotClientOptions` with a single `Connection` of type `RuntimeConnection`, constructed via factories: `RuntimeConnection.Stdio(...)`, `RuntimeConnection.Tcp(...)`, `RuntimeConnection.Uri(...)`. Also renames `ActualPort` -> `RuntimePort` and `CopilotHome` -> `BaseDirectory`. |

The test harness also got a small UX improvement: `Ctx.CreateClient(useStdio: ..., options: ... { Connection = ... })` now throws eagerly if both are supplied. This caught a real bug introduced by Phase 9 bulk rewrite.

## Notable thing we explicitly did NOT do: Phase 1

Phase 1 was originally going to retype every "opaque JSON" RPC field from `object` to `System.Text.Json.JsonElement`, on the theory that `JsonElement` is more honest about wire shape than `object` (which cannot be safely cast). The change was reverted on this branch (see commits `129c281e` and `074ace71`) for two reasons:

1. **The underlying problem is in the runtime schema, not the SDK.** Several RPC parameters/results are polymorphic `anyOf` unions that have no idiomatic mapping in any of our SDK languages. Tracked in [copilot-agent-runtime#8375](https://github.com/github/copilot-agent-runtime/issues/8375): the schema generator should reject shapes that have no idiomatic mapping in C#/Go/Python/Rust, and require an explicit opt-in for cases that genuinely are meant to be opaque JSON. Once that lands, codegen will produce real typed hierarchies and the `JsonElement` workaround stops being needed.

2. **Phase 1 asymmetry created a worse consumer experience for inputs.** Changing input parameters from `object` to `JsonElement` forced consumers to serialize themselves via their own `JsonSerializerOptions`, which silently emitted `"headers": null` and similar (because `WhenWritingNull` is not a default) and caused the runtime to reject otherwise-valid configs. On the previous `object` path, the SDK own `JsonSerializerContext` (which sets `WhenWritingNull`) was used, so this never bit consumers. We could have applied per-property `[JsonIgnore(WhenWritingNull)]` workarounds, but they are band-aids; the right fix is the schema change above. Reverting now keeps the API symmetric and aligned with main.

We will revisit `JsonElement` once #8375 lands and codegen can emit real types.

## Risk / compatibility

This is a breaking-change PR by design. All E2E and unit tests have been updated in lock-step. The `dotnet/samples` and `docs/` C# snippets have also been migrated.

Two issues filed against the runtime as follow-ups to this work:

- [copilot-agent-runtime#8375](https://github.com/github/copilot-agent-runtime/issues/8375) - schema-shape lint (anyOf, opaque-JSON opt-in).
- The runtime-side ApiView fixes that did land - see [copilot-agent-runtime#8283](https://github.com/github/copilot-agent-runtime/pull/8283).


dotnet/src/Client.cs
```diff
@@ -1879,7 +1810,7 @@ public async ValueTask<SystemMessageTransformRpcResponse> OnSystemMessageTransfo
         public async ValueTask<ToolCallResponseV2> OnToolCallV2(string sessionId,
```
stephentoub, line 1803: Do we need to maintain the V2 support?

dotnet/src/Session.cs
```diff
@@ -984,9 +1000,9 @@ public async Task<bool> ConfirmAsync(string message, CancellationToken cancellat
             var schema = new UIElicitationSchema
             {
                 Type = "object",
-                Properties = new Dictionary<string, object>
+                Properties = new Dictionary<string, JsonElement>
                 {
-                    ["confirmed"] = new Dictionary<string, object> { ["type"] = "boolean", ["default"] = true }
+                    ["confirmed"] = JsonDocument.Parse("""{"type":"boolean","default":true}""").RootElement.Clone()
```
stephentoub: You can use JsonElement.Parse. Not only simpler, but JsonDocument.Parse returns a disposable thing that this should otherwise be disposing.

dotnet/src/Types.cs
```diff
@@ -50,23 +50,60 @@ internal static class Diagnostics
...
+/// static members (<see cref="None"/>, <see cref="Error"/>, <see cref="Warning"/>,
+/// <see cref="Info"/>, <see cref="Debug"/>, <see cref="All"/>), or construct
+/// your own with <see cref="CopilotLogLevel(string)"/> if the runtime accepts
+/// additional values. The runtime does not necessarily treat the values as a
+/// linear scale, so do not assume <c>&lt;</c>/<c>&gt;</c> comparisons are meaningful.
 /// </summary>
-[JsonConverter(typeof(JsonStringEnumConverter<ConnectionState>))]
-public enum ConnectionState
+[DebuggerDisplay("{Value,nq}")]
+public readonly struct CopilotLogLevel : IEquatable<CopilotLogLevel>
```
stephentoub, line 61: This is fine if you think it's useful to maintain a separate type, but we're already using ILogger{Factory}... should we just use LogLevel here? Internally we can translate it to whatever the runtime needs.

SteveSandersonMS: I started with LogLevel but started thinking nothing stops the runtime from coming up with new loglevel values at any time that could have any semantics, without being constrained to things that match a .NET LogLevel. My guess is this won't inconvenience .NET devs too much since they aren't programmatically mapping some other loglevel into this in most cases, so they will still just pick whatever level they think is right. We could add an implicit conversion from LogLevel later if we want.

dotnet/src/Session.cs
SteveSandersonMS: Good catch on the disposal issue. This is moot now: the Phase 1 codegen change that introduced `Dictionary<string, JsonElement>` (and this `JsonDocument.Parse(...).RootElement.Clone()` call) was reverted in 129c281e, so `ConfirmAsync` is back to a plain `Dictionary<string, object>` with no `JsonDocument` allocation. Resolving.

dotnet/README.md
```diff
@@ -749,7 +749,7 @@ var session = await client.CreateSessionAsync(new SessionConfig
 
 ### Custom Permission Handler
 
-Provide your own `PermissionRequestHandler` delegate to inspect each request and apply custom logic:
+Provide your own permission handler (`Func<PermissionRequest, ToolInvocation, Task<PermissionRequestResult>>`) to inspect each request and apply custom logic:
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in a50f6a14 - swapped the README signature to PermissionInvocation.

dotnet/src/Client.cs
```diff
@@ -211,15 +197,14 @@ private static Uri ParseCliUrl(string url)
     /// <remarks>
     /// <para>
     /// If the server is not already running and the client is configured to spawn one (default), it will be started.
-    /// If connecting to an external server (via CliUrl), only establishes the connection.
+    /// If connecting to an external runtime (via RuntimeConnection.Uri), only establishes the connection.
     /// </para>
     /// <para>
-    /// This method is called automatically when creating a session if <see cref="CopilotClientOptions.AutoStart"/> is true (default).
     /// </para>
     /// </remarks>
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in a50f6a14 - removed the empty <para/> block.

dotnet/src/Client.cs
```diff
@@ -231,24 +216,24 @@ public Task StartAsync(CancellationToken cancellationToken = default)
         async Task<Connection> StartCoreAsync(CancellationToken ct)
         {
             _logger.LogDebug("Starting Copilot client");
-            _disconnected = false;
+            // Reset disposed flag for restart scenarios
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in a50f6a14 - removed the stale comment; no reset is intended in StartAsync.

docs/setup/backend-services.md
```diff
@@ -215,15 +215,14 @@ response, _ := session.SendAndWait(ctx, copilot.MessageOptions{Prompt: message})
...
+using GitHub.Copilot;
 
 var userId = "user1";
 var message = "Hello";
 
 var client = new CopilotClient(new CopilotClientOptions
 {
-    CliUrl = "localhost:4321",
-    UseStdio = false,
+Connection = RuntimeConnection.Uri("localhost:4321"),
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in a50f6a14 - indented the Connection line in both backend-services.md snippets.

docs/setup/backend-services.md
```diff
@@ -240,8 +239,7 @@ var response = await session.SendAndWaitAsync(
 ```csharp
 var client = new CopilotClient(new CopilotClientOptions
 {
-    CliUrl = "localhost:4321",
-    UseStdio = false,
+Connection = RuntimeConnection.Uri("localhost:4321"),
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in a50f6a14 - same indentation fix.

docs/getting-started.md
```diff
@@ -2067,12 +2067,11 @@ let session = client
...
 
 ```csharp
-using GitHub.Copilot.SDK;
+using GitHub.Copilot;
 
 using var client = new CopilotClient(new CopilotClientOptions
 {
-    CliUrl = "localhost:4321",
-    UseStdio = false
+Connection = RuntimeConnection.Uri("localhost:4321"),
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in a50f6a14 - indented the Connection line.

docs/setup/backend-services.md
```diff
@@ -215,15 +215,14 @@ response, _ := session.SendAndWait(ctx, copilot.MessageOptions{Prompt: message})
...
+using GitHub.Copilot;
 
 var userId = "user1";
 var message = "Hello";
 
 var client = new CopilotClient(new CopilotClientOptions
 {
-    CliUrl = "localhost:4321",
-    UseStdio = false,
+    Connection = RuntimeConnection.Uri("localhost:4321"),
```
stephentoub: This is missing a verb or a preposition. How about RuntimeConnection.ForUri?

docs/setup/local-cli.md
```diff
@@ -139,7 +139,7 @@ if response != nil {
 ```csharp
 var client = new CopilotClient(new CopilotClientOptions
 {
-    CliPath = "/usr/local/bin/copilot",
+    Connection = RuntimeConnection.Stdio(path: "/usr/local/bin/copilot"),
```
stephentoub: Similarly here, ForStdio or CreateWithStdio or something like that?

docs/getting-started.md
```diff
@@ -109,7 +109,7 @@ dotnet new console -n CopilotDemo && cd CopilotDemo
 Then add the SDK:
 
 ```bash
-dotnet add package GitHub.Copilot.SDK
+dotnet add package GitHub.Copilot
```
stephentoub: Is the package name changing?

dotnet/src/Generated/Rpc.cs
```diff
@@ -6622,13 +6622,13 @@ public sealed class Converter : JsonConverter<ModelPickerCategory>
         /// <inheritdoc />
         public override ModelPickerCategory Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
-            return new(GitHub.Copilot.SDK.GeneratedStringEnumJson.ReadValue(ref reader, typeToConvert));
+            return new(GitHub.Copilot.GeneratedStringEnumJson.ReadValue(ref reader, typeToConvert));
```
stephentoub: We shouldn't need the full qualification here, e.g.

```suggestion
            return new(GeneratedStringEnumJson.ReadValue(ref reader, typeToConvert));
```

dotnet/src/GitHub.Copilot.SDK.csproj
```diff
@@ -2,6 +2,7 @@
 
     <PropertyGroup>
         <TargetFrameworks>net8.0;net10.0;netstandard2.0</TargetFrameworks>
+        <RootNamespace>GitHub.Copilot</RootNamespace>
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <Version>0.1.0</Version>
```
stephentoub: nit: we override this when publishing, but maybe we should tweak it to be clear that it's a local build, e.g. 0.0.0-dev or something like that

dotnet/src/Types.cs
```diff
@@ -1098,7 +1208,8 @@ public class PreToolUseHookInput
     /// Unix timestamp in milliseconds when the tool use was initiated.
     /// </summary>
     [JsonPropertyName("timestamp")]
-    public long Timestamp { get; set; }
+    [JsonConverter(typeof(UnixMillisecondsDateTimeOffsetConverter))]
+    public DateTimeOffset Timestamp { get; set; }
```
stephentoub, line 1212: We're inconsistent right now in whether the runtime sends down timestamps as milliseconds since epoch or as an iso datetime string. We should ideally consolidate on one or the other.

stephentoub (approved):

docs/getting-started.md
SteveSandersonMS: No, the package name stays as `GitHub.Copilot.SDK`. Only the C# namespace lost the `.SDK` suffix (with `<RootNamespace>GitHub.Copilot</RootNamespace>` set in the csproj to override the default that would otherwise be derived from the assembly name).

dotnet/src/Generated/Rpc.cs
SteveSandersonMS: Agreed - fixed in c1e5d5c7 (codegen + regenerated Rpc.cs and SessionEvents.cs). The full qualifier was leftover from the pre-rename namespace; the unqualified `GeneratedStringEnumJson` resolves via the parent namespace lookup.

PR #1344, opened 2026-05-20
Fix optional MCP stdio server args

### LINKED ISSUE #1231: MCPStdioServerConfig.args should be optional
The `args` field on `MCPStdioServerConfig` in [`[nodejs/src/types.ts](https://github.com/github/copilot-sdk/blob/main/nodejs/src/types.ts)`](https://github.com/github/copilot-sdk/blob/main/nodejs/src/types.ts) is currently required:

```ts
export interface MCPStdioServerConfig extends MCPServerConfigBase {
  command: string;
  args: string[];   // required
  env?: Record<string, string>;
  cwd?: string;
}
```

This is inconsistent with the MCP TypeScript SDK's own `StdioServerParameters`, which defines `args` as optional:

> [`modelcontextprotocol/typescript-sdk` — `src/client/stdio.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/stdio.ts):
> ```ts
> export type StdioServerParameters = {
>   command: string;
>   args?: string[];  // optional
>   // ...
> };
> ```

A stdio MCP server invoked with no arguments is a valid configuration per the spec, and omitting `args` is explicitly supported by the reference implementation.

**Observed behaviour:** Omitting `args` from a stdio server config causes the SDK's schema validation to reject the entire config with an `invalid_union` error (`Invalid MCP config … args: Required`).

**Expected behaviour:** `args` defaults to `[]` when omitted, consistent with `StdioServerParameters`.

**Suggested fix:** Change `args: string[]` to `args?: string[]` in `MCPStdioServerConfig`, treating a missing value as `[]` at the spawn call site.

Will-thom:
## Summary
- Make Node MCPStdioServerConfig.args optional in the public type
- Normalize omitted stdio MCP args to [] before sending create/resume payloads in Node and Python
- Serialize nil Go MCP stdio args as [] instead of null

Fixes #1231

## Tests
- npm test -- client.test.ts
- npm run typecheck
- node_modules\.bin\prettier.CMD --check src\client.ts src\types.ts test\client.test.ts
- uv run --extra dev pytest -q test_client.py
- uv run --extra dev ruff check copilot\client.py test_client.py
- uv run --extra dev ty check copilot
- go test .
- git diff --check

### DIFF
--- a/go/types.go
+++ b/go/types.go
@@ -489,6 +489,9 @@ func (MCPStdioServerConfig) mcpServerConfig() {}
 // MarshalJSON implements json.Marshaler, injecting the "type" discriminator.
 func (c MCPStdioServerConfig) MarshalJSON() ([]byte, error) {
 	type alias MCPStdioServerConfig
+	if c.Args == nil {
+		c.Args = []string{}
+	}
 	return json.Marshal(struct {
 		Type string `json:"type"`
 		alias
--- a/go/types_test.go
+++ b/go/types_test.go
@@ -123,6 +123,31 @@ func TestProviderConfig_JSONIncludesHeaders(t *testing.T) {
 	}
 }
 
+func TestMCPStdioServerConfig_JSONDefaultsNilArgsToEmptyArray(t *testing.T) {
+	config := MCPStdioServerConfig{
+		Command: "mcp-server",
+		Tools:   []string{"*"},
+	}
+
+	data, err := json.Marshal(config)
+	if err != nil {
+		t.Fatalf("failed to marshal MCP stdio server config: %v", err)
+	}
+
+	var decoded map[string]any
+	if err := json.Unmarshal(data, &decoded); err != nil {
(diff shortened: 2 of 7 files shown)


stephentoub:
Thanks for helping out, @Will-thom. We actually already made args optional on the runtime side, and were just waiting to pull in the new package, which we did this morning. With that, the fix is to make the properties in the SDK optional as well, rather than forcing an empty collection into existence. As such, I'm going to close this PR and let https://github.com/github/copilot-sdk/pull/1347 be the fix, but we appreciate your contribution, regardless.

Will-thom:
Thanks for the explanation, @stephentoub. That makes total sense.

I appreciate you taking the time to review the PR and clarify the intended SDK-side fix. Glad to hear the runtime-side change has landed, and closing this in favor of #1347 sounds good to me.

Thanks again!

PR #1401, opened 2026-05-23
Add SDK canvas runtime support

jmoseley:
## What changed

Adds canvas runtime support across the **Rust, Node, Python, Go, and .NET** SDKs so consumers can declare canvases on `session.create` / `session.resume`, install a single `CanvasHandler`, and receive routed `canvas.open` / `canvas.close` / `canvas.action.invoke` JSON-RPC requests from the runtime. Resume surfaces `openCanvases` for rehydration, and host code can call `session.canvas.*` for native UI / chrome actions.

### Per-language surface

- **Rust** (`rust/src/canvas.rs`): `CanvasDeclaration`, `CanvasHandler` trait, `CanvasHostContext`/`CanvasOpenContext`/`CanvasActionContext`/`CanvasLifecycleContext`, `CanvasError`, `CanvasOpenResponse`, `ExtensionInfo`. Wire types (`OpenCanvasInstance`, `CanvasAction`, `CanvasInstanceAvailability`, `CanvasInvokeActionRequest`/`Result`, etc.) consumed directly from `generated::api_types` — no duplication. `Session::open_canvases()` accessor; host APIs via `session.rpc().canvas()`.
- **Node** (`nodejs/src/canvas.ts`): `createCanvas`, canvas declarations, handler contexts, errors, exports. Session create/resume thread `canvases`, `requestCanvasRenderer`, `requestExtensions`, `extensionInfo`.
- **Python** (`python/copilot/canvas.py`): `CanvasHandler` ABC with default `on_close` / `on_action` (raises `canvas_action_no_handler`). `_jsonrpc` loosened to allow arbitrary JSON return values for action results.
- **Go** (`go/canvas.go`): `CanvasHandler` interface + `CanvasHandlerDefaults` embeddable for partial impls. Inbound RPC registered via `jsonrpc2.RequestHandlerFor` with `{code,message}` error envelope.
- **.NET** (`dotnet/src/Canvas.cs`): idiomatic `ICanvasHandler` with async methods; same dispatch shape.

Java is intentionally not included in this PR and will follow separately.

### Shared design

- App-side dispatch: SDK exposes a single `CanvasHandler` per session. Consumers switch on `canvas_id` themselves — no per-canvas registry inside the SDK.
- Generated wire types are imported directly from each language's codegen module; the hand-written canvas API is purely additive.
- Inbound RPC without a handler installed returns `{"code":"canvas_handler_unset","message":"..."}`.
- `ExtensionInfo { source, name }` lets consumers opt into stable agent-facing extension IDs (`github-app:<provider-name>`).

## Notes for reviewers

Additive on top of the existing permission / user-input / tool handler models — no broad handler refactor. Canvas declarations require `displayName`. `capabilities.ui.canvases` is boolean. `session.canvas.listOpen` is a real RPC.

## Validation

Per language:
- Rust: `cargo +nightly-2026-04-14 fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo test --all-features`, `cargo doc --no-deps --all-features`
- Node: `npm run typecheck`, `npm run lint`, `npx vitest run`
- Python: `uv run ruff format/check`, `uv run ty check copilot`, `uv run pytest`
- Go: `gofmt`, `go build/vet/test ./...`
- .NET: `dotnet format --verify-no-changes`, `dotnet build`, `dotnet test`

### DIFF
--- a/dotnet/src/Canvas.cs
+++ b/dotnet/src/Canvas.cs
@@ -1,0 +1,288 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------------------------------------------*/
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.Copilot.Rpc;
+
+namespace GitHub.Copilot;
+
+/// <summary>
+/// Declarative metadata for a single canvas, sent over the wire on
+/// <c>session.create</c> / <c>session.resume</c>.
+/// </summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasDeclaration
+{
+    /// <summary>Canvas identifier, unique within the declaring connection.</summary>
+    [JsonPropertyName("id")]
+    public string Id { get; set; } = string.Empty;
+
+    /// <summary>Human-readable name shown in host UI and canvas pickers.</summary>
+    [JsonPropertyName("displayName")]
+    public string DisplayName { get; set; } = string.Empty;
+
+    /// <summary>Short, single-sentence description shown to the agent in canvas catalogs.</summary>
+    [JsonPropertyName("description")]
+    public string Description { get; set; } = string.Empty;
+
+    /// <summary>JSON Schema for the <c>input</c> payload accepted by <c>canvas.open</c>.</summary>
+    [JsonPropertyName("inputSchema")]
+    public JsonElement? InputSchema { get; set; }
+
+    /// <summary>Agent-callable actions this canvas exposes.</summary>
+    [JsonPropertyName("actions")]
+    public IList<CanvasAction>? Actions { get; set; }
+}
+
+/// <summary>
+/// Stable extension identity for session participants that provide canvases.
+/// </summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class ExtensionInfo
+{
+    /// <summary>Extension namespace/source, e.g. <c>"github-app"</c>.</summary>
+    [JsonPropertyName("source")]
+    public string Source { get; set; } = string.Empty;
+
+    /// <summary>Stable provider name within the source namespace.</summary>
+    [JsonPropertyName("name")]
+    public string Name { get; set; } = string.Empty;
+}
+
+/// <summary>Response returned from <see cref="ICanvasHandler.OnOpenAsync"/>.</summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasOpenResponse
+{
+    /// <summary>URL the host should render. Optional for canvases with no visual surface.</summary>
+    [JsonPropertyName("url")]
+    public string? Url { get; set; }
+
+    /// <summary>Provider-supplied title shown in host chrome.</summary>
+    [JsonPropertyName("title")]
+    public string? Title { get; set; }
+
+    /// <summary>Provider-supplied status text shown in host chrome.</summary>
+    [JsonPropertyName("status")]
+    public string? Status { get; set; }
+}
+
+/// <summary>Host capabilities passed to canvas provider callbacks.</summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasHostContext
+{
+    /// <summary>Host capability details.</summary>
+    [JsonPropertyName("capabilities")]
+    public CanvasHostCapabilities Capabilities { get; set; } = new();
+}
+
+/// <summary>Host capability details passed to canvas provider callbacks.</summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasHostCapabilities
+{
+    /// <summary>Whether the host supports canvas rendering.</summary>
+    [JsonPropertyName("canvases")]
+    public bool Canvases { get; set; }
+}
+
+/// <summary>Context handed to <see cref="ICanvasHandler.OnOpenAsync"/>.</summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasOpenContext
+{
+    /// <summary>Session that requested the canvas.</summary>
+    public string SessionId { get; init; } = string.Empty;
+
+    /// <summary>Owning provider identifier.</summary>
+    public string ExtensionId { get; init; } = string.Empty;
+
+    /// <summary>Canvas id from the declaring <see cref="CanvasDeclaration"/>.</summary>
+    public string CanvasId { get; init; } = string.Empty;
+
+    /// <summary>Stable instance id supplied by the runtime.</summary>
+    public string InstanceId { get; init; } = string.Empty;
+
+    /// <summary>Validated input payload.</summary>
+    public JsonElement Input { get; init; }
+
+    /// <summary>Host capabilities supplied by the runtime.</summary>
+    public CanvasHostContext? Host { get; init; }
+}
+
+/// <summary>Context handed to <see cref="ICanvasHandler.OnActionAsync"/>.</summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasActionContext
+{
+    /// <summary>Session that invoked the action.</summary>
+    public string SessionId { get; init; } = string.Empty;
+
+    /// <summary>Owning provider identifier.</summary>
+    public string ExtensionId { get; init; } = string.Empty;
+
+    /// <summary>Canvas id targeted by the action.</summary>
+    public string CanvasId { get; init; } = string.Empty;
+
+    /// <summary>Instance id targeted by the action.</summary>
+    public string InstanceId { get; init; } = string.Empty;
+
+    /// <summary>Action name from <see cref="CanvasAction.Name"/>.</summary>
+    public string ActionName { get; init; } = string.Empty;
+
+    /// <summary>Validated input payload.</summary>
+    public JsonElement Input { get; init; }
+
+    /// <summary>Host capabilities supplied by the runtime.</summary>
+    public CanvasHostContext? Host { get; init; }
+}
+
+/// <summary>Context handed to a canvas's close lifecycle hook.</summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasLifecycleContext
+{
+    /// <summary>Session owning the canvas instance.</summary>
+    public string SessionId { get; init; } = string.Empty;
+
+    /// <summary>Owning provider identifier.</summary>
+    public string ExtensionId { get; init; } = string.Empty;
+
+    /// <summary>Canvas id from the declaring <see cref="CanvasDeclaration"/>.</summary>
+    public string CanvasId { get; init; } = string.Empty;
+
+    /// <summary>Instance id this lifecycle event applies to.</summary>
+    public string InstanceId { get; init; } = string.Empty;
+
+    /// <summary>Host capabilities supplied by the runtime.</summary>
+    public CanvasHostContext? Host { get; init; }
+}
+
+/// <summary>Structured error returned from canvas handlers.</summary>
+/// <remarks>
+/// Throw this from <see cref="ICanvasHandler"/> implementations to surface a
+/// machine-readable error code to the runtime. Any other exception is wrapped
+/// in a generic <c>canvas_handler_error</c> envelope.
+/// </remarks>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasError : Exception
+{
+    /// <summary>Initializes a new <see cref="CanvasError"/>.</summary>
+    /// <param name="code">Machine-readable error code.</param>
+    /// <param name="message">Human-readable message.</param>
+    public CanvasError(string code, string message) : base(message)
+    {
+        Code = code;
+    }
+
+    /// <summary>Machine-readable error code.</summary>
+    public string Code { get; }
+
+    /// <summary>
+    /// Default error returned when a custom action has no handler.
+    /// </summary>
+    public static CanvasError NoHandler() => new(
+        "canvas_action_no_handler",
+        "No handler implemented for this canvas action");
+}
+
+/// <summary>
+/// Internal helpers used by the session runtime to translate <see cref="CanvasError"/>
+/// (and other handler-thrown exceptions) into structured JSON-RPC error responses.
+/// </summary>
+internal static class CanvasErrorHelpers
+{
+    private const int InternalError = -32603;
+
+    public static LocalRpcInvocationException HandlerUnset() => Build(
+        "canvas_handler_unset",
+        "No canvas handler is registered on this session");
+
+    public static LocalRpcInvocationException HandlerError(string message) => Build(
+        "canvas_handler_error",
+        message);
+
+    public static LocalRpcInvocationException ToRpcException(CanvasError error) => Build(error.Code, error.Message);
+
+    private static LocalRpcInvocationException Build(string code, string message)
+    {
+        var json = JsonSerializer.Serialize(
+            new CanvasErrorPayload { Code = code, Message = message },
+            CanvasJsonContext.Default.CanvasErrorPayload);
+        using var doc = JsonDocument.Parse(json);
+        return new LocalRpcInvocationException(InternalError, message, doc.RootElement.Clone());
+    }
+
+    internal sealed class CanvasErrorPayload
+    {
+        [JsonPropertyName("code")]
+        public string Code { get; set; } = string.Empty;
+
+        [JsonPropertyName("message")]
+        public string Message { get; set; } = string.Empty;
+    }
+}
+
+[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+[JsonSerializable(typeof(CanvasErrorHelpers.CanvasErrorPayload))]
+internal partial class CanvasJsonContext : JsonSerializerContext;
+
+/// <summary>
+/// Provider-side canvas lifecycle handler.
+/// </summary>
+/// <remarks>
+/// A session installs a single <see cref="ICanvasHandler"/> via
+/// <c>SessionConfigBase.CanvasHandler</c>. The handler receives every
+/// inbound <c>canvas.open</c> / <c>canvas.close</c> / <c>canvas.action.invoke</c>
+/// JSON-RPC request the runtime issues for this session and decides — typically
+/// by inspecting <see cref="CanvasOpenContext.CanvasId"/> — which
+/// application-side canvas should handle the call.
+/// <para>
+/// The SDK does not maintain a per-canvas registry; multiplexing across
+/// declared canvases is the implementor's responsibility.
+/// </para>
+/// <para>
+/// Implementations targeting <c>netstandard2.0</c> cannot rely on default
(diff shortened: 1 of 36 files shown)


rust/src/session.rs
```diff
@@ -616,6 +636,119 @@ impl Drop for Session {
     }
 }
 
+/// Canvas host sub-API for a [`Session`].
+///
+/// Acquired via [`Session::canvas`]. Methods route to `session.canvas.*`
+/// RPCs, except [`list_open`](Self::list_open), which returns the SDK's local
+/// snapshot from resume and calls made through this handle.
```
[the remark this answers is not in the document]
jmoseley: Accepted. I updated the  docs so  is described as calling  and refreshing the SDK snapshot from the RPC response.

rust/src/canvas.rs
```diff
@@ -0,0 +1,727 @@
...
+        .on_open(CanvasOpenContext {
+            session_id: params.session_id,
+            extension_id: params.extension_id,
+            canvas_id: params.canvas_id,
+            instance_id: params.instance_id,
+            input: params.input,
+            host: params.host,
+        })
+        .await?;
+    Ok(serde_json::to_value(response).unwrap_or(Value::Null))
```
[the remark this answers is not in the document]
jmoseley: Accepted.  now maps response serialization failures into a  instead of silently returning .

rust/src/canvas.rs
```diff
@@ -0,0 +1,727 @@
...
+        self.handler = Some(handler);
+        self
+    }
+
+    /// Finalize into a [`Canvas`].
+    pub fn build(self) -> Canvas {
+        let handler = self
+            .handler
+            .expect("Canvas::builder().handler(...) must be called before build()");
+        Canvas {
```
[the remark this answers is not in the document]
jmoseley: Accepted.  now returns  with a  error when no handler was attached, and the Rust tests/usages were updated.

nodejs/src/client.ts
```diff
@@ -1880,6 +1896,24 @@ export class CopilotClient {
...
+        this.connection.onRequest("canvas.reload", async (params: CanvasProviderRequestParams) =>
+            this.handleCanvasProviderRequest("canvas.reload", params)
+        );
+        this.connection.onRequest("canvas.close", async (params: CanvasProviderRequestParams) =>
+            this.handleCanvasProviderRequest("canvas.close", params)
+        );
+        this.connection.onRequest(
+            "canvas.action.invoke",
+            async (params: CanvasActionInvokeParams) =>
+                this.handleCanvasProviderRequest(params.actionName, params)
```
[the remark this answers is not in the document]
jmoseley: Accepted. The  request path now validates the payload with a guard before reading , so malformed input reports the intended invalid payload error.

nodejs/src/client.ts
```diff
@@ -2083,4 +2117,30 @@ export class CopilotClient {
...
+
+    private async handleCanvasProviderRequest(
+        actionName: string,
+        params: CanvasActionInvokeParams | CanvasProviderRequestParams
+    ): Promise<unknown> {
+        if (
+            !params ||
+            typeof params.sessionId !== "string" ||
+            typeof params.canvasId !== "string" ||
+            typeof params.instanceId !== "string"
```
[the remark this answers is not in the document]
jmoseley: Accepted. The shared canvas provider payload validation now requires  to be a string before dispatching to user callbacks.

rust/src/session.rs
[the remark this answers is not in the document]
jmoseley: Correction to my previous reply: accepted. I updated the `SessionCanvas` docs so `list_open` is described as calling `session.canvas.listOpen` and refreshing the SDK snapshot from the RPC response.

rust/src/canvas.rs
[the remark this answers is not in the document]
jmoseley: Correction to my previous reply: accepted. `dispatch_canvas_open` now maps response serialization failures into a `CanvasError` instead of silently returning `null`.

rust/src/canvas.rs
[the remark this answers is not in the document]
jmoseley: Correction to my previous reply: accepted. `CanvasBuilder::build()` now returns `CanvasResult<Canvas>` with a `canvas_builder_missing_handler` error when no handler was attached, and the Rust tests/usages were updated.

nodejs/src/client.ts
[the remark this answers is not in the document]
jmoseley: Correction to my previous reply: accepted. The `canvas.action.invoke` request path now validates the payload with a guard before reading `actionName`, so malformed input reports the intended invalid payload error.

nodejs/src/client.ts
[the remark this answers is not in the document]
jmoseley: Correction to my previous reply: accepted. The shared canvas provider payload validation now requires `extensionId` to be a string before dispatching to user callbacks.

nodejs/src/session.ts
```diff
@@ -635,6 +637,33 @@ export class CopilotSession {
...
+     * @param canvasId - The id of the canvas to retrieve
+     * @returns The registered Canvas if found, or undefined
+     * @internal Used by the SDK's direct `canvas.*` dispatcher.
+     */
+    getCanvas(canvasId: string): Canvas | undefined {
+        return this.canvases.get(canvasId);
+    }
+
     /**
      * Registers command handlers for this session.
```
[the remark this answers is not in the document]
jmoseley: Good catch — accepted. Added `openCanvases` getter and `setOpenCanvases` (internal) on `CopilotSession`, and the resume handler in `CopilotClient.resumeSession` now populates it from the `session.resume` response. Pushed in 04fc7481.

nodejs/src/canvas.ts
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * Optional. Notified when a canvas instance is closed by the user, the
+     * agent, or the host. Fire-and-forget: the return value is ignored and
+     * errors are logged but not surfaced to the runtime.
+     */
+    onClose?: (ctx: CanvasLifecycleContext) => Promise<void> | void;
+}
+
+/** A registered canvas: declarative metadata + in-process handler closures. */
+export class Canvas {
```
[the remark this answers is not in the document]
jmoseley: Partially accepted: documenting the divergence, keeping the API.

Node's per-canvas `createCanvas` / `Canvas` factory pattern is intentional and predates the unified handler model in the other SDKs. It mirrors `DefineTool`'s co-location ergonomics and we want to keep it. Added a JSDoc note on both `Canvas` and `createCanvas` explaining that other SDKs (Rust, Python, Go, .NET) expose a single `CanvasHandler` per session that switches on `canvasId`, and that both shapes target the same JSON-RPC wire protocol. Pushed in 04fc7481.

Not changing the other SDKs' shape — that's a much larger ergonomics-only refactor and out of scope for this PR.

EvanBoyle (approved):

dotnet/src/Canvas.cs
```diff
@@ -0,0 +1,288 @@
...
+    public CanvasHostCapabilities Capabilities { get; set; } = new();
+}
+
+/// <summary>Host capability details passed to canvas provider callbacks.</summary>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasHostCapabilities
+{
+    /// <summary>Whether the host supports canvas rendering.</summary>
+    [JsonPropertyName("canvases")]
+    public bool Canvases { get; set; }
```
stephentoub, line 90: This name confuses me a bit. Should this be specific to rendering, as elaborated on in the comment? It's not clear to me from "Canvases" that it's specific to rendering, and presumably there are other capabilities a canvas host will be capable of related to canvases.

dotnet/src/Canvas.cs
```diff
@@ -0,0 +1,288 @@
...
+    public string CanvasId { get; init; } = string.Empty;
+
+    /// <summary>Stable instance id supplied by the runtime.</summary>
+    public string InstanceId { get; init; } = string.Empty;
+
+    /// <summary>Validated input payload.</summary>
+    public JsonElement Input { get; init; }
+
+    /// <summary>Host capabilities supplied by the runtime.</summary>
+    public CanvasHostContext? Host { get; init; }
```
stephentoub, line 113: I don't think we've used `init` elsewhere in the C# SDK; these should probably all be set.

dotnet/src/Canvas.cs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/// <summary>Structured error returned from canvas handlers.</summary>
+/// <remarks>
+/// Throw this from <see cref="ICanvasHandler"/> implementations to surface a
+/// machine-readable error code to the runtime. Any other exception is wrapped
+/// in a generic <c>canvas_handler_error</c> envelope.
+/// </remarks>
+[Experimental(Diagnostics.Experimental)]
+public sealed class CanvasError : Exception
```
stephentoub, line 169: If we need a dedicated exception type, it should be suffixed with Exception rather than Error

dotnet/src/Canvas.cs
```diff
@@ -0,0 +1,288 @@
...
+        public string Code { get; set; } = string.Empty;
+
+        [JsonPropertyName("message")]
+        public string Message { get; set; } = string.Empty;
+    }
+}
+
+[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase)]
+[JsonSerializable(typeof(CanvasErrorHelpers.CanvasErrorPayload))]
+internal partial class CanvasJsonContext : JsonSerializerContext;
```
stephentoub, line 229: This can be simplified to something like:
```csharp
JsonElement payload = JsonSerializer.SerializeToElement(
    new JsonObject()
    {
        ["code"] = code,
        ["message"] = message,
    },
    TypesJsonContext.Default.JsonObject);
return new LocalRpcInvocationException(InternalError, message, payload);
```
and add JsonObject as an attribute on TypesJsonContext if it's not already there.

dotnet/src/Canvas.cs
```diff
@@ -0,0 +1,288 @@
...
+    /// </summary>
+    Task<object?> OnActionAsync(CanvasActionContext context, CancellationToken cancellationToken);
+}
+
+/// <summary>
+/// Convenience base class for <see cref="ICanvasHandler"/> that supplies
+/// default no-op / no-handler implementations of the optional callbacks.
+/// </summary>
+[Experimental(Diagnostics.Experimental)]
+public abstract class CanvasHandlerBase : ICanvasHandler
```
stephentoub, line 272: Is this necessary? Doesn't seem like it provides much in the way of convenience.

dotnet/src/JsonRpc.cs
```diff
@@ -486,13 +486,21 @@ private async Task HandleIncomingMethodAsync(string methodName, JsonElement mess
             }
             catch (Exception ex) when (ex is not OperationCanceledException)
             {
+                var actual = ex is TargetInvocationException tie && tie.InnerException != null ? tie.InnerException : ex;
```
stephentoub, line 489: Where is the TargetInvocationException coming from? Are we calling something via reflection?

dotnet/src/Session.cs
```diff
@@ -861,6 +880,125 @@ internal void SetCapabilities(SessionCapabilities? capabilities)
...
+
+    private static JsonElement SerializeActionResult(object? value)
+    {
+        var element = CopilotClient.ToJsonElementForWire(value);
+        if (element.HasValue)
+        {
+            return element.Value;
+        }
+        using var doc = JsonDocument.Parse("null");
+        return doc.RootElement.Clone();
```
stephentoub, line 998: If this is expected to be common, it could just cache a JsonElement for null

rust/src/canvas.rs
```diff
@@ -0,0 +1,380 @@
...
+
+use crate::generated::api_types::CanvasAction;
+use crate::types::SessionId;
+
+/// JSON Schema object used for canvas inputs and canvas-scoped tools.
+pub type CanvasJsonSchema = serde_json::Map<String, Value>;
+
+/// Declarative metadata for a single canvas, sent over the wire on
+/// `session.create` / `session.resume`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
stephentoub, line 16: Should we implement Default like we do for all of the generated types, as part of future proofing against additional optional members added in the future? (This applies to all the structs in this file)

rust/src/canvas.rs
```diff
@@ -0,0 +1,380 @@
...
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct CanvasDeclaration {
```
stephentoub, line 19: We have comments on all the RPC types like:
```rust
/// **Experimental.** This type is part of an experimental wire-protocol surface
/// and may change or be removed in future SDK or CLI releases.
```
Should we add that to all of this surface area as well? Or do we believe this is all immediately stable with GA in a week?

(Same question applies to TypeScript, Python, and Go. For C#, you already have [Experimental] on there, which is good.)

dotnet/test/Unit/SerializationTests.cs
```diff
@@ -203,6 +203,34 @@ public void ResumeSessionRequest_CanSerializeEnableSessionTelemetry_WithSdkOptio
...
+            ("OpenCanvases", instances));
+
+        var json = JsonSerializer.Serialize(request, requestType, options);
+        using var document = JsonDocument.Parse(json);
+        var root = document.RootElement;
+        var openCanvases = root.GetProperty("openCanvases");
+        Assert.Equal(1, openCanvases.GetArrayLength());
+        Assert.Equal("canvas-id", openCanvases[0].GetProperty("canvasId").GetString());
+    }
+
```
stephentoub, line 233: Is there a way we can add some good E2E tests in the SDK for this canvas support, in each language? We have a fairly extensive set of hundreds of E2E tests in each of the languages, and while we have some gaps we're working to fill, we're trying to make it so that we have great E2E coverage for every API exposed.

nodejs/src/canvas.ts
```diff
@@ -0,0 +1,286 @@
...
+    inputSchema?: CanvasJsonSchema;
+    /** Required per-action dispatch handler. */
+    handler: (ctx: CanvasActionContext) => Promise<unknown> | unknown;
+}
+
+/**
+ * Declarative metadata for a single canvas, serialized over the wire on
+ * `session.create` / `session.resume`.
+ */
+export interface CanvasDeclaration {
```
SteveSandersonMS, line 43: I see that the .NET equivalents for all these types are marked as experimental, but the Node/Python/Rust ones are not. Should we mark the whole canvas API surface as experimental across all languages? I'm guessing there isn't anything uniquely experimental about .NET.

dotnet/test/Unit/SerializationTests.cs
SteveSandersonMS: I added a tiny bit of E2E coverage in https://github.com/github/copilot-sdk/pull/1413 but @jmoseley will know better what other coverage would be valuable to add.

rust/src/canvas.rs
jmoseley: Done in #1420 — added doc-comment Experimental warnings on every hand-written canvas type/trait in `rust/src/canvas.rs`, matching the wording used on the generated wire-protocol types in `api_types.rs`.

nodejs/src/canvas.ts
jmoseley: Done in #1420 — added `@experimental` JSDoc tags on every exported canvas type/class/function in `nodejs/src/canvas.ts`. Also added equivalents in Python (per-class `# Experimental: ...` comments matching `generated/rpc.py`'s convention) and Go (per-symbol `// Experimental: ...` doc comments matching `rpc/zrpc.go`).

rust/src/canvas.rs
jmoseley: Done in #1420 — derived `Default` on `CanvasDeclaration`, `CanvasOpenContext`, `CanvasActionContext`, and `CanvasLifecycleContext`. The host/response structs already had it; `CanvasError` is skipped because `code` doesn't have a sensible default.

nodejs/src/canvas.ts
```diff
@@ -0,0 +1,286 @@
...
+    /** Provider-supplied status text shown in host chrome. */
+    status?: string;
+}
+
+/** Host capabilities passed to canvas callbacks. */
+export interface CanvasHostContext {
+    capabilities?: {
+        canvases?: boolean;
+    };
+}
```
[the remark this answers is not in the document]
jmoseley: Done in #1420 — `CanvasHostCapabilities` is now a named exported interface and re-exported from `nodejs/src/index.ts`. Matches Rust/Python/Go/.NET.

dotnet/src/Canvas.cs
jmoseley: Done in #1420 — converted all `{ get; init; }` setters on `CanvasOpenContext`, `CanvasActionContext`, and `CanvasLifecycleContext` to `{ get; set; }`.

dotnet/src/Canvas.cs
jmoseley: Done in #1420 — renamed `CanvasError` → `CanvasException`. Internal helpers class kept its existing name (`CanvasErrorHelpers`) since it builds JSON-RPC error envelopes, not exceptions per se. Other SDKs keep `CanvasError` (Rust/Node/Python/Go) since those ecosystems don't use the `Exception` suffix convention.

dotnet/src/Canvas.cs
jmoseley: Done in #1420 — switched to your suggested form using `JsonObject` + `TypesJsonContext.Default.JsonObject`, added `[JsonSerializable(typeof(JsonObject))]` to `TypesJsonContext`, and dropped the bespoke `CanvasErrorPayload` + now-empty `CanvasJsonContext`.

dotnet/src/JsonRpc.cs
jmoseley: Yes — JSON-RPC method handlers are dispatched via reflection (`Delegate.DynamicInvoke` in `InvokeHandlerAsync`, plus `MethodInfo.Invoke` for `ValueTask<T>.AsTask()` and `Task<T>.Result`), which wraps any handler exception in `TargetInvocationException`. Added a comment in #1420 explaining this so future readers don't have to dig.

dotnet/src/Session.cs
jmoseley: Done in #1420 — hoisted into `private static readonly JsonElement NullJsonElement` (built once via `JsonSerializer.SerializeToElement<object?>(null, TypesJsonContext.Default.Object)`); `SerializeActionResult` now returns it directly instead of `JsonDocument.Parse("null")` per call.

dotnet/src/Canvas.cs
jmoseley: `CanvasHandlerBase` is there for `netstandard2.0` consumers — that target can't use default interface methods on `ICanvasHandler`, so without the base class every netstandard2.0 user has to write the `OnCloseAsync` no-op + `OnActionAsync` throw boilerplate themselves. Happy to drop it (and just put the boilerplate in the README) if you'd rather, but leaving as-is in #1420 — let me know if you want it removed in a follow-up.

stephentoub: Thanks, I think we should drop it. Implementing this will be rare, and the cited boilerplate is a trivial one-liner. Better to not add additional surface area to the namespace.

PR #1438, opened 2026-05-26
Add MessageOptions.agentMode and fix per-message mode misuse

MRayermannMSFT:
## What

Adds a public `agentMode` field to `MessageOptions` across all six SDKs (.NET, Node, Python, Go, Java, Rust) so callers can set the per-message UI mode (`interactive` / `plan` / `autopilot` / `shell`) without abusing the queue-delivery `mode` field. The new field is plumbed through `session.send` to the JSON-RPC `agentMode` parameter, matching the runtime's existing wire schema. The misleading XML doc on the .NET `MessageOptions.Mode` property is corrected. Four pre-existing E2E/unit tests in .NET, Node, Python, and Go that were setting `mode = "plan"` (variously via `as unknown as` casts, `# type: ignore`, or silently as a `string?` in .NET) are migrated to use the new `agentMode` field and updated to assert the runtime echoes the value back on the `user.message` event.

## Why

Every SDK exposed a per-message `mode` field but none exposed `agentMode` — so callers who wanted to start a turn in plan/autopilot/etc. had no correct way to do it from the SDKs, and four tests across .NET/Node/Python/Go reached for `mode` instead. The runtime's `mode` only accepts `enqueue` / `immediate`, so those calls had no effect on agent mode and were only "working" because a runtime regression exposed `exit_plan_mode` in every session mode. Once that runtime regression was fixed (in copilot-agent-runtime#8809), the `Should_Invoke_Exit_Plan_Mode_Handler_When_Model_Uses_Tool` test in the .NET SDK began failing because the session was still effectively `interactive` and the model was no longer offered the `exit_plan_mode` tool. Exposing `agentMode` on every SDK's surface gives callers (and the failing test) the correct API to express that intent, closes the same gap in Java and Rust before anyone else trips on it, and the corrected docs prevent the next caller from making the same mistake.

### DIFF
--- a/dotnet/src/Session.cs
+++ b/dotnet/src/Session.cs
@@ -276,6 +276,7 @@ public sealed partial class CopilotSession : IAsyncDisposable
             Prompt = options.Prompt,
             Attachments = options.Attachments,
             Mode = options.Mode,
+            AgentMode = options.AgentMode,
             Traceparent = traceparent,
             Tracestate = tracestate,
             RequestHeaders = options.RequestHeaders,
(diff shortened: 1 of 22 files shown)


MRayermannMSFT:
> This PR correctly adds agentMode to .NET, Node.js, Python, and Go. However, the Java SDK is missing the equivalent change and would be left behind.

Fine, I'll do it for all languages.

patniko (approved):

SteveSandersonMS (approved):

PR #1445, opened 2026-05-27
feat(rust): route cloud sessions through create_session

chuwik:
## Summary

- Adds Rust cloud session config and wire support via `SessionConfig::with_cloud`.
- Routes cloud configs through `Client::create_session` while keeping cloud creation as an internal helper.
- Uses the runtime-assigned cloud session ID returned by `session.create` for registration and session state.

## Context

Builds on [#1394](https://github.com/github/copilot-sdk/pull/1394) while preserving the existing public `Client::create_session` API.


tiagonbotelho (approved):
LGTM

PR #1455, opened 2026-05-27
Rust SDK: add typed per-session capability controls

Morabbin:
## Summary

Adds typed `SessionConfig` / `ResumeSessionConfig` fields and builder methods backed by the generated `SessionCapability` enum, so callers can express per-session capability opt-in / opt-out without stringly-typed flags.

This is a per-session API: the capability fields are sent as JSON-RPC wire parameters on `session.create` and `session.resume`, so it works for every transport including `Transport::External`.

> **Experimental.** The new capability enable/disable surface is part of an experimental wire-protocol API. The Rust SDK marks the new fields and builders with the repository's `<div class="warning">**Experimental.**</div>` rustdoc block (the convention used in `rust/src/generated/`); there is no `#[experimental]` attribute, so the marker is documentation-only.

## Why

Some SDK consumers use `Transport::External`, so spawn-time CLI flags have no effect. A per-session wire API is the correct approach for session-scoped runtime capabilities.

### `SessionCapability` enum

- Reuses the generated protocol enum and re-exports it from the crate root as `github_copilot_sdk::SessionCapability`.
- Includes generated typed variants such as `Memory`, `PlanMode`, `CanvasRenderer`, `Elicitation`, `McpApps`, `SessionStore`, `TuiHints`, `CliDocumentation`, `AskUser`, `InteractiveMode`, and `SystemNotifications`.
- Preserves generated per-variant rustdoc and serde wire names.
- Rejects `SessionCapability::Unknown` as invalid outbound create/resume configuration, since that variant is only a deserialization fallback.

### `SessionConfig` and `ResumeSessionConfig` (`rust/src/types.rs`)

New fields:

- `enabled_capabilities: Vec<SessionCapability>` -- opt this session into additional capabilities.
- `disabled_capabilities: Vec<SessionCapability>` -- opt this session out of capabilities; disable wins over enable on overlap.

Four builders on each config type:

- `with_enable_capability`, `with_disable_capability`
- `with_enabled_capabilities`, `with_disabled_capabilities`

### Wire serialization (`rust/src/wire.rs`)

`SessionCreateWire` and `SessionResumeWire` gain:

- `enabledCapabilities?: SessionCapability[]`
- `disabledCapabilities?: SessionCapability[]`

Both fields are omitted from the wire when empty. The generated enum serde emits the protocol strings such as `memory`, `plan-mode`, and `canvas-renderer`.

## Runtime dependency

Requires a runtime version that supports `enabledCapabilities` and `disabledCapabilities`. On older runtimes the fields are silently ignored.

## Validation

- `cargo +nightly fmt --check`
- `cargo clippy --lib --tests -- -D warnings`
- `cargo test --lib` (154 passed)
- `cargo test --doc` (20 passed, 30 ignored)
- `cargo build`


rust/README.md
```diff
@@ -78,6 +78,96 @@ client.stop().await?;
...
+> releases. The Rust SDK marks the relevant fields and builders with a
+> `<div class="warning">**Experimental.**</div>` rustdoc block (the
+> repository convention used in `rust/src/generated/`); there is no
+> `#[experimental]` attribute, so the marker is documentation-only.
+
+> **Runtime dependency.** Per-session capability controls require
+> [github/copilot-agent-runtime#8918](https://github.com/github/copilot-agent-runtime/pull/8918)
+> or later. On older runtimes the fields are silently ignored.
+> Pairs with [github/agents#1081](https://github.com/github/agents/issues/1081)
+> (Desktop app missing memory capability).
```
SteveSandersonMS: Please avoid referring to closed-source repos anywhere in the open-source code.

Morabbin: Dang it! Will excise, apologies.

rust/src/lib.rs
```diff
@@ -402,6 +402,162 @@ impl OtelExporterType {
...
+/// appears in both the enabled and disabled lists, the disable wins.
+/// The SDK preserves the order callers add capabilities in so the
+/// resulting wire payload is deterministic.
+///
+/// The enum is `#[non_exhaustive]` and carries an [`Other`](Self::Other)
+/// variant so forward-compat capabilities the runtime grows ahead of an
+/// SDK release can still be opted into without waiting for a new
+/// enum variant.
+///
+/// Requires github/copilot-agent-runtime#8918 or later.
```
SteveSandersonMS: Please avoid referring to closed source. There are a few comments like this.

Morabbin: Fix incoming; removing all mentions from all parts of the PR

rust/src/lib.rs
```diff
@@ -402,6 +402,162 @@ impl OtelExporterType {
...
+    SessionStore,
+    /// MCP-Apps (SEP-1865) `ui://` resource passthrough.
+    McpApps,
+    /// Extension-provided canvases rendered by the host.
+    CanvasRenderer,
+    /// A capability name the SDK doesn't have a typed variant for yet.
+    ///
+    /// Pass any kebab-case capability string here to forward it
+    /// verbatim to the runtime.
+    Other(String),
```
SteveSandersonMS: Rather than duplicating the list of capabilities, could we just directly use `crate::generated::api_types::SessionCapability` in the API? That would mean it automatically brings forwards the annotations for individual entries being experimental, plus avoids having to continually maintain this code whenever the set of capabilities changes.

I know that won't have an equivalent to `Other(string)` but the SDK requires a matching runtime version, so in general we don't need to provide forward compatibility for this situation.

SteveSandersonMS: Additionally we plan soon to change session.create/session.resume to work via codegen instead of being hand-authored. At that point the set of capabilities would have to be a regular enum just like all the other enums we reference from those options (unless we also implement some nontrivial codegen changes across 6 languages).

Morabbin: Sounds good to me; change incoming

rust/README.md
Morabbin: Replying on Morabbin's behalf: fixed in the latest pushes. The README/source docs and PR body no longer add closed-source repository references.

rust/src/lib.rs
Morabbin: Replying on Morabbin's behalf: fixed in the latest pushes. The README/source docs and PR body no longer add closed-source repository references.

rust/src/lib.rs
Morabbin: Replying on Morabbin's behalf: updated this to use the generated `SessionCapability` enum directly, re-exported from the crate root. The custom duplicated enum, `Other(String)`, string conversions, and manual string serde are gone.

The create/resume config now stores generated `SessionCapability` values and the wire structs serialize those enum values directly. I also added coverage that `SessionCapability::CanvasRenderer` serializes to `canvas-renderer`, and reject `SessionCapability::Unknown` as invalid outbound config since it is only the generated deserialization fallback. This should line up with the planned generated create/resume path.

rust/src/types.rs
```diff
@@ -1889,6 +1945,87 @@ impl SessionConfig {
...
+    /// surface and may change or be removed in future SDK or CLI releases.
+    ///
+    /// </div>
+    ///
+    /// Requires runtime support for per-session capability controls.
+    pub fn with_enabled_capabilities<I>(mut self, capabilities: I) -> Self
+    where
+        I: IntoIterator<Item = SessionCapability>,
+    {
+        self.enabled_capabilities = capabilities.into_iter().collect();
```
SteveSandersonMS: Since `with_enable_capability` appends an item, I would have expected `with_enable_capabilities` also to append items. It seems surprising that it would also delete any existing items.

rust/README.md
```diff
@@ -78,6 +78,72 @@ client.stop().await?;
...
 With the default `CliProgram::Resolve`, `Client::start()` resolves the CLI in this order: an explicit `CliProgram::Path(path)`, the `COPILOT_CLI_PATH` env var, then the bundled CLI that was embedded at build time. There is no PATH scanning — if you've opted out of bundling (`default-features = false`) you must supply either `CliProgram::Path` or `COPILOT_CLI_PATH`.
 
+### Session capabilities
+
+The `SessionCapability` enum lets callers enable or disable named runtime features on a **per-session** basis. Capabilities are sent to the runtime as part of the `session.create` / `session.resume` JSON-RPC calls via two wire fields:
+
+- `enabledCapabilities` -- capabilities to opt the session into (extends the `SDK_CAPABILITIES` baseline)
+- `disabledCapabilities` -- capabilities to opt the session out of (disable wins on overlap)
+
+This approach works for every transport -- including `Transport::External` -- because it does not rely on CLI spawn arguments.
```
SteveSandersonMS: This is an implementation detail. Let's only document things that are relevant to SDK consumers.

rust/README.md
```diff
@@ -78,6 +78,72 @@ client.stop().await?;
...
+| `Memory`             | `memory`                | `store_memory` tool and `<memories>` system-prompt section |
+| `CliDocumentation`   | `cli-documentation`     | `fetch_copilot_cli_documentation` tool and `<self_documentation>` section |
+| `AskUser`            | `ask-user`              | `ask_user` tool for interactive clarification         |
+| `InteractiveMode`    | `interactive-mode`      | Interactive-CLI identity (vs headless)                |
+| `SystemNotifications`| `system-notifications`  | Automatic batched system notifications to the agent   |
+| `Elicitation`        | `elicitation`           | Elicitation prompts (confirm / select / input)        |
+| `SessionStore`       | `session-store`         | Cross-session history tools and session-store metadata |
+| `McpApps`            | `mcp-apps`              | MCP-Apps `ui://` resource passthrough (SEP-1865)      |
+| `CanvasRenderer`     | `canvas-renderer`       | Host-rendered extension canvases                      |
+| `Unknown`            | *(none)*                | Deserialization fallback; rejected for outbound config |
```
SteveSandersonMS: It's probably best not to list the full set in the docs, because then we have to keep updating the docs (or by default we don't, and the docs get stale). The enum is already self-documenting.

rust/README.md
```diff
@@ -714,6 +780,13 @@ gets to be Rust here — cross-SDK parity for these is a post-release
 conversation, not a release blocker. None of these are deprecated and
 none of them are scheduled for removal.
 
+- **`SessionCapability` enum** -- generated typed enum for per-session
+  capability opt-in / opt-out. Sent via `enabledCapabilities` /
```
SteveSandersonMS, line 765: Please don't document whether things are generated or not - that's an implementation detail and isn't relevant to SDK consumers.

rust/README.md
```diff
@@ -714,6 +780,13 @@ gets to be Rust here — cross-SDK parity for these is a post-release
...
+  `disabledCapabilities` on the `session.create` and `session.resume` wire
+  calls -- works for all transports including `Transport::External`. See
+  [Session capabilities](#session-capabilities) above. Marked
+  **experimental** via the repository's `<div class="warning">` rustdoc
+  convention. Node/Python/Go/.NET accept stringly-typed flags.
```
SteveSandersonMS: > Node/Python/Go/.NET accept stringly-typed flags.

Is that true? Even if so there's no reason for the Rust docs to discuss that.

rust/README.md
Morabbin: Fixed this and similar problems elsewhere in the README

rust/README.md
Morabbin: Fixed

rust/README.md
Morabbin: ACK

rust/src/types.rs
Morabbin: Ah, good call. Fixed.

Morabbin:
Closing in favour of a focused, per-feature approach: an optional, typed `memory` configuration on session create and resume (an extensible `MemoryConfiguration`), rather than a generic per-session capability collection. A new SDK PR will implement that surface.

Morabbin:
The replacement is now open: github/copilot-sdk#1617 (optional `memory` configuration on session create and resume).

PR #1502, opened 2026-05-29
Add session.extensions.sendAttachmentsToMessage SDK wrapper

jmoseley:
Regenerates the SDK clients to expose the new `session.extensions.sendAttachmentsToMessage` RPC endpoint added in copilot-agent-runtime PR https://github.com/github/copilot-agent-runtime/pull/8971.

This lets canvas extensions push attachments — including a new `extension_context` attachment variant — into the next user message's composer.

## What's in this PR

The extension-facing RPC block now exposes:

```ts
sendAttachmentsToMessage(params: {
    instanceId?: string;
    attachments: Array<
        | { type: "extension_context"; title: string; payload: unknown }
        | Attachment // any standard variant
    >;
}): Promise<void>;
```

Plus the wire types for the new `extension_context` attachment variant on both directions (params + `SessionExtensionsAttachmentsPushedEvent` event) so the SDK exports typed shapes consumers can rely on.

**Call path heads-up (JS):** the wrapper lives at `session.rpc.extensions.sendAttachmentsToMessage`, not `session.extensions.sendAttachmentsToMessage`. Sibling methods like `session.send` and `session.getEvents` are on the session class directly, so this is an easy mistake — worth flagging for anyone writing docs or examples.

## Languages regenerated

| Language    | Status |
|-------------|--------|
| TypeScript  | ✅ regenerated, built, tests + lint pass |
| .NET        | ✅ regenerated, `dotnet build` passes |
| Python      | ✅ regenerated, unit tests pass |
| Rust        | ✅ regenerated, `cargo build` passes |
| Java        | ✅ regenerated (mvn not run locally — please verify in CI) |
| Go          | ⚠️ **intentionally not regenerated** — see below |

## Go follow-up

The Go codegen at `scripts/codegen/go.ts` (~line 1840–1855, `emitGoFlatDiscriminatedUnion`) has a duplicate-method bug that surfaces with this schema: when a leaf type is a variant of two discriminated unions with the same discriminator property (here, `SendAttachment` and the new `PushAttachment` both use `"type"`), the generator emits `func (X) Type() FirstUnionEnum` and then `func (X) Type() SecondUnionEnum`, which fails to compile.

Possible fixes (all non-trivial, out of scope here):
- Rename the discriminator method per-union (e.g. `PushAttachmentType()`)
- Share the discriminator enum between unions when values coincide
- Skip per-variant method emission on the second union

I'd suggest landing this PR to unblock the github-app side and tracking the Go regen as a follow-up.

## Approach note

To keep the diff minimal and reviewable, I patched the new method/types into the installed `@github/copilot` schemas directly rather than bumping the dependency — the runtime branch carries ~500 lines of other in-flight changes that aren't part of this work. The patch is additive only and uses the SDK's existing naming conventions (`SendAttachment*`, `UserMessageAttachment*`) rather than runtime's `Attachment*`. When the next `@github/copilot` release ships, a normal dep bump + regen will absorb this cleanly.

## Coordination

Paired session in `github-app` has verified the new method end-to-end against a hand-patched local runtime build; this PR is the proper replacement for that patch.


SteveSandersonMS:
Everything's been regenerated since this, so I'll close.

PR #1520, opened 2026-05-30
Java: Make it so slash command responses are accessible via RPC

edburns:
### Java

Last minute gap analysis promted by this question by Zidong Li:

> Just curious whether you know how to trigger slash commands in SDK. There's one kind of skill that disables agents calling. Those skills can be invoked in copilot cli by /<skill> manually. But using SDKs, if I send the text directly to agents, it's just treated as plain text to LLM, but LLM actually doesn't have the skill in context so it will not invoke the skill. Did you know any conversation about this?

prompted me to investigate this. Turns out the ability to invoke slash commands was there, but the ability to obtain the responses was not.


edburns:
[20260530-slashCommandSupport.md](https://github.com/user-attachments/files/28435306/20260530-slashCommandSupport.md)

PR #1601, opened 2026-06-08
Add `@CopilotExperimental` compile-time gate for experimental APIs

edburns:
## Summary

This change adds `@CopilotExperimental` and a processor that rejects
unapproved references to experimental SDK APIs at compile time unless the
consumer opts in with `-Acopilot.experimental.allowed=true`.

### New files

| File | Purpose |
|---|---|
| `java/src/main/java/com/github/copilot/CopilotExperimental.java` | Public annotation for experimental API types and methods |
| `java/src/main/java/com/github/copilot/CopilotExperimentalProcessor.java` | JSR 269 processor that enforces the opt-in gate |
| `java/src/main/resources/META-INF/services/javax.annotation.processing.Processor` | Processor service registration |
| `java/docs/adr/adr-004-copilotexperimental.md` | ADR documenting the design choice and tradeoffs |
| `java/src/test/java/com/github/copilot/CopilotExperimentalProcessorTest.java` | Compile-time tests for fail-by-default and opt-in behavior |

### Updated files

| File | Change |
|---|---|
| `java/scripts/codegen/java.ts` | Emits `@CopilotExperimental` alongside existing experimental `@apiNote` Javadocs |
| `java/src/generated/java/**` | Regenerated to annotate experimental types and methods |
| `java/pom.xml` | Compiler config updated for internal opt-in build usage |
| `java/src/main/java/module-info.java` | Removed the extra compiler-module dependency |
| `java/README.md` | Documents opt-in usage, caught cases, and known declaration-level limitations |

## How it works

The processor is intentionally declaration-level only:

1. It scans standard Java elements with `javax.lang.model.*`.
2. It checks field types, method parameters, return types, supertypes,
   thrown types, and generic type usage.
3. If any referenced type or method is annotated `@CopilotExperimental`,
   compilation fails unless `-Acopilot.experimental.allowed=true` is set.

This keeps the implementation portable across Java compilers while still
covering the main ways the SDK’s generated experimental APIs are used.

## Important review points

1. The processor no longer depends on compiler-specific AST APIs.
2. The module descriptor stays simple; no extra compiler-module wiring is required.
3. The README explicitly documents what is and is not caught, so the tradeoff is visible to users.
4. The generator still preserves existing experimental `@apiNote` Javadocs; the annotation is additive.
5. The tests prove both failure-by-default and success-with-opt-in behavior using in-memory compilation.

## Test coverage

- Fails when an experimental type appears in declarations
- Fails when a class extends an experimental type
- Passes when the compiler option `-Acopilot.experimental.allowed=true` is present

## Outcome

Experimental APIs are now opt-in at compile time, with a portable implementation and explicit documentation of the remaining limitations.


java/src/main/java/com/github/copilot/CopilotExperimentalProcessor.java
```diff
@@ -0,0 +1,158 @@
...
+        // If the enclosing type is experimental, members are implicitly experimental
+        Element enclosing = element.getEnclosingElement();
+        return enclosing != null && enclosing.getAnnotation(CopilotExperimental.class) != null;
+    }
+
+    private void reportError(Element experimentalElement, Element usageSite, String context) {
+        Messager messager = processingEnv.getMessager();
+        messager.printMessage(Diagnostic.Kind.ERROR,
+                "Use of experimental API '" + experimentalElement.getSimpleName() + "' in " + context
+                        + " is not allowed. Add compiler option" + " -Acopilot.experimental.allowed=true to opt in.",
```
johnoliver: extra concatenation

edburns: Good eye. Fixed in 7054c6e5 .

JonathanGiles (commented):
One thing I'll note is that we played around with similar concepts in the past and they didn't buy us as much as we hoped. This isn't to say it's a bad idea, it's just that we found most customers would be oblivious to both the annotation (because they don't read docs as much as we like), nor do they run the tools we would like them to run. Having said that, if you can generate this and apply it with ease, I think it is better to have it rather than not - but it won't protect you from customers being furious when something you've marked as experimental is ripped away from them in a new release (see our lived experience of com.sun.* for reference!)

The only thing I note (without fully knowing your code base) is that it seems odd to me that the annotation lives in the root package.

edburns:
@JonathanGiles wrote:

> The only thing I note (without fully knowing your code base) is that it seems odd to me that the annotation lives in the root package.

There are very few packages in this API:

image

I tend to oppose creating a new `annotations` package just for these things.

Is that ok?

JonathanGiles (commented):
✅ Looks good to me.

stephentoub (approved):
Thanks!

PR #1604, opened 2026-06-09
Handle session.canvas.closed by removing from open_canvases snapshot

jmoseley:
## What

Completes the symmetric pair with `session.canvas.opened`: the SDKs now **remove** a canvas from their in-memory `open_canvases` snapshot when the runtime emits `session.canvas.closed`.

## Why

The runtime now emits a symmetric `session.canvas.closed` ephemeral event (payload `{ instanceId, extensionId, canvasId }`, `instanceId` required + non-empty) when a canvas is genuinely removed — mirroring the existing `session.canvas.opened` (copilot-agent-runtime #9489, shipped in **CLI 1.0.60**).

The generated event types already landed via codegen, but the hand-written handlers only ever **upserted** on open, so the snapshot was append-only and never shrank on close. This left consumers' open-canvas state stale.

The consumer driving this is **github-app's "sticky terminal canvas" fix**: a right-panel canvas the user closes resurrects on warm session-reselect. That fix needs the SDK snapshot to shrink on close so reselect stops resurrecting closed canvases.

## How

Adds a symmetric removal handler in every SDK that maintains the open-canvases snapshot — **Rust, Node, Python, Go, .NET**. Each:

- removes the matching `instanceId` on `session.canvas.closed`, acquiring the **same lock** the upsert path uses;
- **warns and no-ops** on an empty/missing `instanceId`;
- is **idempotent** (closing an absent `instanceId` is a no-op);
- leaves the **resume path untouched** and does **not** treat opened/stale events as removals (only the explicit closed event removes — provider unregister still marks instances *stale* via a re-emitted `opened`, which continues to upsert).

Generated files were not hand-edited; the change is purely in the hand-written session handlers + unit tests.

### Java — intentionally excluded

Java has **no open-canvases snapshot feature** in `java/src/main` at all (no accessor, no opened-upsert handler, no resume capture) — only the generated event types exist. There is nothing symmetric to add; covering it would mean building the entire feature net-new, which is out of scope for this focused bug fix.

## Tests

New unit tests in each language open two canvases, close one (assert the closed instance is gone and the other remains), and cover idempotency (closing an absent id) and the empty-`instanceId` no-op.

- Rust: `cargo test --features test-support --test session_test` — pass
- Node: `vitest run test/client.test.ts` — 95 pass
- Python: `pytest test_canvas.py` — 12 pass
- Go: `go test .` — pass
- .NET: `dotnet test --filter CanvasTests` — 10 pass

### DIFF
--- a/dotnet/src/Session.cs
+++ b/dotnet/src/Session.cs
@@ -132,7 +132,7 @@ public sealed partial class CopilotSession : IAsyncDisposable
     /// </summary>
     /// <remarks>
     /// Populated from the most recent <c>session.resume</c> response and live
-    /// <c>session.canvas.opened</c> events.
+    /// <c>session.canvas.opened</c> and <c>session.canvas.closed</c> events.
     /// </remarks>
     [Experimental(Diagnostics.Experimental)]
     public IReadOnlyList<OpenCanvasInstance> OpenCanvases => _openCanvases;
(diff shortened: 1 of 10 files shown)


stephentoub (approved):
Did we not add these APIs to Java?

jmoseley:
> Did we not add these APIs to Java?

When they were originally added, the Java implementation was missing a bunch of stuff. I can kick off a session to build this into the Java client as well.

PR #1623, opened 2026-06-10
Java codegen: clean output directory before generating to prevent orphan accumulation

### LINKED ISSUE #1619: Java codegen leaves stale orphaned generated files (codegen does not clean its output dir)
## Summary

The Java code generator (`java/scripts/codegen/java.ts`) only creates directories and writes files — it never removes files it no longer emits. As a result, whenever an upstream schema type is renamed or removed, the previously-generated `.java` file is orphaned and lingers in `java/src/generated/java/`. These orphans have accumulated across many `@github/copilot` releases.

This was noticed while landing the 1.0.61 bump (#1612): the SDK Consistency Review bot flagged 3 files orphaned by the `InstructionsSources` → `InstructionSource` rename (removed in #1612). Investigating further revealed a much larger pre-existing backlog.

## Data (measured against the 1.0.61 schema)

Regenerating from a clean output directory and diffing against the tracked tree:

| Metric | Count |
|---|---|
| Tracked generated files | 1023 |
| Files current codegen actually emits | 819 |
| **Stale files codegen no longer emits** | **~204** |
| └─ no live (non-orphan) references found | ~172 |
| └─ share a name with identifiers in live main/test code | ~23 |

These ~204 files are **not** specific to 1.0.61 — they exist identically on `main` and accumulated over time.

## Root cause

`writeGeneratedFile()` in `java/scripts/codegen/java.ts` does `fs.mkdir(..., { recursive: true })` + `fs.writeFile(...)` and nothing else. There is no step that empties the output directory (or prunes files not produced by the current run). The CI check `Verify Java generated files are up-to-date` (`.github/workflows/java-codegen-check.yml`) runs codegen over the existing tree and then `git status`, so it only detects *content* drift in emitted files — it cannot detect orphans, because codegen never deletes them.

## Recommended fix

1. Make the codegen clean its output package before writing (e.g. remove `java/src/generated/java/com/github/copilot/generated/**` at the start of a run, then regenerate), so orphans cannot accumulate.
2. Regenerate and commit the resulting deletions in a dedicated cleanup PR.
3. (Optional hardening) Update the codegen-check workflow to clean before regenerating so future orphans are caught.

## Caveat — verify the ~23 "still referenced" types before deleting

A handful of the stale files share a type name with identifiers used by live code (e.g. `SessionMetadata`, `PermissionRequest`, `PermissionRule`, and several `Session*Result`/`Session*Params` records referenced by the generated RPC coverage tests). These need per-type verification: some may be genuine load-bearing dependencies (handwritten code relying on a type the current generator no longer emits, which itself needs reconciliation), while others may simply be name collisions with handwritten types in a different package. Deleting them blindly could break compilation, so the cleanup should be done by cleaning + regenerating via codegen (not by hand) and then resolving any resulting compile errors.

## Scope note

This is intentionally **not** being folded into the 1.0.61 bump PR (#1612), which is a green/approved dependency update. The 3 rename-orphans removed there were directly tied to that release; this broader cleanup is independent pre-existing tech debt and deserves its own PR.

Generated by Copilot

edburns:
## Summary

Add a clean step to `java/scripts/codegen/java.ts` that removes the entire generated output directory before writing new files. This ensures that when upstream schema types are renamed or removed, the previously-generated `.java` files are automatically deleted rather than silently accumulating as orphans.

## Changes

- **`java/scripts/codegen/java.ts`** — At the start of `main()`, recursively delete `src/generated/java/com/github/copilot/generated/` and recreate the empty directory before invoking any generators.

- **`java/src/generated/java/`** — 204 orphan files removed (types no longer emitted by the current schema). These accumulated over multiple `@github/copilot` releases as types were renamed or removed upstream.

- **`GeneratedRpcRecordsCoverageTest.java`** — Removed test methods that exercised deleted generated types (`SessionWorkspaceCreateFileParams`, `SessionModeGetResult`, `SessionModeSetResult`, `DiscoveredMcpServerSource`, and 16 others).

- **`GeneratedRpcApiCoverageTest.java`** — Removed test method for the deleted `SessionAgentDeselectResult` type.

## Why this works going forward

The CI workflow (`java-codegen-check.yml`) runs codegen and then checks `git status --porcelain`. Because codegen now cleans before writing, any future orphans will appear as deletions in the diff and the check will fail — preventing orphan accumulation without any additional workflow changes.

Fixes #1619


edburns:
I'm very sorry about this mistake, this PR is ready for review @stephentoub .

stephentoub (approved):

PR #1633, opened 2026-06-11
Surface ModelBilling.tokenPrices on public SDK types

MackinnonBuck:
## Summary

`ModelBilling.tokenPrices` is present on the wire and in every SDK's *generated* protocol types, but the hand-curated public `ModelBilling` types exposed only `multiplier`. Consumers had to narrow through local augmentations to read tier pricing (e.g. to detect a distinct long-context tier and label tiers like 200K / 1M in a model picker).

This adds `tokenPrices` to the curated public `ModelBilling` across all SDKs, mirroring the shape already published in the generated schema.

## Shape

- `ModelBillingTokenPrices`: `inputPrice`, `outputPrice`, `cachePrice` (numbers), `batchSize`, `contextMax` (ints), `longContext`
- `ModelBillingTokenPricesLongContext`: `inputPrice`, `outputPrice`, `cachePrice` (numbers), `contextMax` (int)

All fields optional; JSDoc/descriptions copied from the runtime schema.

## Per-SDK changes

| SDK | Change |
| --- | --- |
| Node | `types.ts`: curated interfaces + `tokenPrices?`; exported from `index.ts` |
| Python | `client.py`: dataclasses with `from_dict`/`to_dict`, wired into `ModelBilling`; exported in `__init__.py` |
| Go | `types.go`: structs + `TokenPrices` pointer field |
| .NET | `Types.cs`: sealed classes + property + `[JsonSerializable]` registrations |
| Java | new `ModelBillingTokenPrices.java` / `...LongContext.java` + field on `ModelBilling.java` |
| Rust | `types.rs`: re-export the two nested generated types (generated `ModelBilling` already carries `token_prices`) |

No generated files were modified — they already carry the field. No runtime changes required.

## Tests

Added coverage exercising the new field in every SDK (round-trip (de)serialization for Python/Go/.NET, deserialization assertions for Java, type + passthrough for Node, and re-export verification for Rust). All passing.


python/copilot/client.py
```diff
@@ -667,24 +667,131 @@ def to_dict(self) -> dict:
...
+        output_price = obj.get("outputPrice")
+        cache_price = obj.get("cachePrice")
+        batch_size = obj.get("batchSize")
+        context_max = obj.get("contextMax")
+        long_context_dict = obj.get("longContext")
+        long_context = (
+            ModelBillingTokenPricesLongContext.from_dict(long_context_dict)
+            if long_context_dict
+            else None
+        )
```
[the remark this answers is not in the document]
MackinnonBuck: Accepted. `longContext` now uses a presence check (`if long_context_dict is not None`) so a present-but-empty `{}` deserializes to a `ModelBillingTokenPricesLongContext` with all fields None instead of collapsing to None. Added a round-trip test covering the empty-object case. Fixed in a2732bb2.

python/copilot/client.py
```diff
@@ -667,24 +667,131 @@ def to_dict(self) -> dict:
...
     def from_dict(obj: Any) -> ModelBilling:
         assert isinstance(obj, dict)
         multiplier = obj.get("multiplier")
-        if multiplier is None:
-            return ModelBilling()
-        return ModelBilling(multiplier=float(multiplier))
+        token_prices_dict = obj.get("tokenPrices")
+        token_prices = (
+            ModelBillingTokenPrices.from_dict(token_prices_dict) if token_prices_dict else None
+        )
```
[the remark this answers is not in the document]
MackinnonBuck: Accepted. `ModelBilling.from_dict` now uses a presence check (`if token_prices_dict is not None`) so a present-but-empty `tokenPrices` (`{}`) deserializes to a `ModelBillingTokenPrices` instance with all fields None, letting callers tell "present but empty" from "not present". Added a round-trip test. Fixed in a2732bb2.

Note: I scoped this to the new `tokenPrices`/`longContext` fields and intentionally left the pre-existing `billing` truthiness check in `ModelInfo.from_dict` unchanged to keep the diff minimal.

java/src/main/java/com/github/copilot/rpc/ModelBilling.java
```diff
@@ -18,6 +18,9 @@ public class ModelBilling {
     @JsonProperty("multiplier")
     private double multiplier;
 
+    @JsonProperty("tokenPrices")
+    private ModelBillingTokenPrices tokenPrices;
+
```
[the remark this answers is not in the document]
MackinnonBuck: Accepted. Changed `multiplier` from primitive `double` to boxed `Double` (getter and setter updated to match), so "not present on the wire" is distinguishable from an explicit 0.0. This also brings Java in line with the other SDKs, which all model multiplier as optional (Go `*float64`, C# `double?`, Python `float | None`, TS `number?`, Rust `Option<f64>`). The SDK's shared ObjectMapper is configured with NON_NULL inclusion, so a null multiplier is omitted on serialization; I added `testModelBillingSerializationOmitsNullMultiplier` to lock that in. Fixed in a2732bb2.

SteveSandersonMS (approved):
Looks great.

@edburns Can you give an opinion on the Java breaking change (double -> Double) here? Will this be acceptable in a patch release or do we need to find an alternative?

dotnet/src/Types.cs
```diff
@@ -3308,6 +3308,70 @@ public sealed class ModelBilling
...
+    /// <summary>Number of tokens per standard billing batch.</summary>
+    [JsonPropertyName("batchSize")]
+    public int? BatchSize { get; set; }
+
+    /// <summary>
+    /// Prompt token budget (max_prompt_tokens) for the default tier. The total
+    /// context window is this value plus the model's max_output_tokens.
+    /// </summary>
+    [JsonPropertyName("contextMax")]
+    public int? ContextMax { get; set; }
```
stephentoub: Is this the same naming we use elsewhere for this concept? I'd have expected MaxContextWindow or something like that.

dotnet/src/Types.cs
```diff
@@ -3308,6 +3308,70 @@ public sealed class ModelBilling
     /// </summary>
     [JsonPropertyName("multiplier")]
     public double? Multiplier { get; set; }
+
+    /// <summary>
+    /// Token-level pricing information for this model.
+    /// </summary>
+    [JsonPropertyName("tokenPrices")]
+    public ModelBillingTokenPrices? TokenPrices { get; set; }
```
stephentoub, line 3316: Do we actually need new types for this, or could / should we just return the Rpc type?

In general, I'm wondering about these "higher-level" APIs that are effectively identical to the RPC ones.

edburns:
[1633-java-preserve-backward-compatibility.md](https://github.com/user-attachments/files/29014291/1633-java-preserve-backward-compatibility.md)

Hi @MackinnonBuck here is a prompt that should preserve your intent, but maintain backward compatibility. Try it out and let me know how it goes.

If it puts up a fight, let me know and I can make a branch of `mackinnonbuck/expose-modelbilling-tokenprices` with the change.

dotnet/src/Types.cs
MackinnonBuck: Just deleted this type, so now the existing, stable, generated RPC type is used instead. It has the same property name (`ContextMax`).

We could consider renaming this on the runtime side, but this would be a breaking change for the Rust SDK, which already exposes it. Alternatively, I could add back the wrapper types and rename the property that way, although this would then introduce inconsistency between the Rust SDK and the other langs.

We could also mark this property experimental, even though the underlying type already isn't, so we could revisit this in the future if we wanted.

dotnet/src/Types.cs
MackinnonBuck: Agreed, we can remove the new types. I've just done so.

dotnet/src/Types.cs
MackinnonBuck: I think I'm going to keep the naming as-is. This type was shipping publicly prior to this change, and already exposed via the Rust SDK.

PR #1635, opened 2026-06-11
Harden bundled Copilot CLI extraction against corrupt/partial installs

jmoseley:
## Summary

On a Windows user's laptop, the GitHub Copilot desktop app's bundled `gh.exe` failed to launch with **`ERROR_BAD_EXE_FORMAT`** ("The specified executable is not a valid application for this OS platform"). Investigation traced the class of bug to non-atomic binary extraction that returns an unverified file. This PR applies the equivalent fix for the **Copilot CLI binary that `copilot-sdk` bundles/extracts** (the `bundled-cli` path in the Rust crate).

> This is the "CLI in copilot-sdk" half of Tim Clem's ask. The "`gh` in the app" half is a separate PR in `github/github-app`. The safe-publish pattern here mirrors `src-tauri/src/embedded_git.rs` (stage in same-dir temp → write integrity marker → atomic `fs::rename`, with explicit Windows non-atomic-overwrite handling).

## Root cause

`rust/src/embeddedcli.rs` extracted the embedded archive and wrote the binary with a non-atomic **open → truncate → write** directly on the live final path, then returned **without re-verifying the written bytes**. The "already installed?" fast path trusted a bare `is_file()` check. So an interrupted write, a multi-process race, or Windows Defender / ASR quarantining the freshly-written executable could leave a truncated or corrupt image that gets handed back as "good" — and fails to launch.

## Fix

- **Atomic write** — extract to a unique temp file (`pid` + counter + nanos) in the **same** target directory, `write_all` → `flush` → `sync_all`, set `0o755` on unix, then `fs::rename` onto the final path. `rename` overwrites atomically on POSIX; on Windows (which won't atomically overwrite) we remove-then-rename, with the race window guarded by re-verification and by accepting a peer's identical install.
- **Verify after publish** — the bytes extracted from the trusted, signed embedded archive are the known-good reference. We verify the staged temp file *and* re-verify the published file byte-for-byte (size checked first), catching truncation or AV tampering between staging and publishing.
- **Trustworthy fast path** — an existing install is reused only after a cheap re-check: an integrity marker recording the expected size **plus** a platform executable-image header check (PE `MZ` / Mach-O / ELF). A truncated, empty, or quarantined binary is re-extracted instead of trusted.
- **Retry + actionable fallback** — re-extract/re-publish up to 3 times to ride out transient AV interference; if it still fails, surface a clear error (*"bundled CLI appears blocked or corrupt … possibly quarantined by antivirus"*) instead of returning a broken path.

The change is scoped to the CLI extraction/install path. Public behavior is unchanged for the happy path; the installer is just no longer able to return a broken executable.

## Tests

New unit tests in `embeddedcli.rs` cover:
- temp + atomic-rename publish writes the exact bytes, records the marker, and leaves no temp files;
- publish overwrites a stale existing binary;
- corrupt / truncated / empty / unmarked / invalid-image installs are detected and rejected by the fast-path re-check;
- post-write verification rejects both size and content mismatches.

The existing `install_bundled_cli_*` integration tests (which exercise the full extract → publish → verify → marker flow against the real downloaded archive and assert idempotency) continue to pass.

## Validation

- `cargo +nightly fmt --check` ✅
- `cargo clippy --all-targets --features test-support -- -D warnings -D clippy::unwrap_used …` ✅ (both the skip-download and real-bundled paths)
- `cargo doc --no-deps --all-features` with `-D warnings` ✅
- `cargo build` (bundled, downloads real CLI) ✅
- `cargo test --lib embeddedcli` + `install_bundled_cli` integration tests ✅

### DIFF
--- a/rust/src/embeddedcli.rs
+++ b/rust/src/embeddedcli.rs
@@ -9,17 +9,36 @@
 //! **raw archive bytes**
 //! into the consumer's compiled artifact via `include_bytes!()`. Extraction
 //! to a real on-disk path is deferred until the first call to
-//! [`path`] / [`install_at`] — at which point the bytes are part of the
-//! consumer's signed binary and trusted, so no further hashing is done.
+//! [`path`] / [`install_at`].
+//!
+//! The embedded bytes are part of the consumer's signed binary and therefore
+//! trusted *as the source of truth* — but the bytes that land on disk are not.
+//! A non-atomic write, a multi-process race, or antivirus quarantining the
+//! freshly-written executable can leave a truncated or corrupt image that, if
+//! handed back as "good", fails to launch (e.g. Windows `ERROR_BAD_EXE_FORMAT`).
+//! Installation therefore: extracts to a unique temp file in the target dir,
+//! fsyncs and marks it executable, verifies the staged bytes against the
+//! trusted in-memory image, atomically renames it into place, re-verifies the
+//! published file, and records an integrity marker. Subsequent runs trust an
+//! existing install only after a cheap re-check (size marker + executable-image
+//! header); anything that looks truncated or quarantined is re-extracted, and
+//! the whole publish is retried before surfacing a clear, actionable error.
 
-#[cfg(has_bundled_cli)]
+// The atomic-publish + verify helpers (and their unit tests) are pure
+// std-only logic that doesn't touch the embedded archive, so they compile
+// whenever the binary is bundled *or* we're building the test harness —
+// the standard `cargo test --no-default-features` job has `has_bundled_cli`
+// off but still needs to exercise them.
+#[cfg(any(has_bundled_cli, test))]
 use std::fs;
 #[cfg(all(has_bundled_cli, not(windows)))]
 use std::io::Read;
-#[cfg(has_bundled_cli)]
+#[cfg(any(has_bundled_cli, test))]
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::sync::OnceLock;
+#[cfg(any(has_bundled_cli, test))]
+use std::sync::atomic::{AtomicU64, Ordering};
 
 #[cfg(has_bundled_cli)]
 use tracing::{info, warn};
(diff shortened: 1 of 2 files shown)


rust/src/embeddedcli.rs
```diff
@@ -9,17 +9,36 @@
...
 //! into the consumer's compiled artifact via `include_bytes!()`. Extraction
 //! to a real on-disk path is deferred until the first call to
-//! [`path`] / [`install_at`] — at which point the bytes are part of the
-//! consumer's signed binary and trusted, so no further hashing is done.
+//! [`path`] / [`install_at`].
+//!
+//! The embedded bytes are part of the consumer's signed binary and therefore
+//! trusted *as the source of truth* — but the bytes that land on disk are not.
+//! A non-atomic write, a multi-process race, or antivirus quarantining the
+//! freshly-written executable can leave a truncated or corrupt image that, if
```
SteveSandersonMS, line 17: I might be wrong but I thought AV quarantining was generally done by setting some other flag in filesystem metadata, not modifying the bytes of the executable. Maybe some AV software differs.

Doesn't change the validity of the PR though

rust/src/embeddedcli.rs
```diff
@@ -136,29 +165,277 @@ fn install(install_dir: &Path, archive: &[u8]) -> Result<PathBuf, EmbeddedCliErr
...
+        ));
+    }
+    Ok(())
+}
+
+/// Best-effort check that the first bytes of `path` are a valid executable
+/// image header for the current platform (PE on Windows, Mach-O on macOS,
+/// ELF elsewhere). Returns `false` on any I/O error or unrecognized header.
+#[cfg(any(has_bundled_cli, test))]
+fn looks_like_valid_image(path: &Path) -> bool {
```
SteveSandersonMS, line 389: If we wanted a more end-to-end problem detector we could consider something like detecting whether the `spawn` fails (that is, we never get to the point of having a valid handshake) and if so track a flag so that next time the app is launched, we do a full re-extraction. Then we wouldn't need so many other checks.

rust/src/embeddedcli.rs
SteveSandersonMS: Update: looks like Windows Defender will actually move the file into a different dir if it quarantines it.

SteveSandersonMS (approved):

stephentoub (approved):

PR #1669, opened 2026-06-14
Add SDK MCP OAuth host token handlers

roji:
## Summary
- Add SDK-facing MCP OAuth host-token callbacks for Node.js, Python, Go, .NET, Rust, and Java.
- Register `mcp.oauth_required` interest when a host auth handler is configured, map runtime OAuth request events into idiomatic per-language handler contexts, and respond via `session.mcp.oauth.handlePendingRequest` with token or cancellation results.
- Update generated/schema-derived surfaces for MCP OAuth request reasons, parsed WWW-Authenticate params, resource metadata, static client config including `clientSecret`, and OAuth completion events.
- Add a shared OAuth-protected MCP test server plus cross-language unit/E2E coverage for initial auth, refresh/replacement tokens, upscope, reauth, cancellation, handler context, and interest registration.

## Validation
- Node MCP OAuth E2Es and type/format checks
- Python MCP OAuth E2Es and type/format checks
- Go MCP OAuth unit/E2E tests
- .NET MCP OAuth unit/E2E tests
- Rust MCP OAuth unit/E2E tests
- Java MCP auth interest registration and MCP OAuth E2Es

## Notes
- Tyler's `idToken` review thread is intentionally left unresolved for follow-up discussion.


nodejs/src/types.ts
```diff
@@ -1554,6 +1545,74 @@ export type ReasoningEffort = "low" | "medium" | "high" | "xhigh";
...
+}
+
+/** Static OAuth client configuration supplied by the MCP server, if available. */
+export interface McpAuthStaticClientConfig {
+    /** OAuth client ID for the server. */
+    clientId: string;
+    /** Optional non-default OAuth grant type. */
+    grantType?: "client_credentials";
+    /** Whether this is a public OAuth client. */
+    publicClient?: boolean;
```
TylerLeonhardt, line 1637: what about the clientSecret? A lot of AS's use client secrets even for public clients. (VS Code allows specifying a client secret, stored in secret storage)

nodejs/src/types.ts
```diff
@@ -1554,6 +1545,74 @@ export type ReasoningEffort = "low" | "medium" | "high" | "xhigh";
...
+export interface McpAuthToken {
+    /** Access token acquired by the SDK host. */
+    accessToken: string;
+    /** OAuth token type. Defaults to Bearer when omitted. */
+    tokenType?: string;
+    /** Refresh token supplied by the host, if available. */
+    refreshToken?: string;
+    /** Token lifetime in seconds, if known. */
+    expiresIn?: number;
+}
```
TylerLeonhardt, line 1666: what about an idToken? idTokens are useful as they contain claims that are good to display like usernames/email/etc.

roji: So unless I'm mistaken, the runtime wouldn't have any actual use for it; all it needs as a response from the host is the access token to be forwarded back to the MCP server for access... In fact, I'm about to remove the refreshToken from here, since as discussed in yesterday's meeting, refresh would be the host's responsibility, and so the runtime has no use for that.

But let me know if I have the wrong mental model here! Plus we can always add it later.

nodejs/src/types.ts
roji: Makes sense, will add.

dotnet/src/Client.cs
```diff
@@ -1166,6 +1172,10 @@ public async Task<CopilotSession> ResumeSessionAsync(string sessionId, ResumeSes
             transformCallbacks,
             hasHooks,
             "CopilotClient.ResumeSessionAsync");
+        if (config.OnMcpAuthRequest is not null)
+        {
+            await session.Rpc.EventLog.RegisterInterestAsync("mcp.oauth_required", cancellationToken);
+        }
```
stephentoub, line 1178: Are these necessarily tied together?

I'm thinking back to permission requests and tool invocation requests. Initially the only way you could handle them was to provide a handler that the SDK would invoke, but that was prohibitive, especially with regards to needing to be able to suspend the runtime and later resume it (possibly on a different machine) and supply the result of the operation in order to keep going. We made it so that providing the callback was optional; if you don't supply it, you're on your own for handling the events and calling the runtime rpc method to supply the results.

Is there a correlary here?

roji: Yeah, at least at the moment, register interest in the callback is necessary since the runtime has two MCP OAuth flows - the new one being introduced here and the in-process one currently used by the CLI; I'm working on a PR to switch the CLI to use the new SDK flow - as part of the CLI-over-SDK effort - but for now the interest registration is required in order for the runtime to know which path to take.

Regardless, even after we remove the CLI flow, I'm not sure how the "you're on your own" approach would work here MCP OAuth... The LLM can request a tool call at an arbitrary time, which could trigger an "oauth_required" request, which must be satisfied by the host; if this doesn't happen, the tool call can't proceed, and the turn is stuck. So I *think* it means the host must be informed (and it must respond in a timely manner)... I think that this oauth flow must indeed be rewired upon resume (even on a different machine), but it doesn't change the fact that an OAuth request can happen at any point, and the host must be able to handle it...

In other words, I think that any SDK host that potentially needs to support OAuth-authorizing MCP servers **must** register this callback, and if such a request is emitted without that callback being registered, I'm not sure what choice we have besides a fail-fast error... But I might have an incorrect mental model here - let me know how you see this.

stephentoub: To make sure I've understood correctly, all the SDK ends up doing is listening for a particular event, invoking the user's callback, and calling the exposed RPC method to supply the result, right? So SDK consumer could do the same thing without providing a callback: they could listen for the same event, do whatever they want in response (whatever they would have supplied in the callback), and then call that same RPC method, right?

The PR is using a callback being set as an indication that the client is handling oauth, but if the above is true, then those needn't be 1:1... supplying a callback could imply it, but not supplying a callback doesn't mean the client can't also handle it. My question is really: should there be a separate knob for opting-in to handling oauth beyond just supplying a callback.

It's fine if the answer is "not right now, we could always add one in the future if supplying a callback is problematic for some reason".

roji: **EDIT: Maybe read my comment below, I think I understood better what you were asking**

Here's how the flow works:

1. The runtime makes an MCP call, and receives a 401/403 HTTP error, signifying that a (new) OAuth access token is required. Crucially, the runtime is the one seeing this - the hosting application (or client) has no idea it's happening.
2. In the new host-delegating flow (when "interest is registered"), the runtime records the pending OAuth request and then invokes the callback in the hosting application, passing it various information from the 401 response (e.g. WWW-Authenticate header contents).
3. The hosting application does whatever is needed in order to obtain the access token (user browser flow, or possibly just a refresh cycle).
4. The hosting application then calls an API back into the SDK (handlePendingRequest)
5. Back in the runtime, we correlate the request ID given to handlePendingRequest with the recorded OAuth request from step 2, and then redoes the MCP request with the new auth token received from the host.

IIUC you're asking whether it makes sense to have a host opt into host-delegated OAuth *without* registering a callback. In that case, I can't see what would happen when the runtime receives a 401/403 HTTP error which is the trigger for OAuth - how would the host application be made aware that OAuth needs to happen? Like are you thinking of some sort of polling mechanism where the host application would periodically check if there are any pending OAuth requests? I guess that's theoretically possible, but would at the very least mean lots of latency no?

Are you asking because reinstating host callbacks with the runtime after resume is problematic? I looked at this and it seems that we have several other callbacks which IIRC are should also be reinstated after resume - but I could be wrong.

Happy to do a quick call tomorrow to discuss all this!

roji: OK, I reread everything and I think I understand your concern... You're saying that in the SDK implementation here (not the runtime), we expose a single callback which must return the token; this would prevents e.g. an implementation that persists the various parameters to disk, and then some other processes loads those and calls handlePendingRequest with the token. In other words, this is less about separating the opt-in from the callback, it's more to have a more low-level split interaction where there's the callback invoked by the runtime doesn't have to return an access token "immediately", and the host application later invokes handlePendingRequest manually.

I can see that... And that's exactly how the low-level RPC API is actually structured (oauth_required is the low-level callback that returns nothing). If so, then I *think* all this should be achievable with the low-level APIs today:

```c#
await session.Rpc.EventLog.RegisterInterestAsync("mcp.oauth_required");

using var sub = session.On<McpOauthRequiredEvent>(evt =>
{
    // Persist requestId + auth params.
    // Do not return token; this is just an event callback.
});

// Later / elsewhere:
await session.Rpc.Mcp.Oauth.HandlePendingRequestAsync(requestId, tokenResult);
```

There may be some issues around registering interest (and the callback) early enough during resume so as to avoid any race conditions... Also, specifically for the multiple process scenario (one process receives oauth_required, persists, another process actually completes with the access token), the runtime itself needs to remember/correlate the pending OAuth requests, and that's in-memory only; so a true resume will have lost that (assuming the runtime is in the same process as the host application).

tl;dr I think we have a nice high-level convenience callback for the common scenario (register a single async callback that returns the access token). In principle, users can still drop down to low-level RPC APIs to do the advanced (cross-process) thing, but it's likely we have some gaps around this - but they go beyond simple API shape here in the SDK, and we should probably think/design that separately (the main question is whether the high-level API as-is seems OK even if the advanced scenario isn't there yet).

Does this all make sense and correspond to what you're asking?

stephentoub: > correspond to what you're asking?

Yup! Being able to write that code sample is what I was asking about.

stephentoub (approved):

dotnet/src/Client.cs
roji: OK great, understood.

Though as above, I suspect that for actual cross-process resume there'll still be some gaps/blockers... We can deal with those later.

PR #1693, opened 2026-06-16
Add per-call tool context provider to Python SDK

idryzhov:
Introduce an optional `on_provide_tool_context` callback on `create_session`/`resume_session`. When registered, the SDK invokes it once per tool call with the `ToolInvocation` (awaiting coroutine results) and assigns the return value to the new `ToolInvocation.context` field before the handler runs. This lets apps inject per-call application context (services, request state) into custom tool handlers without routing it through model-facing tool arguments. The context is filled in locally and is never serialized over the wire.


brettcannon:
Any reason you can't achieve the same thing in userland code by wrapping your tool call handlers?

idryzhov:
> Any reason you can't achieve the same thing in userland code by wrapping your tool call handlers?

@brettcannon that's exactly what we're doing right now. But I thought the use-case should be common enough to support it in the SDK - and it would allow us to have less custom code :)

SteveSandersonMS:
Thanks @idryzhov. We decided not to proceed with this since there haven't been any other requests for it, and because as you say it's achievable in app code already.

PR #1701, opened 2026-06-17
feat(nodejs): plumb AbortSignal through ToolInvocation (#1433)

### LINKED ISSUE #1433: Feature request: Plumb AbortSignal through ToolInvocation so session.abort() can cancel in-flight tool handlers
# Feature request: Plumb AbortSignal through ToolInvocation so session.abort() can cancel in-flight tool handlers

## Summary

`session.abort()` cancels the agentic loop (no further tool calls scheduled) but does not propagate cancellation into a currently-executing tool handler. The handler runs to completion regardless. This forces SDK consumers building wedge-recovery to implement OS-level process tree kills at their own layer, when the cooperative cancellation primitive could live in the SDK.

## Verified state at `@github/copilot-sdk` (`dist/`)

**1. `ToolInvocation` interface (`types.d.ts:224-233`):**
```ts
export interface ToolInvocation {
    sessionId: string;
    toolCallId: string;
    toolName: string;
    arguments: unknown;
    /** W3C Trace Context traceparent from the CLI's execute_tool span. */
    traceparent?: string;
    /** W3C Trace Context tracestate from the CLI's execute_tool span. */
    tracestate?: string;
}
```
No `signal`, no `abortSignal`, no cancellation handle.

**2. `_executeToolAndRespond` (`session.js:278-309`):**
```ts
async _executeToolAndRespond(requestId, toolName, toolCallId, args, handler, traceparent, tracestate) {
  try {
    const rawResult = await handler(args, { sessionId, toolCallId, toolName, arguments: args, traceparent, tracestate });
    ...
  } catch (error) { ... }
}
```
Handler is awaited synchronously; no signal threaded through; no race against an abort source.

**3. `session.abort()` (`session.js:765-769`):**
```ts
async abort() {
  await this.connection.sendRequest("session.abort", { sessionId: this.sessionId });
}
```
JSDoc: "and can continue to be used for new messages" + "resolves when the abort request is **acknowledged**." Confirms scope is the agentic loop, not the handler.

**4. Searched `node_modules/@github/copilot-sdk/dist/` for `AbortSignal|AbortController|signal:|cancel\(|cancellation`:** zero matches.

## Proposed API

Add an optional `AbortSignal` to `ToolInvocation` that aborts when `session.abort()` (or a new `session.cancelToolCall(toolCallId)`) is called:

```ts
export interface ToolInvocation {
    sessionId: string;
    toolCallId: string;
    toolName: string;
    arguments: unknown;
    /** Aborts when session.abort() or session.cancelToolCall(toolCallId) is invoked. */
    signal: AbortSignal;
    traceparent?: string;
    tracestate?: string;
}
```

Optional addition: a more granular `session.cancelToolCall(toolCallId)` that cancels a specific in-flight handler without aborting the broader agentic loop.

## Why it matters

Without handler-level cancellation, consumers building wedge-recovery (long-running shell handlers, network calls, file I/O on large files) must:
1. Track child PIDs at handler-spawn or via shell tool event streams,
2. Issue OS-level kills (Windows `taskkill /F /T /PID`, POSIX `SIGKILL` fan-out across descendant tree),
3. Re-probe survivors with bounded wait,
4. Manage orphan-leak edge cases.

That's ~125 lines of code per consumer plus cross-platform paths. A cooperative `AbortSignal` in the SDK lets handlers self-terminate cleanly — `fetch(url, { signal })`, `await sleep(ms, { signal })`, `child_process.spawn(...).on(signal.aborted, kill)` — which is the well-trodden Node.js cancellation idiom.

## Backwards compatibility

`AbortSignal` is non-required for handlers that don't consume it (existing handlers continue to work). Handlers that opt in get cooperative cancellation; the SDK keeps its existing process-isolation guarantees.

## Concrete consumer evidence

Parley (multi-agent orchestration system built on `@github/copilot-sdk`) ships at present a 313-line wedge-kill primitive in our orchestrator (Windows `taskkill /F /T /PID` + POSIX SIGKILL fan-out + bounded post-kill verify + tracker reset) that exists specifically because `session.abort()` doesn't propagate. With this feature, that primitive would be ~30 lines of `if (signal.aborted) cleanup()` in the consumer's tool handlers.

## Workaround until landed

OS-level process tree kill via tracked child PIDs from tool start events, gated by a bounded post-kill verify. Functional but not durable. Issue tracks the durable layer.

## Affected versions

Verified against `@github/copilot-sdk` shipped in current Parley `node_modules` snapshot (May 2026). Please confirm whether the proposed API has been considered or is on the roadmap.

gimenete:
## Summary

Implements the feature requested in #1433 for the **Node.js SDK**: plumb an `AbortSignal` through `ToolInvocation` so that `session.abort()` (and a new `session.cancelToolCall(toolCallId)`) can cooperatively cancel in-flight tool handlers.

Previously `session.abort()` only cancelled the agentic loop — a currently-executing tool handler ran to completion regardless, forcing consumers to implement OS-level process-tree kills.

## Changes

- **`ToolInvocation`** now carries a required `signal: AbortSignal` that aborts when `session.abort()` or `session.cancelToolCall(toolCallId)` is invoked while the handler is in flight.
- **`session.abort()`** aborts the signals of all in-flight tool handlers before sending the abort RPC.
- **`session.cancelToolCall(toolCallId)`** (new) cancels a single in-flight handler without aborting the broader agentic loop; returns `true` if a matching call was found.
- **`session.disconnect()`** aborts in-flight handlers so they can release resources.
- In-flight handlers are tracked in a `Map<toolCallId, AbortController>`.

## Backwards compatibility

Handlers that ignore the signal continue to run to completion — existing handlers work unchanged.

## Tests

Extended the `abort` e2e test to assert the handler's `AbortSignal` fires when `session.abort()` is called. Typecheck, lint, format check, and unit tests all pass.

Closes #1433

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/nodejs/src/session.ts
+++ b/nodejs/src/session.ts
@@ -121,6 +121,7 @@ export class CopilotSession {
     private typedEventHandlers: Map<SessionEventType, Set<(event: SessionEvent) => void>> =
         new Map();
     private toolHandlers: Map<string, ToolHandler> = new Map();
+    private inFlightToolCalls: Map<string, AbortController> = new Map();
     private canvases: Map<string, Canvas> = new Map();
     private commandHandlers: Map<string, CommandHandler> = new Map();
     private permissionHandler?: PermissionHandler;
@@ -563,12 +564,15 @@ export class CopilotSession {
         traceparent?: string,
         tracestate?: string
     ): Promise<void> {
+        const abortController = new AbortController();
+        this.inFlightToolCalls.set(toolCallId, abortController);
         try {
             const rawResult = await handler(args, {
                 sessionId: this.sessionId,
                 toolCallId,
                 toolName,
                 arguments: args,
+                signal: abortController.signal,
                 traceparent,
                 tracestate,
             });
@@ -592,6 +596,12 @@ export class CopilotSession {
                     throw rpcError;
                 }
                 // Connection lost or RPC error — nothing we can do
+            }
+        } finally {
+            // Only clear if this is still the controller for this toolCallId;
+            // guards against a recycled toolCallId from a later invocation.
+            if (this.inFlightToolCalls.get(toolCallId) === abortController) {
+                this.inFlightToolCalls.delete(toolCallId);
             }
         }
     }
@@ -1170,6 +1180,9 @@ export class CopilotSession {
      * ```
      */
     async disconnect(): Promise<void> {
+        // Abort any in-flight tool handlers so they can release resources.
+        this._abortInFlightToolCalls();
+        this.inFlightToolCalls.clear();
         await this.connection.sendRequest("session.destroy", {
             sessionId: this.sessionId,
         });
@@ -1209,9 +1222,65 @@ export class CopilotSession {
      * ```
      */
     async abort(): Promise<void> {
+        // Cooperatively cancel any in-flight tool handlers that opted in to the
+        // AbortSignal exposed on their ToolInvocation. Handlers that ignore the
+        // signal continue to run to completion.
+        this._abortInFlightToolCalls();
         await this.connection.sendRequest("session.abort", {
             sessionId: this.sessionId,
         });
+    }
+
+    /**
+     * Cooperatively cancels a single in-flight tool handler by aborting the
+     * `AbortSignal` on its `ToolInvocation`, without aborting the broader
+     * agentic loop.
+     *
+     * This only affects handlers that opted in to the signal (e.g. by passing
+     * it to `fetch`, `child_process.spawn`, or checking `signal.aborted`).
+     * Handlers that ignore the signal continue to run to completion.
+     *
+     * @param toolCallId - The `toolCallId` of the in-flight tool invocation to cancel
+     * @returns `true` if a matching in-flight tool call was found and signaled, `false` otherwise
+     *
+     * @example
+     * ```typescript
+     * const session = await client.createSession({
+     *   tools: [
+     *     defineTool("fetch_data", {
+     *       handler: async (args, { signal }) => {
+     *         const res = await fetch(args.url, { signal });
+     *         return await res.text();
+     *       },
+     *     }),
+     *   ],
+     * });
+     *
+     * session.on((event) => {
+     *   if (event.type === "tool.execution_start") {
+     *     // Cancel a specific tool call after a deadline
+     *     setTimeout(() => session.cancelToolCall(event.data.toolCallId), 5000);
+     *   }
+     * });
+     * ```
+     */
+    cancelToolCall(toolCallId: string): boolean {
+        const controller = this.inFlightToolCalls.get(toolCallId);
+        if (!controller) {
+            return false;
+        }
+        controller.abort();
+        return true;
+    }
+
+    /**
+     * Aborts the AbortSignal for every in-flight tool handler.
+     * @internal
+     */
+    private _abortInFlightToolCalls(): void {
+        for (const controller of this.inFlightToolCalls.values()) {
+            controller.abort();
+        }
     }
 
     /**
--- a/nodejs/src/session.ts
+++ b/nodejs/src/session.ts
@@ -1269,6 +1269,9 @@ export class CopilotSession {
         if (!controller) {
             return false;
         }
+        // Remove the entry up front so a subsequent cancelToolCall (or the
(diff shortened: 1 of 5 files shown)


nodejs/src/types.ts
```diff
@@ -475,6 +475,15 @@ export interface ToolInvocation {
...
     arguments: unknown;
+    /**
+     * An `AbortSignal` that aborts when `session.abort()` or
+     * `session.cancelToolCall(toolCallId)` is invoked while this handler is
+     * in flight. Handlers may opt in to cooperative cancellation by forwarding
+     * it to abortable APIs (`fetch(url, { signal })`, `child_process.spawn`,
+     * etc.) or by checking `signal.aborted`. Handlers that ignore it continue
+     * to run to completion, preserving existing behavior.
+     */
+    signal: AbortSignal;
```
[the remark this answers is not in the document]
gimenete: Keeping `signal` required on purpose for cross-SDK consistency — the same field is non-optional in the Python, Go, .NET, Rust, and Java equivalents, and the SDK always injects a live signal at runtime, so handlers never have to null-check it. The only friction is for consumers who hand-construct/mock `ToolInvocation` in tests, which is a one-line addition (`signal: new AbortController().signal`). Given it's a new field on a handler-input type rather than a return type, the runtime guarantee felt worth the minor upgrade cost.

nodejs/src/session.ts
```diff
@@ -1209,11 +1222,70 @@ export class CopilotSession {
      * ```
      */
     async abort(): Promise<void> {
+        // Cooperatively cancel any in-flight tool handlers that opted in to the
+        // AbortSignal exposed on their ToolInvocation. Handlers that ignore the
+        // signal continue to run to completion.
+        this._abortInFlightToolCalls();
         await this.connection.sendRequest("session.abort", {
             sessionId: this.sessionId,
         });
```
SteveSandersonMS, line 1231: I think we need to be sending `session.abort` before calling `_abortInFlightToolCalls`.

If tools get aborted first, their `AbortSignal` will fire, and their promises will be rejected. This sends back a tool call result to the runtime. If the runtime receives this before the session abort message, it will treat it as an error and might even start a new agent invocation to process the failed tool call.

I know that as the code is structured right now, that doesn't actually happen because there are no awaits in between lines 1228 and 1229. But that means it's working just by chance, and could stop working if anything about this code is refactored in the future (e.g., awaiting something between those two lines).

SteveSandersonMS:
Thanks for these PRs, @gimenete.

The feature is a good idea, but the way it's implemented here doesn't look right to me. This assumes there's only one client, and that session abort is always triggered by that client. But in reality there can be multiple clients on the same session, and the tool calls might be running on a different client than the one that requests the abort, or the abort might be triggered by something inside the runtime rather than a client.

Instead, I think the right design is for clients to listen for the session `abort` event coming from the runtime, and they can react to that by cancelling their pending tool calls on that session. This would be much simpler to implement too - you wouldn't even need a separate `AbortController` for each tool call - they could all share the same one that simply signals if the session is aborted (and then is replaced by a different instance for subsequent tool calls).

I'd also prefer not to have the `cancelToolCall` API as it's not essential for this feature and apps can already put in their own logic to cancel their own tool calls. If we really wanted a cancelToolCall API it would have to work across clients and that's much more complicated and would require runtime support. Since we don't have a clear indication that people need a `cancelToolCall` (vs just writing their own abort logic in their own tool call handlers) I think we could just avoid that extra feature.

Does this make sense?

nodejs/src/session.ts
SteveSandersonMS: Update: this would cease to be an concern after updating the flow to match https://github.com/github/copilot-sdk/pull/1701#issuecomment-4790543921

SteveSandersonMS:
Thanks for putting together this coordinated cross-language work. The underlying goal—cooperatively cancelling in-flight tool handlers—is still valid, but these PRs share a design issue: cancellation only occurs when the same client calls `abort()`, so runtime-initiated or other-client aborts are missed. The proposed per-tool cancellation API is likewise local-only despite appearing session-wide.

We need to redesign this around the runtime's authoritative session abort event and reconsider targeted cancellation unless runtime support is added. Since these branches are also now stale, I'm closing them rather than asking for incremental updates. The useful cancellation primitives can be reused in a fresh coordinated implementation.

PR #1705, opened 2026-06-17
go: plumb context cancellation through ToolInvocation for session.Abort()

### LINKED ISSUE #1433: Feature request: Plumb AbortSignal through ToolInvocation so session.abort() can cancel in-flight tool handlers

gimenete:
## Summary

Closes #1433 (Go SDK portion)

When `session.Abort()` is called, in-flight tool handlers now receive a cooperative cancellation signal via `ToolInvocation.Context`. A new `session.CancelToolCall(toolCallID)` method lets callers cancel a single handler without affecting others or the agentic loop.

### `go/types.go`
- Added `Context context.Context` to `ToolInvocation` — carries W3C Trace Context (OTel) **and** is cancelled when `session.Abort()` or `session.CancelToolCall()` is called.
- Deprecated `TraceContext` in favour of `Context` (same value; existing code using `TraceContext` continues to work unchanged).

### `go/session.go`
- Added `toolCallCancels map[string]context.CancelFunc` + mutex to `Session` to track in-flight tool call cancel funcs.
- `executeToolAndRespond` creates a `context.WithCancel` child of the trace context and registers its cancel func; cleans up (deregisters + calls `cancel()`) when the handler returns or panics.
- Two separate defers (LIFO) ensure panic recovery runs before cleanup, so the error RPC response is still sent when a panic occurs.
- RPC response calls (`HandlePendingToolCall`) use `traceCtx` (non-cancellable) so they succeed even if `Abort()` already fired.
- `Abort()` cancels all stored cancel funcs **and removes them from the map** after the RPC acknowledgment.
- **`CancelToolCall(toolCallID string) bool`** — cancels a single in-flight handler by ID. Returns `true` if found and cancelled, `false` if not in flight. Idempotent.

### `go/session_test.go`
- Extract `newRPCDrainSession` helper (pipe-based mock RPC, drains all requests with empty success responses).
- `TestToolInvocation_ContextCancelledOnAbort` — handler context cancelled by abort simulation; map cleaned up.
- `TestToolInvocation_ContextPopulated` — `Context` and `TraceContext` both set and equal.
- `TestCancelToolCall_cancelsTargetedHandlerOnly` — unknown ID returns false; targeted handler cancelled; sibling handler stays live; second call on same ID returns false.

### `go/README.md`
- "Cooperative Cancellation via session.Abort" section with HTTP example.
- "Cancelling a single tool call" section with `CancelToolCall` example and `ToolCallID` capture pattern.

## Backwards compatibility

Existing handlers that ignore `inv.Context` / `inv.TraceContext` are unaffected. `TraceContext` holds the same value as before (additive change).

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -355,6 +355,31 @@ session, _ := client.CreateSession(context.Background(), &copilot.SessionConfig{
 
 When the model selects a tool, the SDK automatically runs your handler (in parallel with other calls) and responds to the CLI's `tool.call` with the handler's result.
 
+#### Cooperative Cancellation via session.Abort
+
+`ToolInvocation.Context` is a `context.Context` that is cancelled when `session.Abort` is called. Pass it to any cancellable operation (HTTP requests, DB queries, sleeps) so the handler stops promptly when the session is aborted:
+
+```go
+lookupIssue := copilot.DefineTool("lookup_issue", "Fetch issue details from our tracker",
+    func(params LookupIssueParams, inv copilot.ToolInvocation) (any, error) {
+        // Pass inv.Context so the HTTP request is cancelled on session.Abort.
+        req, err := http.NewRequestWithContext(inv.Context, "GET",
+            "https://api.example.com/issues/"+params.ID, nil)
+        if err != nil {
+            return nil, err
+        }
+        resp, err := http.DefaultClient.Do(req)
+        if err != nil {
+            return nil, err // returns context.Canceled when aborted
+        }
+        defer resp.Body.Close()
+        // ...
+        return summary, nil
+    })
+```
+
+Handlers that don't use `inv.Context` are unaffected; they run to completion as before.
+
 #### Overriding Built-in Tools
 
 If you register a tool with the same name as a built-in CLI tool (e.g. `edit_file`, `read_file`), the SDK will throw an error unless you explicitly opt in by setting `OverridesBuiltInTool = true`. This flag signals that you intend to replace the built-in tool with your custom implementation.
--- a/go/README.md
+++ b/go/README.md
@@ -380,6 +380,40 @@ lookupIssue := copilot.DefineTool("lookup_issue", "Fetch issue details from our 
 
 Handlers that don't use `inv.Context` are unaffected; they run to completion as before.
 
+#### Cancelling a single tool call
+
+Use `session.CancelToolCall(toolCallID)` to cancel one specific in-flight handler without aborting the session or any other concurrent handlers. It returns `true` if the tool call was found and cancelled, `false` if it was not in flight.
+
+```go
+// Cancel a specific tool call by its ID.
+if cancelled := session.CancelToolCall(toolCallID); !cancelled {
+    log.Println("tool call was not in flight")
+}
+```
+
+`toolCallID` is available inside the handler as `inv.ToolCallID`. You can capture it to enable external cancellation of a specific operation:
+
+```go
+var mu sync.Mutex
+activeCalls := map[string]string{} // label → toolCallID
+
+slowTool := copilot.DefineTool("slow_op", "A long-running operation",
+    func(params SlowOpParams, inv copilot.ToolInvocation) (any, error) {
+        mu.Lock()
+        activeCalls[params.Label] = inv.ToolCallID
+        mu.Unlock()
+
+        // ... do work, checking inv.Context.Done() ...
+        return result, nil
+    })
+
+// Elsewhere, cancel by label:
+mu.Lock()
+id := activeCalls["my-label"]
+mu.Unlock()
+session.CancelToolCall(id)
+```
+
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
Thanks for putting together this coordinated cross-language work. The underlying goal—cooperatively cancelling in-flight tool handlers—is still valid, but these PRs share a design issue: cancellation only occurs when the same client calls `abort()`, so runtime-initiated or other-client aborts are missed. The proposed per-tool cancellation API is likewise local-only despite appearing session-wide.

We need to redesign this around the runtime's authoritative session abort event and reconsider targeted cancellation unless runtime support is added. Since these branches are also now stale, I'm closing them rather than asking for incremental updates. The useful cancellation primitives can be reused in a fresh coordinated implementation.

PR #1711, opened 2026-06-17
Add capi.enableWebSocketResponses and provider.transport session options

dereklegenzoff:
## Why

Two related WebSocket transport controls for the SDK's hand-written session options, both follow-ups to runtime PRs that were already merged/under review and explicitly called out as needing a corresponding SDK change:

1. **`capi.enableWebSocketResponses`** — WebSocket transport is now the default for the CAPI (Copilot API) Responses API whenever a model advertises the `ws:/responses` endpoint. That breaks for users behind proxies or in networks where WebSockets fail, so they need a way to opt out and fall back to the HTTP Responses transport. SDK-side follow-up to github/copilot-agent-runtime#10551 (requested in review by @SteveSandersonMS).
2. **`provider.transport`** — BYOK OpenAI-compatible providers need the opposite knob: an opt-in to deliver Responses API requests over a persistent WebSocket connection instead of HTTP (better for long-running, tool-call-heavy sessions that benefit from incremental `previous_response_id` continuations). SDK-side follow-up to github/copilot-agent-runtime#9557.

These are two distinct, complementary toggles: `capi.enableWebSocketResponses: false` opts **out** of WS for GitHub's Copilot API; `provider.transport: "websockets"` opts **in** to WS for your own BYOK endpoint. They are independent and coexist.

### 1. `capi.enableWebSocketResponses`

New nested session option `capi: { enableWebSocketResponses?: boolean }` on session **create** and **resume**. WebSocket transport is enabled by default when the model advertises `ws:/responses`; set this to `false` to force the HTTP Responses transport instead. Setting it to `false` is equivalent to setting the `COPILOT_CLI_DISABLE_WEBSOCKET_RESPONSES` environment variable (the env var name is unchanged). Scoped under a `capi` namespace (not a top-level option) by design: a single session can host multiple providers (CAPI + BYOK), so transport choice is conceptually provider-level.

Wire shape: `session.create` / `session.resume` params gain `capi: { enableWebSocketResponses: false }`, omitted entirely when unset.

### 2. `provider.transport`

New field `transport?: "http" | "websockets"` (default `"http"`) on the BYOK `ProviderConfig`. When `"websockets"`, Responses API requests for that OpenAI-compatible provider are delivered over a persistent WebSocket connection. Flows through the already-wired `provider` option on session create and resume.

Wire shape: `provider.transport` serializes as a plain camelCase string, omitted when unset.

## Approach

Both options are hand-written pass-throughs. Session create/resume options — and the `ProviderConfig` type — are hand-written in every language (not codegen-driven), so these ship independently of the runtime releases:

- `capi` is a new nested type mirroring the existing `provider` (BYOK) nested option exactly, wired into both the create and resume RPC params in every language.
- `transport` is a single new field on `ProviderConfig` mirroring the existing `wireApi` field; it needs no new wiring because the whole `provider` object is already serialized end-to-end.

Per-language summary:

- **Node.js** (reference): `CapiSessionOptions` interface + `SessionConfigBase.capi` wired in create/resume + re-export; `transport` on `ProviderConfig`. Tests added.
- **Python**: `CapiSessionOptions` TypedDict + camelCase wire helper + `capi` kwarg + export; `transport` on `ProviderConfig` TypedDict + wire mapping. Tests added.
- **Go**: `CapiSessionOptions` struct + `Capi` on configs/wire requests/builders; `Transport` on `ProviderConfig`. Tests added.
- **Java**: new `CapiSessionOptions` (fluent + Jackson) wired in `SessionRequestBuilder` for create/resume; `transport` field + getter/fluent-setter on `ProviderConfig`. Tests added.
- **.NET**: `CapiSessionOptions` class + `Capi` on `SessionConfigBase` (copy-ctor, both request records, serializer context); `Transport` on `ProviderConfig`. Tests added.
- **Rust**: `CapiSessionOptions` (`#[non_exhaustive]` + builders) + `capi` on configs/wire; `transport` field + `with_transport` builder on `ProviderConfig`. Tests added.

## Compatibility

Both fields are harmless no-ops on a CLI runtime that predates the corresponding runtime PR, so the SDK changes are safe to land ahead of (or independently of) the runtime releases.

## Deferred follow-up (by design)

- No `@github/copilot` dependency bump / codegen re-run. The current schema does not contain a generated `CapiSessionOptions`, and the generated `ProviderConfig` is not the consumer-facing type (the SDK exports a hand-written one). Once the runtime PRs publish, an optional bump + codegen re-run can produce generated types; the hand-written pass-throughs are sufficient in the meantime.
- No `session.options.update` (live update) wiring. This PR covers session create and resume only.

## Testing

Built and ran the per-language unit suites for every change — Node.js, Python, Go, Java, Rust, and .NET all pass locally (formatting/lint/clippy/spotless clean where applicable). e2e suites that require the replay-proxy harness were out of scope for local verification.

Refs github/copilot-agent-runtime#10551, github/copilot-agent-runtime#9557


dotnet/src/Types.cs
```diff
@@ -1991,6 +1991,15 @@ public sealed class ProviderConfig
     [JsonPropertyName("wireApi")]
     public string? WireApi { get; set; }
 
+    /// <summary>
+    /// Transport for OpenAI Responses requests ("http" or "websockets"). Defaults to "http".
+    /// Set to "websockets" to deliver Responses API requests over a persistent WebSocket
+    /// connection instead of HTTP. Applies to OpenAI-compatible providers using
```
stephentoub, line 1997: Do we ignore it for others or throw for others?

dotnet/src/Types.cs
```diff
@@ -2058,6 +2067,26 @@ public sealed class ProviderConfig
     public int? MaxOutputTokens { get; set; }
 }
 
+/// <summary>
+/// Provider-scoped options for the CAPI (Copilot API) provider.
+/// </summary>
+public sealed class CapiSessionOptions
```
stephentoub, line 2073: Do we use "capi" elsewhere in public API / documentation?

dotnet/src/Types.cs
```diff
@@ -2058,6 +2067,26 @@ public sealed class ProviderConfig
...
+    /// </summary>
+    /// <remarks>
+    /// WebSocket transport is the default for CAPI Responses API requests when the model advertises
+    /// the <c>ws:/responses</c> endpoint. Set this option for users behind proxies where WebSockets
+    /// fail. This is equivalent to setting the <c>COPILOT_CLI_DISABLE_WEBSOCKET_RESPONSES</c>
+    /// environment variable. The option is scoped under the <c>capi</c> namespace because a single
+    /// session can host multiple providers, such as CAPI and BYOK, so transport choice is provider-level.
+    /// </remarks>
+    [JsonPropertyName("disableWebSocketResponses")]
+    public bool? DisableWebSocketResponses { get; set; }
```
stephentoub: Should this be EnableWebSocketResponses and default to true rather than DisableWebSocketResponses and default to false? Personally I have a hard time reasoning about double negatives.

dotnet/src/Types.cs
dereklegenzoff: it throws currently although happy to make the behavior whatever you prefer.

dotnet/src/Types.cs
dereklegenzoff: happy to flip it

stephentoub (approved):

dotnet/src/Types.cs
dereklegenzoff: Not as a public option or type name — CapiSessionOptions and the capi session option are new in this PR. There's no other public property, type, or wire field named capi today.

"CAPI" does already show up in our docs, but only as the spelled-out "Copilot API (CAPI)" or as a telemetry term (e.g. the "CAPI interaction ID" field in streaming-events docs, and the CAPI_HMAC_KEY env var) — never as a config option name.

dotnet/src/Types.cs
dereklegenzoff: flipped

dotnet/src/Types.cs
dereklegenzoff: Updated the doc comments to mention "Copilot API (CAPI)" which aligns with other mentions

PR #1722, opened 2026-06-18
Recover JSON-RPC messages with malformed unicode escapes

ellismg:
## Summary

The Rust SDK's JSON-RPC read loop parsed each framed body with a strict `serde_json::from_slice`. A backend/CLI response containing a malformed `\u` escape — specifically a lone (unpaired) UTF-16 surrogate — fails with `unexpected end of hex escape`. That single parse error aborts the entire read loop (`JsonRpcClient::read_loop`), drains all pending requests, and tears down the transport. The list-client then reconnects and re-breaks on the same deterministic bytes, so every RPC on that channel (`list_models`, `get_account_quota`, `list_global_agents`, `list_global_skills`) fails together — on the github-app desktop app this manifests as a permanently empty model dropdown, quota, and agent/skill lists.

This is the **consumer-side defensive fix** for github/app#1055. Sibling producer-side fixes already landed in copilot-agent-runtime (#10300, #10231) to stop the CLI from emitting lone surrogates; this SDK change hardens the reader so a bad payload from **any** source (including an enterprise backend) can't brick the channel.

## What changed

- **New `rust/src/surrogate_safe.rs`** — `sanitize_json_escapes(&[u8]) -> (Cow<[u8]>, usize)`, a string-context-aware byte scanner that rewrites lone high/low UTF-16 surrogates and otherwise-malformed `\u` escapes to `\ufffd` (U+FFFD). It preserves valid surrogate pairs, valid BMP escapes, and never misreads an escaped `\\u`. Returns `Cow::Borrowed` (zero allocation) when the input is already strict-JSON safe.
- **`rust/src/jsonrpc.rs`** — `JsonRpcClient::read_message` now recovers on the parse-error path only: on failure it runs the sanitizer; if something was actually repaired (`Cow::Owned`) it `warn!`s with the replacement count + original error and re-parses; if nothing needed repair (`Cow::Borrowed`) the original error propagates unchanged. The happy path stays zero-cost.
- **`rust/tests/jsonrpc_test.rs`** — adds `malformed_unicode_escape_is_recovered_and_loop_survives`, a regression test that frames a notification carrying a lone high surrogate `\ud83d`, asserts it's delivered with the surrogate replaced by U+FFFD, and asserts a following well-formed message still arrives (proving the read loop survived).

No public API change and no consumer call-site change. `read_message` is the single chokepoint for every message from the CLI subprocess, so one guard here covers list-client RPCs, session-stream notifications, and resume frames. This mirrors the runtime's approach of substituting U+FFFD rather than dropping data.

## Validation

Run from `rust/` (matching `.github/workflows/rust-sdk-tests.yml`):

- `cargo +nightly-2026-04-14 fmt --all -- --config-path .rustfmt.nightly.toml --check` ✅
- `cargo clippy --all-targets --features test-support -- --no-deps -D warnings -D clippy::unwrap_used -D clippy::disallowed_macros -D clippy::await_holding_invalid_type` ✅
- `cargo test --no-default-features --features test-support` — 13 `surrogate_safe` unit tests + the new integration test + the existing `jsonrpc_test` suite all pass ✅ (the `e2e` suite requires a locally installed CLI and is unrelated to this change.)

Fixes the SDK side of github/app#1055.

### DIFF
--- a/rust/src/jsonrpc.rs
+++ b/rust/src/jsonrpc.rs
@@ -428,7 +428,30 @@ impl JsonRpcClient {
         let mut body = vec![0u8; length];
         reader.read_exact(&mut body).await?;
 
-        let message: JsonRpcMessage = serde_json::from_slice(&body)?;
+        let message: JsonRpcMessage = match serde_json::from_slice(&body) {
+            Ok(message) => message,
+            Err(parse_error) => {
+                // A strict parse failure is often a lone UTF-16 surrogate
+                // (or other malformed `\u` escape) emitted by a lenient
+                // upstream serializer. Rather than killing the read loop —
+                // which tears down the transport and retries forever on the
+                // same bytes (github/app#1055) — rewrite the bad escapes to
+                // U+FFFD and try once more. The sanitizer is a no-op for any
+                // other syntax error, so unrelated failures still propagate.
+                match crate::surrogate_safe::sanitize_json_escapes(&body) {
+                    (std::borrow::Cow::Owned(sanitized), replacements) => {
+                        warn!(
+                            replacements,
+                            original_error = %parse_error,
+                            "recovered JSON-RPC message containing malformed unicode escapes by \
+                             substituting U+FFFD"
+                        );
+                        serde_json::from_slice(&sanitized)?
+                    }
+                    (std::borrow::Cow::Borrowed(_), _) => return Err(parse_error.into()),
+                }
+            }
+        };
         Ok(Some(message))
     }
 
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
@ellismg I see this has been in draft for 7 weeks or so. Is it something you still want to complete and merge? Happy to close if not still required.

SteveSandersonMS:
I'll close this as stale since it doesn't seem like it's been needed in a long while, but feel free to reopen if this is mistaken!

PR #1745, opened 2026-06-22
Add Azure Managed Identity support for BYOK providers

SteveSandersonMS:
## Summary

Adds Azure Managed Identity (MI) support to the Node SDK for BYOK providers, matching the runtime contract. App developers can authenticate a BYOK provider with a system- or user-assigned managed identity instead of a static `apiKey`/`bearerToken`, and the runtime handles token acquisition, caching, and refresh automatically.

## Usage

```ts
// System-assigned identity (default cognitiveservices scope)
provider: { type: "openai", baseUrl, modelId, managedIdentity: {} }

// User-assigned identity (clientId | objectId | resourceId) and/or custom scope
provider: { type: "openai", baseUrl, modelId, managedIdentity: { clientId: "..." } }
```

`managedIdentity` is the generated `ManagedIdentityConfig` type. It is a single all-optional object: an empty object selects the system-assigned identity; setting at most one of `clientId`/`objectId`/`resourceId` selects a user-assigned identity; `scope` overrides the token audience. It is mutually exclusive with `apiKey`/`bearerToken`.

## What's included

- Re-export the generated `ManagedIdentityConfig` type from the public SDK surface (`types.ts`, `index.ts`).
- A shared, language-agnostic mock identity endpoint (`test/harness/`) plus a Node entrypoint, so other-language SDK e2e suites can reuse the same fake IMDS/App-Service identity server.
- Hermetic e2e coverage (`managed_identity.e2e.test.ts`): system-assigned bearer injection, user-assigned `clientId` with custom scope, token caching across turns, and token refresh within the expiry buffer — no network access.

## Wire shape

`managedIdentity` is a single object (not a `boolean | object` union). The union shape was rejected by the runtime's own SDK-mappability lint and could not be code-generated for statically-typed SDKs, so the runtime contract uses one all-optional object; `{}` replaces the former `true`.

## Testing

- `npm run typecheck` (both tsconfigs), prettier, eslint: green
- 4/4 managed-identity e2e tests pass against the runtime `dist-cli`

PR #1746, opened 2026-06-22
Add getBearerToken callback for BYOK providers (Managed Identity v1)

SteveSandersonMS:
## Summary

Adds **SDK-side bearer-token resolution** for BYOK providers — the v1 design for Azure Managed Identity support. Consumers supply a per-provider `getBearerToken` callback (e.g. wrapping `@azure/identity`'s `DefaultAzureCredential` / `ManagedIdentityCredential`). The Copilot SDK takes **zero** Azure dependency, and the Rust runtime stays **token-agnostic**: it asks for a token before each request, uses it as the `Authorization` header, and caches/refreshes automatically.

Pairs with the runtime half: github/copilot-agent-runtime#10933.

## Public surface

`getBearerToken?: (args: { providerName, scope }) => Promise<{ token, expiresOnTimestamp } | string>` plus `bearerTokenScope?` on `ProviderConfig` and `NamedProviderConfig`. Return the rich object so the runtime refreshes only near expiry; a bare string is also accepted (re-fetched every request).

## How it works

The callback isn't serializable over JSON-RPC, so the SDK strips it and sends a `bearerTokenProvider: true` wire flag (same pattern as hooks / userInput). The runtime synthesizes a `getBearerToken` that dispatches back over the session-scoped `providerToken.acquire` RPC; the SDK routes each acquisition to the matching **per-provider** callback by provider name, so multi-provider sessions stay correctly partitioned.

The bearer-token surface is **provider-agnostic** — `bearerTokenScope` is forwarded verbatim (empty when unset), with no baked-in default scope.

## Changes

- `types.ts` — public `ProviderTokenArgs` / `ProviderBearerToken` / `GetBearerToken` types; `getBearerToken` / `bearerTokenScope` on the provider configs.
- `generated/rpc.ts` — wire fields, `providerToken.acquire` DTOs + handler, registration.
- `session.ts` — per-provider `bearerTokenProviders` registry; routes `providerToken.acquire` by provider name.
- `client.ts` — extracts callbacks, registers them, and emits the wire flag in create/resume payloads.

## Tests

New e2e suite (`byok_bearer_token_provider.e2e.test.ts`) over the real RPC transport + replay proxy:
1. callback token reaches the model as `Authorization: Bearer <token>`;
2. token is re-acquired when the previous one has expired;
3. per-provider dispatch routes each turn to its own callback.

All three pass in replay mode.

Supersedes the earlier declarative-config approach (#1745).

### DIFF
--- a/nodejs/src/client.ts
+++ b/nodejs/src/client.ts
@@ -47,11 +47,14 @@ import type {
     ExitPlanModeResult,
     ForegroundSessionInfo,
     GetAuthStatusResponse,
+    GetBearerToken,
     GetStatusResponse,
     InternalRuntimeConnection,
     LargeToolOutputConfig,
     MCPServerConfig,
     ModelInfo,
+    NamedProviderConfig,
+    ProviderConfig,
     ResumeSessionConfig,
     SectionTransformFn,
     SessionConfig,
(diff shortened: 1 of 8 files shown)


SteveSandersonMS:
Closing in favour of https://github.com/github/copilot-sdk/pull/1748

PR #1748, opened 2026-06-22
Add getBearerToken callback for BYOK providers (Managed Identity)

SteveSandersonMS:
## Summary

Adds an experimental **`getBearerToken` callback** to BYOK provider configs so the SDK consumer can resolve bearer tokens (e.g. **Azure Managed Identity** via the consumer's identity library) on demand. The Copilot SDK takes **zero** Azure dependency — the consumer fills in the callback with whatever identity library they like. The runtime does **no caching** — it invokes the callback once per request, so the consumer (or the identity library it wraps) owns caching/refresh. Scope/audience is closed over by the callback and never crosses the wire.

This PR ships the feature across **all six SDKs** — TypeScript/Node, .NET, Python, Go, Rust, and Java — each with an equivalent three-scenario e2e test.

## Usage examples

Each example wires `DefaultAzureCredential` (Azure Managed Identity) into the callback for the `https://cognitiveservices.azure.com/.default` scope. The identity library caches/refreshes the underlying token, so calling it per request is cheap.

### TypeScript / Node

```ts
import { DefaultAzureCredential } from "@azure/identity";
import type { GetBearerToken, ProviderConfig } from "@github/copilot";

const cred = new DefaultAzureCredential();

const provider: ProviderConfig = {
    type: "openai",
    baseUrl,
    getBearerToken: async () => (await cred.getToken()).token,
};
```

### .NET

```csharp
using Azure.Core;
using Azure.Identity;

var cred = new DefaultAzureCredential();

var provider = new ProviderConfig
{
    Type = "openai",
    BaseUrl = baseUrl,
    GetBearerToken = async args => (await cred.GetTokenAsync()).Token,
};
```

### Python

```python
from azure.identity.aio import DefaultAzureCredential

cred = DefaultAzureCredential()

async def get_bearer_token(args) -> str:
    token = await cred.get_token()
    return token.token

provider = {
    "type": "openai",
    "base_url": base_url,
    "get_bearer_token": get_bearer_token,
}
```

### Go

```go
import (
    "context"
    "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
    "github.com/Azure/azure-sdk-for-go/sdk/azidentity"
)

cred, err := azidentity.NewDefaultAzureCredential(nil)
if err != nil {
    return err
}
scope := "https://cognitiveservices.azure.com/.default"

provider := copilot.ProviderConfig{
    Type:    "openai",
    BaseURL: baseURL,
    GetBearerToken: func(args copilot.ProviderTokenArgs) (string, error) {
        tok, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{
            Scopes: []string{scope},
        })
        if err != nil {
            return "", err
        }
        return tok.Token, nil
    },
}
```

### Rust

```rust
use std::sync::Arc;
use azure_identity::DefaultAzureCredential;
use copilot::{ProviderConfig, ProviderTokenArgs};

let cred = DefaultAzureCredential::new()?;
let scope = "https://cognitiveservices.azure.com/.default";

let provider = ProviderConfig::new(base_url)
    .with_provider_type("openai")
    .with_get_bearer_token(Arc::new(move |_args: ProviderTokenArgs| {
        let cred = cred.clone();
        async move {
            let token = cred
                .get_token(&[scope])
                .await
                .map_err(|e| BearerTokenError::new(e.to_string()))?;
            Ok(token.token.secret().to_string())
        }
    }));
```

### Java

```java
import com.azure.identity.DefaultAzureCredentialBuilder;
import com.azure.core.credential.TokenRequestContext;

var cred = new DefaultAzureCredentialBuilder().build();
var ctx = new TokenRequestContext().addScopes("https://cognitiveservices.azure.com/.default");

var provider = new ProviderConfig()
    .setType("openai")
    .setBaseUrl(baseUrl)
    .setGetBearerToken(args ->
        cred.getToken(ctx).map(t -> t.getToken()).toFuture());
```


SteveSandersonMS:
Thanks — a couple of these points are now moot after a contract simplification pushed since this review ran:

- **`bearerTokenScope` has been removed entirely** from the wire contract and from both provider configs. Scope/audience is closed over by the consumer''s `getBearerToken` callback (e.g. `@azure/identity` already binds the scope at `getToken(scope)` time), so it never needs to cross the wire. Point 1 no longer applies to any SDK.
- The runtime now does **no token caching** — `expiresOnTimestamp` was also dropped from the wire result; the callback owns caching/refresh.

So the remaining cross-SDK surface is just (2) the `getBearerToken` callback and (3) the `providerToken.acquire` server→client handler. This is intentionally Node-first and `@experimental`; parity for Python/Go/.NET/Java/Rust is planned as follow-up, not part of this PR. I''ll open tracking issues for the other SDKs once the Node + runtime contract lands.

dotnet/src/Types.cs
```diff
@@ -2032,6 +2034,28 @@ public sealed class ProviderConfig
...
+    /// Per-request callback that resolves a bearer token on demand for this BYOK provider (for
+    /// example via Azure Managed Identity). The Copilot SDK takes no identity dependency: supply a
+    /// callback backed by your own identity library. Never serialized — setting it makes the SDK send
+    /// <c>hasBearerTokenProvider: true</c> on the wire and answer the runtime's
+    /// <c>providerToken.getToken</c> requests. Mutually exclusive with <see cref="ApiKey"/> and
+    /// <see cref="BearerToken"/>.
+    /// </summary>
+    [JsonIgnore]
+    [Experimental(Diagnostics.Experimental)]
+    public Func<ProviderTokenArgs, Task<string>>? GetBearerToken { get; set; }
```
scottaddie, line 2066: Two design concerns here that are worth addressing while this is still in draft:

**Naming**: `GetBearerToken` sitting directly below `BearerToken` reads as a variant of the same setting rather than a fundamentally different auth mode. In an object initializer, the difference between a static string and a callback is easy to miss. I'd suggest `GetBearerTokenProvider` — it signals "dynamic token source" rather than "another kind of bearer token value." I'm drawing inspiration from what was done for the OpenAI SDK, which uses https://github.com/Azure/azure-sdk-for-js/blob/a5cdd18b24553a0886bd59964bd9fad58bddb435/sdk/identity/identity/src/tokenProvider.ts#L48.

**Mutual exclusion**: All 3 auth options (`ApiKey`, `BearerToken`, `BearerTokenProvider`) are flat, independent, settable properties with no SDK-side guard. The docs mark them mutually exclusive, but nothing prevents a caller from setting 2 (or all 3) at once. Without a guard, which one wins is a runtime implementation detail the caller can't easily discover. Two alternatives worth considering:
- Add validation at session startup (throw `InvalidOperationException` if more than one is non-null)
- Model auth as a single union-like property, e.g. `Authentication = new ApiKey(...) | new BearerToken(...) | new BearerTokenProvider(...)`

stephentoub (approved):

Issue #1762, opened 2026-06-22
[Java] @CopilotTool ergonomics 4.5: E2E integration test with replay proxy


edburns:
## Overview

Create an E2E failsafe integration test that proves the ergonomic `@CopilotTool` + `ToolDefinition.fromObject()` API produces identical wire behavior to the low-level `ToolDefinition.create()` API, tested against the replay proxy.

**Branch:** `edburns/1682-java-tool-ergonomics` on `upstream` (⚠️ NOT `main` — PRs must target this branch)

## Prerequisites

- Tasks 4.1–4.4 must be complete and merged to the branch.
- Before writing any code, read the entire implementation plan at:
  `1682-java-tool-ergonomics-prompts-remove-before-merge/dd-3018003-ignorance-reduction-for-implementation-plan.md`

## Relevant plan sections to carefully re-read

- **Section 4.5 — E2E integration test** (the primary task description, includes test code outline)
- **Phase 2 ✅ — Verify the existing low-level path works in Java** (context: `LowLevelToolDefinitionIT` is the reference)
- The existing skill file for creating Java E2E tests: `.github/skills/new-java-e2e-test-yaml-and-test/SKILL.md` — read this skill file for the exact patterns, conventions, and harness usage for creating new E2E tests.

### Files to create

1. `test/snapshots/tools/ergonomic_tool_definition.yaml` — replay proxy snapshot (may be identical to `low_level_tool_definition.yaml` since the wire format is the same; if so, symlink or copy is acceptable).
2. `java/src/test/java/com/github/copilot/e2e/ErgonomicToolDefinitionIT.java` — the failsafe IT class.

### Test specification

The test must define a tools class using the ergonomic API:

```java
class MyTestTools {
    String currentPhase;

    @CopilotTool("Sets the current phase of the agent")
    String setCurrentPhase(@Param("The phase to transition to") String phase) {
        currentPhase = phase;
        return "Phase set to " + phase;
    }

    @CopilotTool("Search for items by keyword")
    String searchItems(@Param("Search keyword") String keyword) {
        return "Found: item_alpha, item_beta";
    }

    @CopilotTool(value = "Custom grep override", name = "grep", overridesBuiltInTool = true)
    String grepOverride(@Param("Search query") String query) {
        return "CUSTOM_GREP: " + query;
    }
}
```

The test method:
1. Creates `MyTestTools` instance.
2. Calls `ToolDefinition.fromObject(tools)` to get tool definitions.
[...]

Issue #1761, opened 2026-06-22
[Java] @CopilotTool ergonomics 4.4: ToolDefinition.fromObject(Object) registration API


edburns:
Iterating in #1779 

## Overview

Add static methods `ToolDefinition.fromObject(Object)` and `ToolDefinition.fromClass(Class<?>)` that load processor-generated `$$CopilotToolMeta` classes and return `List<ToolDefinition>` with fully working tool definitions (schema + invocation handlers).

**Branch:** `edburns/1682-java-tool-ergonomics` on `upstream` (⚠️ NOT `main` — PRs must target this branch)

## Prerequisites

- Tasks 4.1 (annotations), 4.2 (SchemaGenerator), and 4.3 (CopilotToolProcessor) must be complete and merged to the branch.
- Before writing any code, read the entire implementation plan at:
  `1682-java-tool-ergonomics-prompts-remove-before-merge/dd-3018003-ignorance-reduction-for-implementation-plan.md`

## Relevant plan sections to carefully re-read

- **Section 3.6 — `ToolDefinition.fromObject(Object)` registration API** (Resolution: processor-only approach, no reflection fallback)
- **Section 3.7 — `module-info.java` impact** (Resolution: Class.forName works within same module/classloader)
- **Section 4.4 — `ToolDefinition.fromObject(Object)`** (the primary task description)

### Files to modify

1. `java/src/main/java/com/github/copilot/rpc/ToolDefinition.java` — add `fromObject(Object)` and `fromClass(Class<?>)` static methods.

### ⚠️ Critical: ObjectMapper contract with generated code

The generated `$$CopilotToolMeta.definitions()` method accepts an `ObjectMapper` as its **second parameter**. This is an internal contract established in PR #1777 (task 4.3, `CopilotToolProcessor`).

**Why:** The generated code uses the ObjectMapper for:
1. Argument coercion — converting complex types (enums, records, POJOs, `java.time.*`) from the args Map
2. Result serialization — converting non-String return values to JSON

The mapper must be configured identically to `JsonRpcClient.createObjectMapper()`:
- `JavaTimeModule` registered
- `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES = false`
- `SerializationFeature.WRITE_DATES_AS_TIMESTAMPS = false`
- `JsonInclude.Include.NON_NULL`

**Generated method signature (internal contract):**
```java
// Generated $$CopilotToolMeta class (package-private)
[...]

Issue #1760, opened 2026-06-22
[Java] @CopilotTool ergonomics 4.3: Annotation processor (CopilotToolProcessor)


edburns:
## Overview

Create the JSR 269 annotation processor (`CopilotToolProcessor`) that finds `@CopilotTool`-annotated methods and generates `$$CopilotToolMeta` companion classes containing tool definitions, JSON Schema, and invocation lambdas.

**Branch:** `edburns/1682-java-tool-ergonomics` on `upstream` (⚠️ NOT `main` — PRs must target this branch)

## Prerequisites

- Task 4.1 (annotations) and Task 4.2 (SchemaGenerator) must be complete and merged to the branch.
- Before writing any code, read the entire implementation plan at:
  `1682-java-tool-ergonomics-prompts-remove-before-merge/dd-3018003-ignorance-reduction-for-implementation-plan.md`

## Relevant plan sections to carefully re-read

- **Section 3.2 — `@CopilotTool` annotation design** (Resolution: RUNTIME retention, ToolDefer support)
- **Section 3.3 — `@Param` annotation design** (Resolution: defaultValue semantics and validation rules)
- **Section 3.5 — Generated code shape** (Resolution: access levels, return type handling, argument deserialization patterns, ObjectMapper source)
- **Section 3.7 — `module-info.java` impact** (Resolution: no issues, generated class in same package)
- **Section 3.8 — Processor registration** (Resolution: standard JSR 269 multi-processor registration)
- **Section 4.3 — Annotation processor (`CopilotToolProcessor`)** (the primary task description)

### Files to create

1. `java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java`

### Files to modify

1. `java/src/main/resources/META-INF/services/javax.annotation.processing.Processor` — add `com.github.copilot.tool.CopilotToolProcessor`
2. `java/src/main/java/module-info.java` — add `provides javax.annotation.processing.Processor with ..., com.github.copilot.tool.CopilotToolProcessor;`

### Generated code specification

For a class `com.example.MyTools` containing `@CopilotTool` methods, the processor generates `com.example.MyTools$$CopilotToolMeta` in the same package:

```java
// GENERATED by CopilotToolProcessor — do not edit
package com.example;

import com.github.copilot.rpc.ToolDefinition;
import com.github.copilot.rpc.RpcMapper;
import com.fasterxml.jackson.databind.ObjectMapper;
import java.util.*;
[...]

Issue #1759, opened 2026-06-22
[Java] @CopilotTool ergonomics 4.2: Schema generation utility (TypeMirror to JSON Schema)


edburns:
## Overview

Create a compile-time schema generation utility that maps `javax.lang.model` types (`TypeMirror`, `DeclaredType`) to JSON Schema represented as Java source code literals (`Map.of(...)` expressions).

**Branch:** `edburns/1682-java-tool-ergonomics` on `upstream` (⚠️ NOT `main` — PRs must target this branch)

## Prerequisites

- Task 4.1 (annotations) must be complete and merged to the branch.
- Before writing any code, read the entire implementation plan at:
  `1682-java-tool-ergonomics-prompts-remove-before-merge/dd-3018003-ignorance-reduction-for-implementation-plan.md`

## Relevant plan sections to carefully re-read

- **Section 3.4 — Type-to-JSON-Schema mapping** (Resolution: maximum viable set of 23 type mappings with full table)
- **Section 3.5 — Generated code shape** (Resolution: how generated code uses schema maps)
- **Section 4.2 — Schema generation utility (compile-time)** (the primary task description)

### Files to create

1. `java/src/main/java/com/github/copilot/tool/SchemaGenerator.java`

### Functional specification

`SchemaGenerator` is a **compile-time utility** invoked by the annotation processor. It receives `javax.lang.model.type.TypeMirror` instances and produces a `String` containing the Java source code literal for the JSON Schema `Map`. For example, given a `String` type, it should produce the source code string: `Map.of("type", "string")`.

#### Type mappings to implement (from Resolution 3.4)

| Java type | JSON Schema source literal |
|-----------|---------------------------|
| `String` | `Map.of("type", "string")` |
| `int`, `Integer` | `Map.of("type", "integer")` |
| `long`, `Long` | `Map.of("type", "integer")` |
| `double`, `Double` | `Map.of("type", "number")` |
| `float`, `Float` | `Map.of("type", "number")` |
| `boolean`, `Boolean` | `Map.of("type", "boolean")` |
| `String[]` | `Map.of("type", "array", "items", Map.of("type", "string"))` |
| `enum` types | `Map.of("type", "string", "enum", List.of("V1", "V2", ...))` (enumerate all constants) |
| `UUID` | `Map.of("type", "string", "format", "uuid")` |
| `OffsetDateTime` | `Map.of("type", "string", "format", "date-time")` |
| `JsonNode` | `Map.of()` (any) |
[...]

PR #1792, opened 2026-06-24
Java: Implement `@CopilotTool` ergonomics

edburns:
# PR 1792 Reviewer’s Guide — Java `@CopilotTool` Ergonomics

PR: https://github.com/github/copilot-sdk/pull/1792  
Epic: https://github.com/github/copilot-sdk/issues/1682  
ADR: `java/docs/adr/adr-005-tool-definition.md`

This PR introduces the Java high-level tool API (`@CopilotTool` + `@Param`) and compile-time metadata generation, while keeping low-level `ToolDefinition.create(...)` as the foundation.

## What to optimize for in review

Please focus on **API semantics**, **generated-invocation correctness**, and **schema fidelity** rather than style. The main question is whether this design is robust enough for long-term Java SDK ergonomics and interoperability.

## Scope mapped to epic child issues

| Epic item | Delivered in PR 1792 |
|---|---|
| #1758 | `@CopilotTool` and `@Param` annotations |
| #1759 | Compile-time `SchemaGenerator` (`TypeMirror` -> JSON Schema source) |
| #1760 | JSR-269 `CopilotToolProcessor` generating `$$CopilotToolMeta` |
| #1761 | `ToolDefinition.fromObject(...)` / `fromClass(...)` registration bridge |
| #1762 | E2E integration test and snapshot for ergonomic path |
| #1789 | Follow-up item remains open; related hardening landed here (default/optional handling) |

## Commit stream reading order (recommended)

Ignore WIP/prompt commits and review in this logical order:

1. `feat(java): create @CopilotTool and @Param annotations (#1763)`
2. `feat(java): add SchemaGenerator compile-time type-to-JSON-Schema utility (#1766)`
3. `feat(java): Add CopilotToolProcessor annotation processor (task 4.3) (#1777)`
4. `feat(java): Add ToolDefinition.fromObject() and fromClass() registration API (#1779)`
5. `Add E2E integration test for ergonomic @CopilotTool + ToolDefinition.fromObject() API (#1787)`
6. Hardening commits: Optional extraction fix, static-tool qualified-name fix, java.time schema format hints, CodeQL cleanup, snapshot stabilization, test generation fixes.

## File-level review map

| Area | Primary files |
|---|---|
| Public API surface | `java/src/main/java/com/github/copilot/tool/CopilotTool.java`, `java/src/main/java/com/github/copilot/tool/Param.java`, `java/src/main/java/com/github/copilot/rpc/ToolDefinition.java` |
| Metadata provider contract | `java/src/main/java/com/github/copilot/tool/CopilotToolMetadataProvider.java` |
| Processor/codegen | `java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java`, `java/src/main/resources/META-INF/services/javax.annotation.processing.Processor`, `java/src/main/java/module-info.java` |
| Schema mapping | `java/src/main/java/com/github/copilot/tool/SchemaGenerator.java` |
| Unit tests (processor) | `java/src/test/java/com/github/copilot/tool/CopilotToolProcessorTest.java`, `java/src/test/java/com/github/copilot/tool/SchemaGeneratorTest.java` |
| Unit tests (runtime bridge) | `java/src/test/java/com/github/copilot/rpc/ToolDefinitionFromObjectTest.java` + fixtures under `java/src/test/java/com/github/copilot/rpc/fixtures/` |
| E2E parity proof | `java/src/test/java/com/github/copilot/e2e/ErgonomicToolDefinitionIT.java`, `test/snapshots/tools/ergonomic_tool_definition.yaml` |

## Architectural intent (as implemented)

1. **User-facing model**: langchain4j-style method annotations.
2. **Implementation strategy**: Micronaut-style compile-time processor/codegen, not reflection-first.
3. **Bridge**: `ToolDefinition.fromObject/fromClass` loads generated `$$CopilotToolMeta`.
4. **No silent fallback**: missing generated metadata throws explicit `IllegalStateException`.
5. **Low-level path preserved**: `ToolDefinition.create*` remains available.

## High-value review questions (actionable feedback requested)

1. Is `@Retention(RUNTIME)` on `@CopilotTool`/`@Param` still the right long-term choice now that `fromObject()` is processor-only by default (no runtime fallback path)?
2. Is requiring non-private tool methods the right constraint vs. adding MethodHandles-based access?
3. Is the current default-value policy complete and safe?  
   - `required=true` + `defaultValue` is compile-time error  
   - defaults applied only when key is absent  
   - generated path semantics match intended runtime behavior
4. Is return-type handling policy correct for real agents?  
   - `String` passthrough  
   - `void -> "Success"`  
   - `CompletableFuture<String>` passthrough (object-cast)  
   - non-string results JSON-serialized
5. Is exception policy acceptable in generated lambdas (`RuntimeException` wrapping during serialization)?
6. Is `SchemaGenerator` coverage sufficient for v1 (primitives, boxed, collections, maps, records/classes, enums, sealed interfaces, java.time formats, Optional variants)?
7. Is `Param.name()` interaction with generated extraction/schema intuitive and predictable for advanced use?
8. Is the static-tool path (`fromClass`) now robust enough after qualified-class-name and instance-method guard fixes?
9. Is `ToolDefer.NONE -> null` wire behavior the right compatibility choice?
10. Any JPMS edge cases missed around generated metadata loading, module boundaries, or processor registration?

## Deliberate non-goals / constraints in this PR

- No broad reflection fallback in `fromObject()` when metadata is missing (fail fast instead).
- No change to low-level tool-definition API contract.
- Ergonomic API remains `@CopilotExperimental`.

## Extra context from iteration artifacts

The prompt/design trail in `copilot-sdk-01/1682-java-tool-ergonomics-prompts-remove-before-merge` tracks the ignorance-reduction phase and key decisions that shaped this PR (especially annotation shape, default semantics, processor-only direction, and schema coverage expansions).

If you see a mismatch between ADR intent and implementation behavior, that is exactly the class of feedback I want.

## Note on `CopilotToolMetadataProvider`

`CopilotToolMetadataProvider<T>` is the runtime seam between generated metadata and `ToolDefinition.fromObject(...)` loading.  
The processor emits `$$CopilotToolMeta` classes that implement this interface and expose:

```java
List<ToolDefinition> definitions(T instance, ObjectMapper mapper)
```

This keeps `fromObject/fromClass` decoupled from generated-class internals, and also allows advanced/manual implementations of the provider contract when needed.

### 1. `pom.xml` — experimental + annotation processor

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <version>3.15.0</version>
    <configuration>
        <compilerArgs>
            <arg>-Acopilot.experimental.allowed=true</arg>
        </compilerArgs>
        <annotationProcessorPaths>
            <path>
                <groupId>com.github</groupId>
                <artifactId>copilot-sdk-java</artifactId>
                <version>${copilot.sdk.version}</version>
            </path>
        </annotationProcessorPaths>
    </configuration>
</plugin>
```

### 2. Tool declaration with `@CopilotTool`

```java
@CopilotTool("Get the current weather for a given city")
public String getWeather(
        @Param(value = "The city to get weather for", required = true) String city,
        @Param(value = "Temperature unit: celsius or fahrenheit", required = false, defaultValue = "celsius") String unit) {
    // ...
}
```

### 3. Registration via `ToolDefinition.fromObject(...)`

```java
WeatherTools weatherTools = new WeatherTools();
List<ToolDefinition> toolDefs = ToolDefinition.fromObject(weatherTools);

SessionConfig sessionConfig = new SessionConfig()
        .setOnPermissionRequest(PermissionHandler.APPROVE_ALL)
        .setAvailableTools(new ToolSet().addCustom("*"))
        .setTools(toolDefs);
```


java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,567 @@
...
+                writeMetaClass(out, packageName, simpleClassName, metaClassName, classElement, methods);
+            }
+        } catch (IOException e) {
+            processingEnv.getMessager().printMessage(Diagnostic.Kind.ERROR,
+                    "Failed to generate " + metaClassName + ": " + e.getMessage(), classElement);
+        }
+    }
+
+    private void writeMetaClass(PrintWriter out, String packageName, String simpleClassName, String metaClassName,
+            TypeElement classElement, List<ExecutableElement> methods) {
```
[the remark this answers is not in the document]
edburns: Fixed in e70f1a91 .

java/src/main/java/com/github/copilot/tool/SchemaGenerator.java
```diff
@@ -0,0 +1,384 @@
...
+        List<? extends TypeMirror> permittedSubclasses = typeElement.getPermittedSubclasses();
+        if (permittedSubclasses != null && !permittedSubclasses.isEmpty()) {
+            List<String> schemas = permittedSubclasses.stream().map(sub -> generateSchema(sub, typeUtils, elementUtils))
+                    .collect(Collectors.toList());
+            return "Map.of(\"oneOf\", List.of(" + String.join(", ", schemas) + "))";
+        }
+        return "Map.of(\"type\", \"object\")";
+    }
+
+    private boolean isOptionalType(TypeMirror type, Types typeUtils, Elements elementUtils) {
```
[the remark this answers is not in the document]
edburns: Fixed in e70f1a91 .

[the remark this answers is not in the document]
edburns: Fixed in e70f1a91 .

java/src/main/java/com/github/copilot/tool/SchemaGenerator.java
```diff
@@ -0,0 +1,384 @@
...
+            return false;
+        }
+        DeclaredType declaredType = (DeclaredType) type;
+        TypeElement element = (TypeElement) declaredType.asElement();
+        String name = element.getQualifiedName().toString();
+        return "java.util.Optional".equals(name) || "java.util.OptionalInt".equals(name)
+                || "java.util.OptionalDouble".equals(name) || "java.util.OptionalLong".equals(name);
+    }
+
+    private TypeMirror unwrapOptional(TypeMirror type, Types typeUtils, Elements elementUtils) {
```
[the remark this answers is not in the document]
edburns: Fixed in e70f1a91 .

java/src/test/java/com/github/copilot/e2e/ErgonomicTestTools.java
```diff
@@ -0,0 +1,32 @@
...
+    String currentPhase;
+
+    @CopilotTool("Sets the current phase of the agent")
+    public String setCurrentPhase(@Param("The phase to transition to") String phase) {
+        currentPhase = phase;
+        return "Phase set to " + phase;
+    }
+
+    @CopilotTool("Search for items by keyword")
+    public String searchItems(@Param("Search keyword") String keyword) {
```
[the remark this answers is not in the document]
edburns: Fixed in e70f1a91 .

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,567 @@
...
+                                .append(" = ").append(generateArgExtractionFromMap(paramName, paramType)).append(";\n");
+                    }
+                }
+            }
+        }
+
+        // Generate method invocation based on return type
+        TypeMirror returnType = method.getReturnType();
+        String methodCall = "instance." + method.getSimpleName() + "(" + generateArgList(params) + ")";
+
```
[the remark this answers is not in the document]
edburns: This is a well-known Java language quirk. Static method dispatch is resolved at **compile time** based on the declared type, not at runtime via the object reference. The reference is never actually dereferenced.

```java
StaticTools instance = null;
instance.greet("World");  // No NPE! Compiles to: invokestatic StaticTools.greet
```

The compiler sees that `greet` is `static` on `StaticTools`, so it emits `invokestatic` bytecode. The `instance` variable is completely ignored at runtime — the JVM never reads the pointer. It's equivalent to writing `StaticTools.greet("World")`.

The JLS specifies this in §15.12.4.1: for static method invocations via an instance expression, the expression is evaluated (for side effects) but the result is discarded; the method is invoked without using it.

IDEs will warn ("Accessing static member via instance reference") but it's legal and NPE-safe. The reviewer's assertion is incorrect — the test passing confirms it.


Even so, we applied your suggestion in b7a9cbd5 .

java/src/main/java/com/github/copilot/tool/SchemaGenerator.java
```diff
@@ -0,0 +1,384 @@
...
+
+        // UUID
+        if ("java.util.UUID".equals(qualifiedName)) {
+            return "Map.of(\"type\", \"string\", \"format\", \"uuid\")";
+        }
+
+        // OffsetDateTime
+        if ("java.time.OffsetDateTime".equals(qualifiedName)) {
+            return "Map.of(\"type\", \"string\", \"format\", \"date-time\")";
+        }
```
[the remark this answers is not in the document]
edburns: Fixed in `0a99a872`

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,567 @@
...
+            }
+            if (isBoxedNumeric(qualifiedName)) {
+                return generateBoxedNumericExtraction("args.get(\"" + paramName + "\")", qualifiedName);
+            }
+            if ("java.lang.Boolean".equals(qualifiedName)) {
+                return "(Boolean) args.get(\"" + paramName + "\")";
+            }
+            // Complex types: enums, records, POJOs
+            return "mapper.convertValue(args.get(\"" + paramName + "\"), " + qualifiedName + ".class)";
+        }
```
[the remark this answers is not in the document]
edburns: Fixed in 35c74509 .

rotationtv1-crypto:
edburns/1682-java-tool-ergonomics-review-draft-01

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,636 @@
...
+                        sb.append("                    Object ").append(varName).append("Raw = args.containsKey(\"")
+                                .append(paramName).append("\") ? args.get(\"").append(paramName).append("\") : ")
+                                .append(generateDefaultLiteral(paramType, defaultValue)).append(";\n");
+                        sb.append("                    ").append(getTypeString(paramType)).append(" ").append(varName)
+                                .append(" = ").append(generateArgExtraction(varName + "Raw", paramType)).append(";\n");
+                    } else if (isOptionalType(paramType)) {
+                        generateOptionalExtraction(sb, paramName, varName, paramType);
+                    } else {
+                        sb.append("                    ").append(getTypeString(paramType)).append(" ").append(varName)
+                                .append(" = ").append(generateArgExtractionFromMap(paramName, paramType)).append(";\n");
```
jdubois, line 364: Optional primitives NPE at runtime. A `@Param(required=false) int limit` is correctly dropped from the schema's `required`, but this branch still emits `int limit = ((Number) args.get("limit")).intValue();`. When the model omits the arg, `args.get("limit")` is null and the unboxing throws `NullPointerException: Cannot invoke "java.lang.Number.intValue()" because ... Map.get(Object) is null` (confirmed by running the real processor).

A primitive simply can't represent "absent." I'd reject this at compile time — emit an error telling the user to use a boxed type (`Integer`), an `Optional`, or supply a `defaultValue` — which matches the existing compile-time validations for private methods and required+default conflicts.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,636 @@
...
+        // Generate argument extraction
+        if (!params.isEmpty()) {
+            sb.append("Map<String, Object> args = invocation.getArguments();\n");
+
+            // Check if single-record-parameter shortcut applies
+            if (params.size() == 1 && isRecord(params.get(0).asType())) {
+                String typeName = getTypeString(params.get(0).asType());
+                String paramName = params.get(0).getSimpleName().toString();
+                sb.append("                    ").append(typeName).append(" ").append(paramName)
+                        .append(" = mapper.convertValue(args, ").append(typeName).append(".class);\n");
```
jdubois: Single-record tools are broken — the schema and the handler disagree. `generateSchemaWithParamMetadata` nests the record under the parameter name (`properties: { req: { ...fields... } }, required: [req]`), but here the handler does `mapper.convertValue(args, SearchArgs.class)` against the *whole* args map. So when the model sends the schema-conformant `{"req":{"query":"hello","limit":7}}`, the record comes back with null/0 fields — silent data loss with the SDK's lenient mapper (`FAIL_ON_UNKNOWN_PROPERTIES=false`), or `UnrecognizedPropertyException` with a strict one. This is exactly the record pattern that `ToolDefinition`'s own Javadoc recommends, and no fixture exercises it so it's green in CI.

I verified this by running the real processor: schema `{properties={req={...}}, required=[req]}`, payload `{req:{query=hello,limit=7}}` → handler returned `query=null, limit=0`.

Fix: flatten the schema for the single-record case so the record's components are the top-level properties (matching `convertValue(args, Record.class)`). Also, declaring the record straight from `invocation.getArguments()` avoids the secondary bug where a record param literally named `args` collides with the `Map args` local declared just above.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,636 @@
...
+                return "(String) args.get(\"" + paramName + "\")";
+            }
+            if (isBoxedNumeric(qualifiedName)) {
+                return generateBoxedNumericExtraction("args.get(\"" + paramName + "\")", qualifiedName);
+            }
+            if ("java.lang.Boolean".equals(qualifiedName)) {
+                return "(Boolean) args.get(\"" + paramName + "\")";
+            }
+            // Complex types: enums, records, POJOs
+            return "mapper.convertValue(args.get(\"" + paramName + "\"), " + qualifiedName + ".class)";
```
jdubois, line 435: Generics are erased here. `qualifiedName + ".class"` produces e.g. `mapper.convertValue(args.get("ids"), java.util.List.class)`, so a `List<UUID>` parameter deserializes to a raw `List` whose elements stay `String` (or `LinkedHashMap`) — never `UUID`. The schema is correct (`items: {type: string, format: uuid}`), but real user code doing `for (UUID id : ids)` hits a `ClassCastException`. Same for `List<MyRecord>`, `Set<Foo>`, `Map<String,Long>`, etc.

Confirmed with the real processor: payload `{ids:["...uuid..."]}` → element class came back as `java.lang.String`.

Use Jackson's `TypeReference` built from the full generic type (`type.toString()`, which yields `java.util.List<java.util.UUID>`) instead of `.class` — e.g. `mapper.convertValue(args.get("ids"), new com.fasterxml.jackson.core.type.TypeReference<java.util.List<java.util.UUID>>(){})`.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,636 @@
...
+        }
+        return "(Object) args.get(\"" + paramName + "\")";
```
jdubois, line 437: Array parameters generate code that doesn't compile. Arrays aren't primitive or `DECLARED`, so they fall through to `(Object) args.get(...)`, while the variable is declared with the real array type. For `@Param String[] names` the processor emits:

```java
java.lang.String[] names = (Object) args.get("names");
```

→ `incompatible types: java.lang.Object cannot be converted to java.lang.String[]` (confirmed by compiling the real generated output). `SchemaGenerator` already supports arrays, so any tool with an array param breaks the consumer's build entirely. Route arrays (and any non-scalar) through `mapper.convertValue(args.get("x"), new TypeReference<T>(){})` — the same fix as the generics issue covers this.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -0,0 +1,636 @@
...
+    }
+
+    private String generateDefaultLiteral(TypeMirror type, String defaultValue) {
+        if (type.getKind().isPrimitive()) {
+            switch (type.getKind()) {
+                case INT :
+                case LONG :
+                case SHORT :
+                case BYTE :
+                    return defaultValue;
```
jdubois, line 582: `defaultValue` is injected as a raw, unvalidated source token, in both the schema and the lambda. For `@Param(required=false, defaultValue="abc") int count` the processor emits `... withMeta(Map.of("type","integer"), "count", abc)` and `Object countRaw = ... : abc;` → `cannot find symbol: variable abc`, i.e. the consumer's build fails (confirmed against the real processor). Even when it compiles, types can disagree: `defaultValue="1.5"` on an `int` advertises `1.5` in the schema's `default` but truncates to `1` at runtime.

Two things here: (1) validate the default against the parameter type at compile time and emit a clear error on unparseable values (consistent with the other `@Param` validations in `process()`); (2) emit a correctly-typed literal — quote strings/chars, append `L`/`F`/`D` for long/float/double — so the schema default and the runtime value match.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
edburns: Fixed in d9b6fbb6 but not yet pushed. Will push in batch to minimize CI/CD spend.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
edburns: Fixed in 8f5635a but not yet pushed. Will push in batch to minimize CI/CD spend.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
edburns: Fixed in b3a451e2 but not yet pushed. Will push in batch to minimize CI/CD spend.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
edburns: Fixed in ebd8b6d2 but not yet pushed. Will push in batch to minimize CI/CD spend.

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
edburns: Fixed in c58fd1eb but not yet pushed. Will push in batch to minimize CI/CD spend.

edburns:
| Comment | Verdict | Evidence |
|---|---|---|
| [3480477597](https://github.com/github/copilot-sdk/pull/1792#discussion_r3480477597) | Yes | `45b539cf` adds compile-time validation for optional primitives without defaults, plus the regression test. |
| [3480477599](https://github.com/github/copilot-sdk/pull/1792#discussion_r3480477599) | Yes | `2e6aa4fc` flattens single-record schema generation and switches handler binding to direct `invocation.getArguments()` mapping; it also rejects wrapper-only `@Param` metadata. |
| [3480477601](https://github.com/github/copilot-sdk/pull/1792#discussion_r3480477601) | Yes | `31ee6e32` changes generic declared-parameter extraction to use `TypeReference`, preserving element/value types. |
| [3480477605](https://github.com/github/copilot-sdk/pull/1792#discussion_r3480477605) | Yes, with caveat | The fix is the processor change that uses the same `TypeReference` path for arrays; `92193c94` is the regression test that confirms array params now compile. |
| [3480477612](https://github.com/github/copilot-sdk/pull/1792#discussion_r3480477612) | Yes | `2096937d` validates `defaultValue` against the target type and rejects mismatched numeric defaults like `1.5` for `int`. |

PR #1821, opened 2026-06-28
fix(python): forward binary tool results in HandlePendingToolCall RPC

### LINKED ISSUE #1644: Binary tool results not passed to ExternalToolTextResultForLlm
In https://github.com/github/copilot-sdk/blob/1600b57fa901cac2f82d5152938805e61d966982/python/copilot/session.py#L1649, an ExternalToolTextResultForLlm is created to be sent to HandlePendingToolCallRequest. However, the `binary_results_for_llm` attribute from `tool_result` is not copied over, but only the text result.
I believe this could be the reason why the agent I am building is unable to process image results of tools (stating that it cannot view the image, while directly sending an image as a blob attachment works fine).

syf2211:
## Summary

When a Python tool handler returns a `ToolResult` with `binary_results_for_llm` (e.g. base64 image data) or `session_log`, those fields were dropped while building the `ExternalToolTextResultForLlm` payload in `_execute_tool_and_respond`. The model only received text fields, so image tool outputs appeared unavailable.

## Motivation

Fixes #1644. Users building agents with image-returning tools saw the model report it could not view images, even though direct blob attachments worked.

## Changes

- Add `tool_result_to_external_tool_text_result_for_llm()` in `python/copilot/tools.py` to map `ToolResult` → `ExternalToolTextResultForLlm`, including binary result type enum conversion.
- Use the helper in `CopilotSession._execute_tool_and_respond` instead of an inline constructor that omitted `binary_results_for_llm` and `session_log`.
- Add unit tests covering binary forwarding and the no-binary case.

## Tests

```bash
cd python && uv run pytest test_tools.py::TestToolResultToExternalToolTextResultForLlm -q
cd python && uv run ruff check copilot/tools.py copilot/session.py test_tools.py
```

- 2 passed
- ruff: All checks passed

## Notes

- Scope is Python SDK only; Node passes the full result object through RPC already.
- Composer-2.5 code review: APPROVE

### DIFF
--- a/go/session.go
+++ b/go/session.go
@@ -1575,6 +1575,20 @@ func (s *Session) executeToolAndRespond(requestID, toolName, toolCallID string, 
 	if result.Error != "" {
 		rpcResult.Error = &result.Error
 	}
+	if result.SessionLog != "" {
+		rpcResult.SessionLog = &result.SessionLog
+	}
+	for _, b := range result.BinaryResultsForLLM {
+		entry := rpc.ExternalToolTextResultForLlmBinaryResultsForLlm{
+			Data:     b.Data,
+			MIMEType: b.MIMEType,
+			Type:     rpc.ExternalToolTextResultForLlmBinaryResultsForLlmType(b.Type),
+		}
+		if b.Description != "" {
+			entry.Description = &b.Description
+		}
+		rpcResult.BinaryResultsForLlm = append(rpcResult.BinaryResultsForLlm, entry)
+	}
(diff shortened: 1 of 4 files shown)


SteveSandersonMS (approved):
Thanks @syf2211!

mohamadhosenzademohamad-oss:
cd python && uv run pytest test_tools.py::TestToolResultToExternalToolTextResultForLlm -q
cd python && uv run ruff check copilot/tools.py copilot/session.py test_tools.py

PR #1856, opened 2026-06-30
docs: add session limits guidance

szabta89:
## Context

The Copilot runtime session-limits feature is now part of the SDK baseline through #1865, which updated the CLI/schema dependency and exposed `sessionLimits` on session create/resume options across SDKs.

## Problem

SDK users can now set an AI Credits budget, but the documentation does not explain how to configure it, that the configured value is a soft cap, or which budget-related events applications can observe.

## Solution

This PR is a docs-only follow-up:

* Adds a session-limits feature guide with create/resume examples for TypeScript, Python, Go, .NET, Java, and Rust.
* Explains that `maxAiCredits` is an AI Credits soft cap: usage is checked after model calls return, so one response can exceed the configured value before the runtime blocks the next model call.
* Extends the streaming event reference with `session.session_limits_changed`, `session.usage_checkpoint`, `session_limits_exhausted.requested`, and `session_limits_exhausted.completed`.
* Links the new guide from the docs feature indexes.

## Validation

The branch was rebuilt on current `main` with documentation-only changes, so no generated code or SDK test files are modified.

### DIFF
--- a/docs/features/session-limits.md
+++ b/docs/features/session-limits.md
@@ -1,10 +1,10 @@
 # Session limits
 
-Session limits let an application set an AI Credits budget for a Copilot session. Use `sessionLimits` when creating or resuming a session to cap the current accounting window.
+Session limits let an application set an AI Credits budget for a Copilot session. Use `sessionLimits` when creating or resuming a session to set a soft cap for the current accounting window.
 
 ## Configure a session limit
 
-Set `maxAiCredits` to the maximum AI Credits the session can consume in its current accounting window. The SDK forwards this value to the Copilot CLI when it creates or resumes the session.
+Set `maxAiCredits` to the AI Credits soft cap for the session's current accounting window. Usage is checked after model calls return, so one response can exceed the configured value before the runtime blocks the next model call. The SDK forwards this value to the Copilot CLI when it creates or resumes the session.
 
 <details open>
 <summary><strong>TypeScript</strong></summary>
@@ -150,13 +150,13 @@ let resumed = client
 
 ## Observe budget events
 
-Applications can subscribe to session events to update UI when the limit changes or the session exhausts its budget.
+Applications can subscribe to session events to update UI when the soft cap changes or the session reaches the exhausted-budget flow.
 
 | Event type | When it is emitted | Important fields |
 |---|---|---|
 | `session.session_limits_changed` | Active session limits changed. A `null` `sessionLimits` value means no limits are active. | `sessionLimits.maxAiCredits?` |
 | `session.usage_checkpoint` | The runtime records durable aggregate usage for resume and accounting. | `totalNanoAiu`, `totalPremiumRequests?` |
-| `session_limits_exhausted.requested` | The session exhausted its configured budget and needs a user decision before continuing. | `requestId`, `maxAiCredits`, `usedAiCredits` |
+| `session_limits_exhausted.requested` | The session reached the exhausted-budget flow and needs a user decision before continuing. | `requestId`, `maxAiCredits`, `usedAiCredits` |
 | `session_limits_exhausted.completed` | The exhausted-limit prompt was resolved. | `requestId`, `response.action`, `response.additionalAiCredits?`, `response.maxAiCredits?` |
 
 Use the generated event types for the SDK language you are using. For example, TypeScript narrows by `event.type`:
@@ -172,7 +172,3 @@ session.on((event) => {
     }
 });
 ```
-
-## Current limitations
-
-The SDKs expose session limits when creating or resuming a session. They do not currently expose a high-level convenience API for changing or clearing limits on an already-running session. To apply a different limit through the public SDK surface, create or resume a session with a new `sessionLimits` value.
--- a/docs/features/session-limits.md
+++ b/docs/features/session-limits.md
@@ -1,0 +1,178 @@
+# Session limits
+
+Session limits let an application set an AI Credits budget for a Copilot session. Use `sessionLimits` when creating or resuming a session to cap the current accounting window.
+
+## Configure a session limit
+
+Set `maxAiCredits` to the maximum AI Credits the session can consume in its current accounting window. The SDK forwards this value to the Copilot CLI when it creates or resumes the session.
+
+<details open>
+<summary><strong>TypeScript</strong></summary>
+
+<!-- docs-validate: skip -->
+
+```typescript
+const session = await client.createSession({
+    onPermissionRequest: approveAll,
+    sessionLimits: {
+        maxAiCredits: 30,
+    },
+});
+
+const resumed = await client.resumeSession(session.sessionId, {
+    onPermissionRequest: approveAll,
+    sessionLimits: {
+        maxAiCredits: 30,
+    },
+});
+```
+
+</details>
+<details>
+<summary><strong>Python</strong></summary>
+
+<!-- docs-validate: skip -->
+
+```python
+session = await client.create_session(
+    on_permission_request=PermissionHandler.approve_all,
+    session_limits={
+        "max_ai_credits": 30,
+    },
+)
+
+resumed = await client.resume_session(
+    session.session_id,
+    on_permission_request=PermissionHandler.approve_all,
+    session_limits={
+        "max_ai_credits": 30,
+    },
+)
+```
+
+</details>
+<details>
+<summary><strong>Go</strong></summary>
+
+<!-- docs-validate: skip -->
+
+```go
+session, err := client.CreateSession(ctx, &copilot.SessionConfig{
+	OnPermissionRequest: copilot.PermissionHandler.ApproveAll,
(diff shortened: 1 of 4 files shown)


nodejs/src/types.ts
```diff
@@ -1578,6 +1578,26 @@ export interface MemoryConfiguration {
...
 
+/**
+ * AI credit budget configuration for a session.
+ */
+export interface SessionLimitsConfig {
+    /**
+     * Maximum AI credits available to the session.
+     */
+    maxAiCredits?: number;
+}
```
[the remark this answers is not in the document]
szabta89: Fixed by aliasing `SessionLimitsConfig` to the generated session-events type so Node has a single source of truth for this shape.

java/src/main/java/com/github/copilot/rpc/SessionLimitsConfig.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
```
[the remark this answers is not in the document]
szabta89: Fixed by switching `SessionLimitsConfig.java` to the standard header used by the Java RPC package.

dotnet/src/Generated/Rpc.cs
```diff
@@ -18,6 +18,26 @@
 
 namespace GitHub.Copilot.Rpc;
 
+/// <summary>Represents an optional RPC parameter that can distinguish omitted values from explicit null values.</summary>
```
stephentoub: The PR should not be modifying this file.

dotnet/src/Generated/Rpc.cs
```diff
@@ -4417,6 +4445,75 @@ internal sealed class WorkspacesDiffRequest
     public string SessionId { get; set; } = string.Empty;
 }
 
+/// <summary>Characters that, when typed in the composer, should trigger a `completions.request`. Empty when the session has no host-driven completions (e.g. local sessions, or a relay host that does not advertise `completionTriggerCharacters`).</summary>
```
stephentoub, line 4432: What's producing all of these changes? They seem unrelated

dotnet/src/Generated/Rpc.cs
```diff
@@ -6826,6 +6923,10 @@ internal sealed class SessionUpdateOptionsParams
     [JsonPropertyName("eventsLogDirectory")]
     public string? EventsLogDirectory { get; set; }
 
+    /// <summary>Built-in subagent names to exclude from this session. Excluded built-ins are hidden from agent discovery and cannot be dispatched unless a custom agent with the same name is available.</summary>
+    [JsonPropertyName("excludedBuiltinAgents")]
+    public IList<string>? ExcludedBuiltinAgents { get; set; }
```
stephentoub, line 6912: Where did this come from? This doesn't look related to session limits?

dotnet/src/Session.cs
```diff
@@ -150,6 +150,11 @@ public SessionCapabilities Capabilities
     /// </remarks>
     public ISessionUiApi Ui => field ?? Interlocked.CompareExchange(ref field, new SessionUiApiImpl(this), null) ?? field;
 
+    /// <summary>
+    /// Gets the session limits API for updating this session's AI credit budget.
+    /// </summary>
+    public ISessionLimitsApi Limits => field ?? Interlocked.CompareExchange(ref field, new SessionLimitsApiImpl(this), null) ?? field;
```
stephentoub: Why is this needed?

dotnet/src/Types.cs
```diff
@@ -1099,6 +1099,19 @@ public interface ISessionUiApi
...
+/// </summary>
+public interface ISessionLimitsApi
+{
+    /// <summary>
+    /// Sets session limits for subsequent agent turns.
+    /// </summary>
+    /// <param name="sessionLimits">The new session limits, or null to clear the current limits.</param>
+    /// <param name="cancellationToken">Optional cancellation token.</param>
+    Task SetLimitsAsync(SessionLimitsConfig? sessionLimits, CancellationToken cancellationToken = default);
+}
```
stephentoub: Can we just rely on the RPC surface area for this, or it's important that this exists? i.e. someone can already do:
session.Rpc.Options.UpdateAsync(sessionLimits: ...)
right?

stephentoub (changes requested):
Thanks, @szabta89. I left some comments on the .NET implementation, but they extend to all six. Essentially:
- This PR shouldn't be touching the autogenerated rpc and session event files
- The main contribution I think should be the additional properties on the session creation/resume options bags.
- I question whether we actually need new "high-level" surface area for this (e.g. session.Limits.SetLimitsAsync) or for modification intrasession can just rely on the RPC surface area (e.g. session.Rpc.Options.UpdateAsync).

szabta89:
> Thanks, @szabta89. I left some comments on the .NET implementation, but they extend to all six. Essentially:
> 
>     * This PR shouldn't be touching the autogenerated rpc and session event files
> 
>     * The main contribution I think should be the additional properties on the session creation/resume options bags.
> 
>     * I question whether we actually need new "high-level" surface area for this (e.g. session.Limits.SetLimitsAsync) or for modification intrasession can just rely on the RPC surface area (e.g. session.Rpc.Options.UpdateAsync).

Thanks, I am gonna take a look at these. You mentioned a couple seemingly unrelated things ending up in this PR. They seem to be collateral schema-regeneration output from `@github/copilot@1.0.66`. What is the right approach to handle these? That schema includes session limits, but also other runtime schema additions such as the session.completions API and adaptive-thinking support. So these are unrelated to session limits semantically, but they came from the same schema refresh.

stephentoub:
> What is the right approach to handle these?

I'll get 1.0.66 imported now. (I thought I already did, but apparently not.)

szabta89:
Update on this PR: I reset the branch onto current `main` after #1865 landed, so the previous generated-code and high-level API changes are no longer part of this diff. The PR is now intentionally scoped to a docs/tests follow-up only:

* Adds a session-limits feature guide with create/resume examples across SDKs.
* Documents the generated session-limit budget events in the streaming events reference.
* Adds focused Node and Go coverage for the generated budget event surface.

I resolved the old review threads because they were all attached to outdated code that was removed by the reset.

docs/features/session-limits.md
```diff
@@ -0,0 +1,178 @@
...
+
+```go
+session, err := client.CreateSession(ctx, &copilot.SessionConfig{
+	OnPermissionRequest: copilot.PermissionHandler.ApproveAll,
+	SessionLimits: &rpc.SessionLimitsConfig{
+		MaxAiCredits: copilot.Float64(30),
+	},
+})
+
+resumed, err := client.ResumeSessionWithOptions(ctx, session.SessionID, &copilot.ResumeSessionConfig{
```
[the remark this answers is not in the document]
szabta89: Good catch — updated the Go example to use `client.ResumeSession(...)` so it matches the rest of the docs.

szabta89:
Removed the Node and Go test changes and rebuilt the branch on current `main` as a docs-only PR. The remaining diff is limited to the session-limits guide, docs index links, and streaming event reference updates.

docs/features/session-limits.md
```diff
@@ -0,0 +1,178 @@
+# Session limits
+
+Session limits let an application set an AI Credits budget for a Copilot session. Use `sessionLimits` when creating or resuming a session to cap the current accounting window.
+
+## Configure a session limit
+
```
stephentoub, line 6: Is this a hard limit? Or is it possible more AI credits than this may be consumed?

docs/features/session-limits.md
```diff
@@ -0,0 +1,178 @@
...
+            maxAiCredits: event.data.maxAiCredits,
+            usedAiCredits: event.data.usedAiCredits,
+        });
+    }
+});
+```
+
+## Current limitations
+
+The SDKs expose session limits when creating or resuming a session. They do not currently expose a high-level convenience API for changing or clearing limits on an already-running session. To apply a different limit through the public SDK surface, create or resume a session with a new `sessionLimits` value.
```
stephentoub: The APIs for configuring it are available, it's a one-liner using session.Rpc.Options.Update.

I would delete this section.

stephentoub (approved):
Left a few comments to be addressed, otherwise LGTM.

docs/features/session-limits.md
szabta89: Good point — updated this to call it an AI Credits soft cap and explain the caveat: usage is checked after model calls return, so one response can exceed the configured value before the runtime blocks the next model call.

docs/features/session-limits.md
szabta89: Agreed — deleted the section. It was misleading because in-session updates are already available through the existing `session.Rpc.Options.Update` surface.

PR #1864, opened 2026-07-01
Add opaque `metadata` passthrough to SDK tool definitions

belaltaher8:
Add an optional `_metadata` bag to the `Tool` definition (and `defineTool`) and forward it verbatim over the session.create/resume RPC. This lets hosts attach namespaced, opaque metadata to tools without expanding the typed public contract; unknown keys are round-tripped untouched.


nodejs/src/types.ts
```diff
@@ -560,6 +560,15 @@ export interface Tool<TArgs = unknown> {
...
     defer?: "auto" | "never";
+    /**
+     * Opaque, host-defined metadata associated with the tool definition.
+     *
+     * Keys are namespaced and are not part of the stable public API. The SDK
+     * does not interpret these values; it forwards them verbatim to the runtime,
+     * which may recognize specific namespaced keys to inform host-specific
+     * behavior. Unknown keys are preserved and round-tripped untouched.
+     */
+    _meta?: Record<string, unknown>;
```
stephentoub: I'd prefer not to adopt MCP's naming here. We could just use "metadata".

We will also need similar support added across all the 6 SDKs.

belaltaher8: pushed up changes to add to the other languages and changed the field name!

dotnet/src/CopilotTool.cs
```diff
@@ -151,6 +159,16 @@ public sealed class CopilotToolOptions
     /// SDK forwards it to the CLI as the tool's <c>defer</c> mode. Defaults to "auto".
     /// </remarks>
     public CopilotToolDefer? Defer { get; set; }
+
+    /// <summary>
+    /// Gets or sets opaque, host-defined metadata associated with the tool definition.
+    /// </summary>
+    /// <remarks>
+    /// Keys are namespaced and not part of the stable public API. When set, the SDK forwards
+    /// the metadata verbatim to the CLI as the tool's <c>metadata</c> object, which the runtime
```
SteveSandersonMS: Nitpick: please call it "runtime" not "CLI" since we're decoupling from CLI

SteveSandersonMS: Also you don't need to distinguish the SDK from the runtime here, so probably don't even need to mention that at all. From the consumer's point of view, this is a library they are using and the way it's split into SDK-vs-runtime is an implementation detail that's not relevant to them here.

Altogether you can just remove the `<remarks>` section.

go/types.go
```diff
@@ -1269,6 +1269,12 @@ type Tool struct {
 	// Defer controls whether the tool may be deferred (loaded lazily via tool
 	// search) rather than always pre-loaded. When empty, the runtime decides.
 	Defer ToolDefer `json:"defer,omitempty"`
+	// Metadata is opaque, host-defined metadata associated with the tool
+	// definition. Keys are namespaced and not part of the stable public API;
+	// the SDK forwards them verbatim to the runtime, which may recognize
+	// specific keys to inform host-specific behavior. Unknown keys are
+	// preserved and round-tripped untouched.
```
SteveSandersonMS: Same comment about not mentioning runtime here as if it's a separate thing

SteveSandersonMS: There are other places in this PR too but I'll not comment on the others - I'm sure you can find them :)

java/src/main/java/com/github/copilot/rpc/ToolDefinition.java
```diff
@@ -83,7 +87,8 @@
 public record ToolDefinition(@JsonProperty("name") String name, @JsonProperty("description") String description,
         @JsonProperty("parameters") Object parameters, @JsonIgnore ToolHandler handler,
         @JsonProperty("overridesBuiltInTool") Boolean overridesBuiltInTool,
-        @JsonProperty("skipPermission") Boolean skipPermission, @JsonProperty("defer") ToolDefer defer) {
+        @JsonProperty("skipPermission") Boolean skipPermission, @JsonProperty("defer") ToolDefer defer,
+        @JsonProperty("metadata") Map<String, Object> metadata) {
```
SteveSandersonMS, line 90: Seems like this would be a breaking change.

@edburns What's the recommended way to handle this in Java? Should this be added purely as a setter and not a constructor parameter? Or should it be on a new constructor overload?

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
```diff
@@ -276,7 +276,8 @@ private void writeToolDefinition(PrintWriter out, ExecutableElement method) {
         out.println("                },");
         out.println("                " + overridesArg + ",");
         out.println("                " + skipPermArg + ",");
-        out.println("                " + deferArg);
+        out.println("                " + deferArg + ",");
+        out.println("                null");
```
SteveSandersonMS: Should we also be adding "metadata" to CopilotTool and including it in the output here instead of null? Or is there no way to represent arbitrary dictionaries on a Java annotation?

dotnet/src/Client.cs
```diff
@@ -2705,17 +2705,20 @@ internal record ToolDefinition(
         JsonElement Parameters, /* JSON schema */
         bool? OverridesBuiltInTool = null,
         bool? SkipPermission = null,
-        CopilotToolDefer? Defer = null)
+        CopilotToolDefer? Defer = null,
+        [property: JsonPropertyName("metadata")] IDictionary<string, object>? Metadata = null)
```
SteveSandersonMS: I'm curious if the `[property: JsonPropertyName("metadata")]` here makes any difference. I can't see what it would do. Is it possibly redundant?

rust/src/types.rs
```diff
@@ -351,6 +351,13 @@ pub struct Tool {
     /// runtime decide.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub defer: Option<DeferMode>,
+    /// Opaque, host-defined metadata associated with the tool definition.
+    /// Keys are namespaced and not part of the stable public API; the SDK
+    /// forwards them verbatim to the runtime, which may recognize specific
+    /// keys to inform host-specific behavior. Unknown keys are preserved and
+    /// round-tripped untouched.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
```
SteveSandersonMS: I think `skip_serializing_if = "HashMap::is_empty"` might be inconsistent with the other SDKs, which would send empty dictionaries. Is there any situation where it would be important to (or not to) send an empty dictionary? My guess is it's semantically equivalent but just want to check.

dotnet/src/CopilotTool.cs
```diff
@@ -151,6 +159,16 @@ public sealed class CopilotToolOptions
...
+    /// may recognize to inform host-specific behavior. Unknown keys are preserved.
+    /// </remarks>
+    public IDictionary<string, object>? Metadata { get; set; }
```
SteveSandersonMS: I don't think we can define this type as `IDictionary<string, object>` because the `object` won't work with NativeAOT serialization.

A relatively easy fix would be to replace this with:

```suggestion
    public IDictionary<string, JsonNode?>? Metadata { get; set; }
```

dotnet/src/Client.cs
belaltaher8: yea good call it was redundant

dotnet/src/CopilotTool.cs
belaltaher8: removed the <remarks> section 👍🏼

dotnet/src/CopilotTool.cs
belaltaher8: pushed up changes to make it `JsonNode?`

go/types.go
belaltaher8: changed teh wording here to note mention runtime/SDK 👍🏼

java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java
belaltaher8: yea there's no way to represent arbitrary dictionaries like `Map<String, Object> ` on a Java annotation

rust/src/types.rs
belaltaher8: i think we're fine here. An empty metadata object carries no keys, so it decodes to absent in the other languages too. I believe that Go and Rust would `omitempty`. And the other SDKs default to null/undefined which omits

it'd be `{}` if someone passes an empty bag

java/src/main/java/com/github/copilot/rpc/ToolDefinition.java
belaltaher8: after looking into it more yea this would be a breaking change 😭 it seems like overloading the record wth multiple constructors to maintain backwards compatibility is the recommendation from googling, but ill defer to @edburns for guidance 🙇🏼

edburns: Hello @belaltaher8 , thanks for looping me in. Thankfully, even if its a breaking change, it's ok because we have the protection of the `@CopilotExperimental` annotation on some of the methods. But perhaps the impacted methods do not have that annotation on it. Will evaluate.

edburns: Thanks for doing this, @belaltaher8. From my perspective as the Java SME, these two items are the most important. I realize this may come across as a pain in the ass, but Microsoft has decided that language idiomatic SDK surface area is a key differentiator from our competition. So when we add new features, we need to "localize" them with as much fidelity as possible.

1. [ADR-005](https://github.com/github/copilot-sdk/blob/main/java/docs/adr/adr-005-tool-definition.md) gap: annotation-based tools still can’t express metadata

   The new property bag is available on `ToolDefinition`, but `@CopilotTool` users (`ToolDefinition.fromObject(...)` / `fromClass(...)`) still can’t set it. Today:

      - `com.github.copilot.tool.CopilotTool` has no metadata member.

      - `com.github.copilot.tool.CopilotToolProcessor#writeToolDefinition(...)` always emits `null` for metadata.

   So ADR-005 is functionally behind ADR-006 unless users manually post-process definitions.

   What to change:

   - `java/src/main/java/com/github/copilot/tool/CopilotTool.java`

      Add nested annotation types and a new member:

      - `MetadataEntry[] metadata() default {};`

      - nested `@interface MetadataEntry { String key(); MetadataValue value(); }`

      - nested `@interface MetadataValue { MetadataFlag[] flags() default {}; boolean bool() default false; String str() default ""; }`

      - nested `@interface MetadataFlag { String name(); boolean value(); }`

      This keeps everything in one source file and avoids adding multiple top-level annotation classes.

   - `java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java`

      In `writeToolDefinition(...)`, replace hardcoded `null` metadata emission with generated map source:

      - when `metadata().length == 0` emit `null`

      - otherwise emit `Map.of(...)` / nested `Map.of(...)` matching annotation values.

      Add helper(s) that convert `CopilotTool.MetadataEntry[]` to Java source literals.

   - `java/src/test/java/com/github/copilot/tool/CopilotToolProcessorTest.java`

      Add compile-time generation assertions for:

      - metadata present → generated `new ToolDefinition(..., ..., ..., metadataMap)` includes expected key(s) and nested flag map

      - metadata absent → generated constructor argument is `null`.

      - Docs/examples

      - Update ADR-005 example snippet (currently old ctor shape) to include metadata position or, better, use stable factory-style examples.

2. Use an overloaded constructor for compatibility

   Adding `metadata` as a new record component changed constructor arity. To preserve compatibility for existing direct constructor call sites (including previously generated ADR-005 classes), add a backward-compatible overload.

   What to change:

   - `java/src/main/java/com/github/copilot/rpc/ToolDefinition.java`

   Keep canonical 8-arg record constructor and add:

   ```java
   public ToolDefinition(String name, String description, Object parameters, ToolHandler handler,
           Boolean overridesBuiltInTool, Boolean skipPermission, ToolDefer defer) {
       this(name, description, parameters, handler, overridesBuiltInTool, skipPermission, defer, null);
   }
   ```
   

   - Keep PR’s 8-arg updates in factory methods and processor output.

   - Keep fluent copy methods carrying `metadata` through (`overridesBuiltInTool(...)`, `skipPermission(...)`, `defer(...)`) and `.metadata(...)` setter-copy method.

Testing guidance (specific):

   - `ToolDefinitionTest`

   - 7-arg overload sets `metadata == null`

   - 8-arg/copy `.metadata(...)` serializes `metadata`

   - chaining flags before/after `.metadata(...)` preserves metadata

   - `CopilotToolProcessorTest`

   - nested-annotation metadata generates expected `Map.of(...)` source

   - no metadata generates `null` argument

   - combined flags + metadata all appear in generated constructor call

   - Generated fixture coverage

   - update `$$CopilotToolMeta` fixture files to include the final constructor shape and metadata emission behavior where applicable

   - Run
   
      - `cd java && mvn verify` to cover processor tests + runtime tests end-to-end.

      - `cd java && mvn spotless:apply` to ensure formatting rules are kept.

I will paste the AI generated sketch of the new inner-annotations to enable `@CopilotTool` availability for your useful new feature next.

edburns:
```java
/*---------------------------------------------------------------------------------------------
 *  Copyright (c) Microsoft Corporation. All rights reserved.
 *--------------------------------------------------------------------------------------------*/

package com.github.copilot.tool;

import java.lang.annotation.Documented;
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

import com.github.copilot.CopilotExperimental;
import com.github.copilot.rpc.ToolDefer;

/**
 * Marks a method as a Copilot tool. The annotated method will be exposed to the
 * model as a callable tool during a session.
 */
@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.METHOD)
@CopilotExperimental
public @interface CopilotTool {

    /** Tool description (sent to the model). */
    String value();

    /** Tool name. Defaults to method name converted to snake_case. */
    String name() default "";

    /** Whether this tool overrides a built-in tool. */
    boolean overridesBuiltInTool() default false;

    /** Whether to skip permission checks. */
    boolean skipPermission() default false;

    /** Defer configuration for this tool. */
    ToolDefer defer() default ToolDefer.NONE;

    /**
     * Optional metadata bag for host-defined properties.
     *
     * Example emitted shape:
     * Map.of("github.com/copilot:safeForTelemetry",
     *        Map.of("name", true, "inputsNames", false))
     */
    MetadataEntry[] metadata() default {};

    @Documented
    @Retention(RetentionPolicy.RUNTIME)
    @Target({})
    @interface MetadataEntry {
        String key();
        MetadataValue value();
    }

    @Documented
    @Retention(RetentionPolicy.RUNTIME)
    @Target({})
    @interface MetadataValue {
        // Scalar options for simple cases
        boolean bool() default false;
        String str() default "";

        // Object-like value for structured cases
        MetadataFlag[] flags() default {};
    }

    @Documented
    @Retention(RetentionPolicy.RUNTIME)
    @Target({})
    @interface MetadataFlag {
        String name();
        boolean value();
    }
}
```

Example usage:

```java
@CopilotTool(
    value = "Reports phase",
    metadata = {
        @CopilotTool.MetadataEntry(
            key = "github.com/copilot:safeForTelemetry",
            value = @CopilotTool.MetadataValue(
                flags = {
                    @CopilotTool.MetadataFlag(name = "name", value = true),
                    @CopilotTool.MetadataFlag(name = "inputsNames", value = false)
                }
            )
        )
    }
)

public String reportPhase(String phase) {
    return phase;
}
```

edburns (changes requested):
Please see [this comment](https://github.com/github/copilot-sdk/pull/1864#discussion_r3572668490).

java/src/main/java/com/github/copilot/rpc/ToolDefinition.java
edburns: 1. ADR-005 metadata gap for annotation-based tools: implemented.
`@CopilotTool` now has nested metadata annotations and `metadata()` (`java/src/main/java/com/github/copilot/tool/CopilotTool.java:75-135`).
`CopilotToolProcessor#writeToolDefinition` now emits metadata from annotations and emits `null` when absent (`java/src/main/java/com/github/copilot/tool/CopilotToolProcessor.java:248-299`).
ADR-005 was updated with the new constructor shape and metadata examples (`java/docs/adr/adr-005-tool-definition.md:184-221`).
Processor tests were added for nested metadata and metadata+flags (`java/src/test/java/com/github/copilot/tool/CopilotToolProcessorTest.java:216-293`).
2. Backward-compatible constructor overload in `ToolDefinition`: implemented.
The 7-arg overload delegating to the 8-arg canonical constructor is present (`java/src/main/java/com/github/copilot/rpc/ToolDefinition.java:117-120`).
Fluent copy methods preserve metadata and `.metadata(...)` exists (`ToolDefinition.java:308-353`).
Runtime tests for overload behavior, metadata serialization, and chaining are present (`java/src/test/java/com/github/copilot/ToolDefinitionTest.java:87-130`).

Please see https://github.com/belaltaher8/copilot-sdk/pull/1 for one very small additional assertion to add to the test.

edburns (changes requested):
Please see https://github.com/github/copilot-sdk/pull/1864#discussion_r3583492877
Do that and I'll give the green checkmark. Thanks.

edburns (approved):

java/src/main/java/com/github/copilot/rpc/ToolDefinition.java
belaltaher8: pushed up changes to address this!

java/src/main/java/com/github/copilot/rpc/ToolDefinition.java
```diff
@@ -83,7 +86,38 @@
...
+
+    /**
+     * Backward-compatible constructor without the {@code metadata} component.
```
SteveSandersonMS: We shouldn't have phrases like "Backward-compatible" in docs because that will be meaningless in the future. It only means something in the context of this PR and won't mean anything after the PR is merged.

Please make sure docs, code comments, and everything else refers to the overall state of the code, not a moment in time.

SteveSandersonMS (approved):
One minor comment about terminology in a comment (trivial fix), but beside that this looks great. Thanks!

java/src/main/java/com/github/copilot/rpc/ToolDefinition.java
belaltaher8: pushed up changes fixing this!

rust/src/types.rs
```diff
@@ -471,6 +477,13 @@ impl Tool {
         self
     }
 
+    /// Set opaque, host-defined metadata for the tool. Keys are namespaced and
+    /// not part of the stable public API. Replaces any previously-set metadata.
+    pub fn with_metadata(mut self, metadata: HashMap<String, Value>) -> Self {
```
stephentoub: Is HashMap the right data type here? That's going to serialize in a random order. Is that ok?

belaltaher8: good call, i think its a minor concern but the other languages do guarantee an order either by built in alphabetical serialization or insertion order 👍🏼 changed this to be `IndexMap`

stephentoub (approved):

PR #1899, opened 2026-07-03
forward CustomAgentsLocalOnly in session.create and session.resume

### LINKED ISSUE #1888: CustomAgentsLocalOnly - Incorrect Behaviour
The `SessionConfig.CustomAgentsLocalOnly` property's summary tag reads:
```cs
    /// <summary>
    /// When <see langword="true"/>, custom-agent discovery is restricted to the
    /// session's local working directory (no organisation-level discovery).
    /// When <see langword="null"/>, the SDK chooses based on
    /// <see cref="CopilotClientOptions.Mode"/>: <c>true</c> under
    /// <see cref="CopilotClientMode.Empty"/>, <c>null</c> otherwise.
    /// </summary>
    public bool? CustomAgentsLocalOnly { get; set; }
```

Using this simple example, you can discover it's not this the case. All agents at all organisation levels will be loaded. Which is counter to:
"Custom instruction files (.github/copilot-instructions.md, AGENTS.md, etc. are always loaded from the working directory regardless of this setting"

```cs
using GitHub.Copilot;

// path to a sub .github folder containing an agents folder like "myworkspace\subfolder\.github\agents"
// and a higher organisation .github folder also with an agents folder like in the example below:
//
//     myworkspace\.github\agents
//     myworkspace\subfolder\.github\agents
const string WorkingDirectory = @"C:\some-subfolder\";

await using var session = await client.CreateSessionAsync(new SessionConfig
{
    WorkingDirectory = WorkingDirectory,
    EnableConfigDiscovery = true, // Note that this has to be true to get any agents to load from disk - (see https://github.com/github/copilot-sdk/issues/1887)
    CustomAgentsLocalOnly = true, // This should prevent loading agents at the higher organisation level, but it does not.
});

var agentsList = await session.Rpc.Agent.ListAsync();

foreach (var agent in agentsList.Agents)
{
    // You'll see both agents picked up, no matter the value of CustomAgentsLocalOnly
    Console.WriteLine(agent.Name);
}
```

First agent file I was using in a `.github/agents/sample.agent.md` file:
```md
---
name: sample
description: Sample Agent
---

You are an sample agent for this repository, you're just happy to be here!
```

And second agent in a `subfolder/.github/agents/example.agent.md file:
```md
---
name: example
description: Example Agent
---

You are an example agent for this repository, you're not happy to be here!
```

The property looks to disable the "walking upwards" agent loading behaviour as defined in https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-command-reference#custom-agent-locations (I can't seem to find this documented for the SDK, but happy to be pointed towards it)

syf2211:
## Summary

Forward `CustomAgentsLocalOnly` from `SessionConfig` / `ResumeSessionConfig` into the .NET SDK's `session.create` and `session.resume` wire payloads.

## Motivation

Fixes #1888. Agent discovery runs during session initialization. The .NET SDK previously only sent `customAgentsLocalOnly` via the post-create `session.options.update` call, which arrives too late to restrict discovery to the working directory. The Go SDK already includes this field in `session.create` / `session.resume`.

## Changes

- Add `customAgentsLocalOnly` to internal `CreateSessionRequest` and `ResumeSessionRequest` records
- Populate the field from `config.CustomAgentsLocalOnly` in `CreateSessionAsync` and `ResumeSessionAsync`
- Add unit serialization tests (serialize when set, omit when null)
- Add E2E wire-capture tests for create and resume requests

## Tests

- `dotnet test test/GitHub.Copilot.SDK.Test.csproj --filter FullyQualifiedName~GitHub.Copilot.Test.Unit` — **155 passed**
- New serialization tests for `customAgentsLocalOnly` — **passed**
- E2E wire-capture tests added (same harness as existing `ClientOptionsE2ETests`; require `tsx` in CI)

## Notes

- Post-create `session.options.update` behavior is unchanged; this only ensures the value is present at initialization time.
- Pre-submit review (composer-2.5): **APPROVE 4/5** — minimal fix matching Go SDK behavior for explicitly configured values.

Fixes #1888


roji:
Thanks. I've pushed an additional commit on top that propagates this to all other language SDKs and also fixes the CustomAgentsLocalOnly  behavior in Empty mode.

roji:
@github/copilot-sdk can someone give this a look please?

SteveSandersonMS (approved):
Fix is correct, complete, and consistent across all 6 SDK languages. Verified via manual regression testing (Python, Node.js, Rust) and .NET build. The timing bug — `customAgentsLocalOnly` arriving too late via post-create `session.options.update` — is resolved by forwarding the field in the `session.create`/`session.resume` wire payloads, matching the existing Go SDK behavior. Rebased onto latest main.

PR #1916, opened 2026-07-06
Remove P2 installation, use simple retry instead

edburns:


edburns (commented):
Self review of Copilot generated code. Approved.

PR #1928, opened 2026-07-07
Draft: use-after-free guards

SteveSandersonMS:
## Status

There's a shutdown race that affects .NET+Windows+inproc specifically. This does not seem likely to affect any real application, since real applications typically start up one runtime instance and then only shut it down when the application itself is shutting down. Our E2E tests, on the other hand, start up many hundreds of runtime instances in the same process in quick succession, rapidly creating and destroying them.

This PR adds a set of additional guards in the .NET FFI transport to detect or block use-after-free in both inbound and outbound message cases. But it's still not enough - there's a remaining use-after-free (or similar) in the Rust->napi-oop->Node bridge in certain shutdown cases. This manifests in the build error you can see for this PR:

```
Error: peer is closed
    at Object.call (C:\Users\runneradmin\AppData\Local\copilot\pkg\win32-x64\1.0.69-2\napi-oop-runtime\node_modules\napi-oop-runtime\dist\index.js:68:2418)
    at Proxy.S (C:\Users\runneradmin\AppData\Local\copilot\pkg\win32-x64\1.0.69-2\napi-oop-runtime\node_modules\napi-oop-runtime\dist\index.js:68:3671)
    at t.filterSecrets (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:92:42580)
    at HC.filterSecrets (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:92:44685)
    at HC.error (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:92:46758)
    at HC.writeLog (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:93:163)
    at qae.logToLevel (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:61:1536)
    at qae.error (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:61:1719)
    at t.handleError (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:5226:625)
    at process.<anonymous> (file:///C:/Users/runneradmin/AppData/Local/copilot/pkg/win32-x64/1.0.69-2/app.js:5226:360)
Node.js v24.16.0
```

It might be that before we get to dealing with this, the problem disappears naturally (in ~2 weeks we hope not to have napi-oop any more). Once we do remove napi-oop, we should see if the .NET+Windows+inproc combination immediately becomes robust when re-enabling its CI leg and then we can simply close this draft PR.

If not, we should consider productizing one or more of the use-after-free guards/locks from this PR.

This PR also temporarily disables the net472 part of the CI leg. That is simply to help diagnose whether issues are specific to .NET Framework or not. We do not plan to remove net472 support.

### DIFF
--- a/.github/workflows/dotnet-sdk-tests.yml
+++ b/.github/workflows/dotnet-sdk-tests.yml
@@ -37,10 +37,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         transport: ["default", "inprocess"]
-        # TODO: Re-enable after fixing in-process sqlite file locking on shutdown on Windows
-        exclude:
-          - os: windows-latest
-            transport: "inprocess"
     runs-on: ${{ matrix.os }}
     defaults:
       run:
--- a/.github/workflows/dotnet-sdk-tests.yml
+++ b/.github/workflows/dotnet-sdk-tests.yml
@@ -88,4 +88,15 @@ jobs:
       - name: Run .NET SDK tests
         env:
           COPILOT_HMAC_KEY: ${{ secrets.COPILOT_DEVELOPER_CLI_INTEGRATION_HMAC_KEY }}
-        run: dotnet test --no-build -v n
+        run: |
+          # In-process (FFI) hosting is only exercised on .NET (Core). The .NET
+          # Framework (net472) target does not add coverage for the in-process
+          # transport, so on Windows restrict the inprocess leg to net8.0. net472
+          # is still covered by the windows default-transport leg.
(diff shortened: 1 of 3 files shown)


SteveSandersonMS:
I don't think this is relevant any more now the underlying runtime's napi-oop threading was fixed. We can bring it back later if needed.

PR #1933, opened 2026-07-07
Tool search configuration support

almaleksia:
Adding tool_search option to CreateSession/ResumeSession and tool_references as optional field in ToolResult.


dotnet/src/Session.cs
```diff
@@ -841,6 +848,24 @@
...
+            {
+                try
+                {
+                    var metadata = await Rpc.Tools.GetCurrentMetadataAsync();
+                    invocation.AvailableTools = metadata.Tools;
+                }
+                catch
+                {
+                    // Leave AvailableTools null on failure.
+                }
```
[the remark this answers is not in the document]
almaleksia: addressed

rust/src/types.rs
```diff
@@ -4744,6 +4810,15 @@ pub struct ToolInvocation {
...
     pub arguments: Value,
+    /// Snapshot of the session's currently initialized tools.
+    ///
+    /// The SDK populates this only when the invocation targets the built-in
+    /// tool-search tool (`tool_search_tool`), so a tool-search override can
+    /// rank/filter the live catalog — including MCP tools configured in
+    /// settings — without issuing its own RPC. Empty for every other tool
+    /// invocation. This field is not part of the wire protocol.
+    #[serde(skip)]
+    pub available_tools: Vec<CurrentToolMetadata>,
```
bhavyaus: available_tools is a non-optional Vec, whereas every other binding uses a nullable sentinel (None/null/undefined) to distinguish "not populated" from "zero tools." Consider Option<Vec<…>> for parity, matching tool_references just below.

rust/src/types.rs
```diff
@@ -4826,6 +4901,9 @@ pub struct ToolResultExpanded {
     /// Tool-specific telemetry emitted with the result.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub tool_telemetry: Option<HashMap<String, Value>>,
+    /// Names of tools returned by a tool-search tool.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tool_references: Option<Vec<String>>,
```
bhavyaus, line 5011: ToolResultExpanded isn't marked #[non_exhaustive] (unlike ToolInvocation just above), so adding tool_references here is a source-level break for any external struct-literal construction of ToolResultExpanded. Worth marking the struct  #[non_exhaustive] for parity?

nodejs/src/session.ts
```diff
@@ -121,6 +122,13 @@ export type AssistantMessageEvent = Extract<SessionEvent, { type: "assistant.mes
  * await session.disconnect();
  * ```
  */
+/**
+ * Fixed name of the runtime's built-in tool-search tool. A client can replace
+ * its behavior by registering a {@link Tool} with this exact name and
+ * `overridesBuiltInTool: true`.
+ */
+const TOOL_SEARCH_TOOL_NAME = "tool_search_tool";
```
MackinnonBuck: This const got inserted between the `CopilotSession` class JSDoc (the big `@example` block just above) and the class declaration, so TypeDoc and IDE hover now attach that example to `TOOL_SEARCH_TOOL_NAME` instead of the class. Can we move the const above the class doc comment to keep the class docs intact?

java/src/main/java/com/github/copilot/RpcHandlerDispatcher.java
```diff
@@ -162,6 +169,20 @@ private void handleToolCall(JsonRpcClient rpc, String requestId, JsonNode params
                 var invocation = new ToolInvocation().setSessionId(sessionId).setToolCallId(toolCallId)
                         .setToolName(toolName).setArguments(arguments);
 
+                // The built-in tool-search tool receives a snapshot of the session's
+                // currently initialized tools so an override can filter the live
+                // catalog without issuing its own RPC. Fetch it only for that tool to
+                // avoid a round-trip on every tool call; a failed fetch leaves the
+                // snapshot null rather than failing the tool.
+                if (TOOL_SEARCH_TOOL_NAME.equals(toolName)) {
```
MackinnonBuck: This snapshot-fetch block (and the `TOOL_SEARCH_TOOL_NAME` constant) is duplicated verbatim in `CopilotSession.java`. Java is the only SDK with the logic in two dispatch paths. They're identical today, but with no test on either they can easily drift. Worth pulling into a shared helper?

python/copilot/tools.py
```diff
@@ -49,6 +53,14 @@ class ToolInvocation:
     tool_call_id: str = ""
     tool_name: str = ""
     arguments: Any = None
+    available_tools: list[CurrentToolMetadata] | None = None
```
MackinnonBuck, line 56: `available_tools` is now a public field typed `list[CurrentToolMetadata]`, but `CurrentToolMetadata` is only imported under `TYPE_CHECKING` and isn't re-exported from the `copilot` package. A user writing a `tool_search_tool` override receives these objects at runtime but can't import the type to annotate their handler or introspect it. Can we export it from `__init__.py` and add it to `__all__` too?

java/src/main/java/com/github/copilot/rpc/ToolResultObject.java
```diff
@@ -55,7 +57,8 @@ public record ToolResultObject(@JsonProperty("resultType") String resultType,
         @JsonProperty("textResultForLlm") String textResultForLlm,
         @JsonProperty("binaryResultsForLlm") List<ToolBinaryResult> binaryResultsForLlm,
         @JsonProperty("error") String error, @JsonProperty("sessionLog") String sessionLog,
-        @JsonProperty("toolTelemetry") Map<String, Object> toolTelemetry) {
+        @JsonProperty("toolTelemetry") Map<String, Object> toolTelemetry,
+        @JsonProperty("toolReferences") List<String> toolReferences) {
```
MackinnonBuck, line 62: I believe adding this new constructor arg is a source and binary breaking change. @edburns do you have advice on how to resolve this? Maybe we keep a 6-arg compatibility constructor that delegates to this new one with `toolReferences = null`?

java/src/main/java/com/github/copilot/rpc/ToolInvocation.java
```diff
@@ -174,4 +177,35 @@ public ToolInvocation setArguments(JsonNode arguments) {
...
+     * Gets a snapshot of the session's currently initialized tools.
+     * <p>
+     * The SDK populates this only when the invocation targets the built-in
+     * tool-search tool ({@code "tool_search_tool"}), so a tool-search override can
+     * rank or filter the live catalog — including MCP tools configured in settings
+     * — without issuing its own RPC. It is {@code null} for every other tool
+     * invocation.
+     *
+     * @return the available tools snapshot, or {@code null} if not applicable
+     * @since 1.0.0
```
MackinnonBuck: Nit: `@since 1.0.0` isn't right here. It should probably be `@since 1.0.7`, which is the version of the next release. However, I see that there are many existing doc comments that also don't look right (like `@since 1.3.0`, which is not a real version). @edburns, how is `@since` usually determined? Should we do an audit of existing doc comments?

rust/tests/e2e/tool_results.rs
```diff
@@ -220,6 +220,7 @@ fn expanded(text: impl Into<String>, result_type: impl Into<String>) -> ToolResu
         session_log: None,
         error: None,
         tool_telemetry: None,
+        tool_references: None,
```
MackinnonBuck: It looks like this `tool_references: None` change is essentially the only test change in the PR. Could we add unit and E2E tests for all languages?

rust/src/types.rs
almaleksia: I can mark it non_exhaustive but it will anyway break struct-literal construction in this release as it will be forbidden. The only difference it will break once and we're free to add other fields

almaleksia: done

rust/src/types.rs
almaleksia: done!

nodejs/src/session.ts
almaleksia: oops, fixed

java/src/main/java/com/github/copilot/RpcHandlerDispatcher.java
almaleksia: extracted to CopilotSession

python/copilot/tools.py
almaleksia: done

java/src/main/java/com/github/copilot/rpc/ToolResultObject.java
almaleksia: Added 6-arg constructor for backwards-compatibility

java/src/main/java/com/github/copilot/rpc/ToolInvocation.java
almaleksia: fixed

rust/tests/e2e/tool_results.rs
almaleksia: added more tests

stephentoub (approved):

PR #1945, opened 2026-07-08
Add host-mediated GitHub API requests for extensions

jchuerva:
Extensions need a tokenless way to fetch GitHub data. This adds a host-mediated GitHub REST request surface so extension processes can request current-repository data while the host/runtime keeps authentication, repository resolution, and policy enforcement inside the trusted boundary.

## Background

The immediate motivating use case is the Security canvas flow for importing Code Scanning alerts from GitHub and converting them into the existing findings JSON shape.

This API follows the existing extension boundary instead: extensions ask the session host to perform a scoped GitHub request. The extension never receives a bearer token, and the host remains responsible for choosing the account, deriving the repository from session metadata, enforcing policy, handling pagination, and normalizing errors.

The default API is intentionally narrow: **read-only** (GET), **current-repository scoped**, and **tokenless from the extension's perspective**. Broader access such as write methods, cross-repository requests, organization-wide APIs, or GraphQL should be added later behind explicit capabilities and approval rather than expanding this default surface.

## Why GET only for now

The initial surface is intentionally read-only because the motivating endpoint, Code Scanning alerts, only needs `GET`. Allowing `POST`, `PATCH`, `PUT`, or `DELETE` would let extension processes mutate GitHub state through a host-held credential, which is a much larger trust and UX boundary than read-only data access.

Write methods should be added later only with a capability model that can express which endpoints and verbs an extension needs, plus explicit host/user approval, auditing, and clear revocation behavior. Keeping the default API GET-only gives extensions a safe path to fetch data, without turning this into a general-purpose authenticated GitHub proxy.

## Design notes

The currently installed `@github/copilot` schema does not include this RPC yet, so this adds a small SDK codegen overlay and updates Rust codegen to use the shared API schema loader. That keeps Node and Rust generated outputs consistent until the upstream CLI schema includes the method directly.

## Summary

- Adds `session.api.github.request<T>(...)` to the Node extension/session API.
- Introduces a `session.api.github.request` RPC shape with `scope: "current_repository"`, `method: "GET"`, repository-relative `path`, optional `query`, and optional `paginate`.
- Rejects unsafe client input before dispatch, including non-GET methods, full URLs, `/repos/{owner}/{repo}/...`, paths without a leading slash, protocol-relative URLs, and `..` traversal segments.
- Regenerates Node and Rust RPC surfaces and updates extension docs.
- Adds focused Node tests for the wire payload, public type exports, and validation failures.

## Follow-up

The runtime/host still needs to implement `session.api.github.request` by deriving the current repository from session metadata, authenticating internally, enforcing endpoint policy, and handling pagination behavior.


SteveSandersonMS:
It looks like the underlying runtime work plan was changed and these APIs are no longer planned. Closing the SDK PR. If this later changes we can always reopen it.

PR #1952, opened 2026-07-09
GPT-4.1 is not available anymore

agoncal:
gpt-4.1 is not available in the CLI anymore. So when running the getting started sample I get:

```
com.github.copilot.JsonRpcException: Request session.create failed with message: Model "gpt-4.1" is not available
```

I changed it to gpt-5.4


scottaddie (approved):
Looks good.

Issue #1934, opened 2026-07-07
[Tracking] In-process (FFI) items to be cleaned up


SteveSandersonMS:
## Summary

Several `CopilotClientOptions` are implemented by lowering them to **environment variables** set on the spawned CLI child process. With the new **in-process (FFI) transport** (`RuntimeConnection.ForInProcess()` / `Transport::InProcess` / `forInProcess()`), there is no child process the SDK controls the environment of for these values, so **these options are not honored in-process**. This affects **all** SDKs adding the in-process transport (.NET already merged; TypeScript and Rust in progress on `stevesa/ffi-inproc-rust-ts`).

This issue tracks the gap and the fact that the **current workaround is to set the corresponding environment variable on the host/parent process yourself** before constructing the client.

## Why it happens

For stdio/TCP, the SDK spawns a child process and populates `startInfo.Environment` (or the equivalent), so per-client values are coherent — each client owns its own OS process.

For in-process, the native runtime cdylib is loaded into the **shared host process**, and host-side native reads consult the **ambient host process environment**, not a per-client env block. .NET's `ValidateEnvironmentOptions` therefore *rejects* `Environment`/`Telemetry` for in-process, and the in-process start path passes an effectively **empty** env to `host_start` — so the remaining env-lowered options are simply never applied.

## Complete list of options lowered to env vars (from `dotnet/src/Client.cs`)

| Option | Env var(s) | In-process behavior today |
| --- | --- | --- |
| `Environment` | (replaces whole env block) | **Rejected** (throws) |
| `Telemetry` | `COPILOT_OTEL_ENABLED`, `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_PROTOCOL`, `COPILOT_OTEL_FILE_EXPORTER_PATH`, `COPILOT_OTEL_EXPORTER_TYPE`, `COPILOT_OTEL_SOURCE_NAME`, `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` | **Rejected** (throws) |
| `GitHubToken` | `COPILOT_SDK_AUTH_TOKEN` (+ `--auth-token-env` CLI arg) | **Silently ignored** |
| `BaseDirectory` | `COPILOT_HOME` | **Silently ignored** |
| `Mode == Empty` | `COPILOT_DISABLE_KEYTAR=1` | **Silently ignored** |
[...]

PR #1953, opened 2026-07-09
Add in-process (FFI) transport for the Node.js SDK

SteveSandersonMS:
## Summary

Adds an **in-process (FFI) transport** for the Node.js SDK. Instead of spawning the Copilot runtime as a child process, the SDK can host it in-process by loading the native runtime library and speaking JSON-RPC over its C ABI via [`koffi`](https://koffi.dev/).

This is the **Node-only split** of the combined Node+Rust in-process FFI work (`stevesa/ffi-inproc-rust-ts`); the Rust SDK changes will follow in a separate PR.

## What's included

- `RuntimeConnection.forInProcess()` + `InProcessRuntimeConnection` (experimental)
- `FfiRuntimeHost`: koffi-based C ABI host with deferred, non-reentrant inbound frame delivery and an async broker pump so foreign-thread outbound callbacks are serviced while the loop is idle
- `client.ts` wiring for `start`/`stop`/`forceStop` over the FFI host
- E2E harness support (`isInProcessTransport`, per-test proxy/env mirroring), a dedicated `inprocess_ffi` E2E test, and an `inprocess` CI matrix cell
- `skipIf(isInProcessTransport)` for tests that require per-client env or a child process (telemetry, request-cancel, checkpoint u32, process-kill), documented against #1934
- Graceful `stop()` (not `forceStop`) for inproc resume clients so the runtime closes `session-store.db` and Windows temp-dir cleanup can unlink it
- Bump `@github/copilot` to `^1.0.70-0`; add `koffi ^3.1.0`

## Excluded (temporary scaffolding)

All env-gated diagnostic tracing (`COPILOT_FFI_TRACE`, `COPILOT_EVENT_TRACE`) and the koffi-callback repro test + its CI job have been stripped out — this branch contains only the production feature.

## Testing

- `typecheck` ✅ · `eslint` ✅
- Inproc E2E smoke: `streaming_fidelity` + `inprocess_ffi` → 6/6 ✅
- Full `default` + `inprocess` transport matrix runs in CI

### DIFF
--- a/nodejs/src/client.ts
+++ b/nodejs/src/client.ts
@@ -914,8 +914,12 @@ export class CopilotClient {
         // terminating a child and the file stays locked (Windows), preventing
         // removal of the session-state directory. Aborting lets the turn cancel
         // and release the handle. Best-effort and idempotent: a session with no
-        // active turn is a no-op.
-        await Promise.allSettled(activeSessions.map((session) => session.abort()));
+        // active turn is a no-op. Skip for external servers: we don't own that
+        // runtime, and aborting would cancel pending work other clients may
+        // still resume.
+        if (!this.isExternalServer) {
+            await Promise.allSettled(activeSessions.map((session) => session.abort()));
+        }
         for (const session of activeSessions) {
             const sessionId = session.sessionId;
             let lastError: Error | null = null;
--- a/nodejs/src/client.ts
+++ b/nodejs/src/client.ts
@@ -907,17 +907,18 @@ export class CopilotClient {
 
         // Disconnect all active sessions with retry logic
         const activeSessions = [...this.sessions.values()];
-        // Abort any in-flight turn before teardown. A turn still running when
-        // the runtime disposes the session can leave that session's SQLite
-        // session.db handle open in the runtime; over the in-process transport
-        // the runtime shares this process, so the handle is not reclaimed by
-        // terminating a child and the file stays locked (Windows), preventing
-        // removal of the session-state directory. Aborting lets the turn cancel
-        // and release the handle. Best-effort and idempotent: a session with no
-        // active turn is a no-op. Skip for external servers: we don't own that
-        // runtime, and aborting would cancel pending work other clients may
-        // still resume.
-        if (!this.isExternalServer) {
+        // TEMPORARY: over the in-process (FFI) transport the runtime shares this
+        // process, so a turn still running when the runtime disposes the session
+        // can leave that session's SQLite session.db handle open — it isn't
+        // reclaimed by terminating a child process, so the file stays locked
+        // (Windows) and the session-state directory can't be removed. Abort any
+        // in-flight turn first so it cancels and releases the handle. Best-effort
+        // and idempotent: a session with no active turn is a no-op. Scoped to
+        // in-process only: stdio/tcp runtimes run in a child process that we kill
+        // on shutdown (which frees the handle), and for external servers we don't
+        // own the runtime and aborting would cancel pending work other clients
+        // may still resume. Remove once the runtime cleans up fully on shutdown.
+        if (this.connectionConfig.kind === "inprocess") {
             await Promise.allSettled(activeSessions.map((session) => session.abort()));
         }
         for (const session of activeSessions) {
--- a/nodejs/src/client.ts
+++ b/nodejs/src/client.ts
@@ -40,6 +40,7 @@ import type {
 } from "./generated/rpc.js";
 import { getSdkProtocolVersion } from "./sdkProtocolVersion.js";
 import { CopilotSession } from "./session.js";
+import { FfiRuntimeHost } from "./ffiRuntimeHost.js";
 import { createSessionFsAdapter, type SessionFsProvider } from "./sessionFsProvider.js";
 import { createCopilotRequestAdapter } from "./copilotRequestHandler.js";
 import type { CopilotRequestHandler } from "./copilotRequestHandler.js";
@@ -58,6 +59,7 @@ import type {
     BearerTokenProvider,
     GetStatusResponse,
     InternalRuntimeConnection,
+    RuntimeConnection,
     LargeToolOutputConfig,
     MCPServerConfig,
     ModelInfo,
@@ -473,6 +475,7 @@ class TeardownResilientStreamMessageWriter extends StreamMessageWriter {
 export class CopilotClient {
     private cliStartTimeout: ReturnType<typeof setTimeout> | null = null;
     private cliProcess: ChildProcess | null = null;
+    private ffiHost: FfiRuntimeHost | null = null;
     private connection: MessageConnection | null = null;
     private messageWriter: TeardownResilientStreamMessageWriter | null = null;
     private socket: Socket | null = null;
@@ -564,6 +567,33 @@ export class CopilotClient {
     }
 
     /**
+     * Environment variable that overrides the transport when the caller does not set
+     * {@link CopilotClientOptions.connection}. Accepts `"inprocess"` or `"stdio"`
+     * (case-insensitive); unset preserves the default stdio transport. Any other value
+     * is an error.
+     */
+    private static readonly DEFAULT_CONNECTION_ENV_VAR = "COPILOT_SDK_DEFAULT_CONNECTION";
+
+    /**
+     * Resolves the default {@link RuntimeConnection} for the no-connection case,
+     * honoring {@link CopilotClient.DEFAULT_CONNECTION_ENV_VAR}.
+     */
+    private static resolveDefaultConnection(options: CopilotClientOptions): RuntimeConnection {
+        const env = options.env ?? process.env;
+        const value = env[CopilotClient.DEFAULT_CONNECTION_ENV_VAR];
+        if (!value || value.toLowerCase() === "stdio") {
+            return { kind: "stdio" };
+        }
+        if (value.toLowerCase() === "inprocess") {
+            return { kind: "inprocess" };
+        }
+        throw new Error(
+            `Invalid ${CopilotClient.DEFAULT_CONNECTION_ENV_VAR} value '${value}'. ` +
+                `Expected 'inprocess', 'stdio', or unset.`
+        );
+    }
+
+    /**
      * Creates a new CopilotClient instance.
      *
      * @param options - Configuration options for the client
@@ -594,8 +624,10 @@ export class CopilotClient {
         // Resolve the connection mode. `_internalConnection` is set by
         // `joinSession()` to opt into the parent-process stdio path; consumers
         // should always go through the public `connection` field.
-        const conn: InternalRuntimeConnection = options._internalConnection ??
-            options.connection ?? { kind: "stdio" };
+        const conn: InternalRuntimeConnection =
+            options._internalConnection ??
+            options.connection ??
+            CopilotClient.resolveDefaultConnection(options);
 
         if (
             conn.kind === "uri" &&
@@ -810,7 +842,9 @@ export class CopilotClient {
 
         try {
             // Only start CLI server process if not connecting to external server
-            if (!this.isExternalServer) {
+            if (this.connectionConfig.kind === "inprocess") {
+                await this.startInProcessFfi();
+            } else if (!this.isExternalServer) {
                 await this.startCLIServer();
             }
 
@@ -910,7 +944,7 @@ export class CopilotClient {
         // Ask SDK-owned runtimes to flush and clean up before we tear down
         // their transport/process. External runtimes may be shared, so only
         // close our connection to them.
-        if (this.connection && this.cliProcess && !this.isExternalServer) {
+        if (this.connection && (this.cliProcess || this.ffiHost) && !this.isExternalServer) {
             const runtimeShutdownStart = Date.now();
             const shutdownPromise = this.rpc.runtime.shutdown();
             void shutdownPromise.catch(() => undefined);
@@ -1007,6 +1041,21 @@ export class CopilotClient {
                 errors.push(
                     new Error(
                         `Failed to kill CLI process: ${error instanceof Error ? error.message : String(error)}`
+                    )
+                );
+            }
+        }
+        // Tear down the in-process FFI host (closes the native connection and
+        // shuts down the native runtime host) for SDK-owned in-process runtimes.
+        if (this.ffiHost) {
+            const host = this.ffiHost;
+            this.ffiHost = null;
+            try {
+                host.dispose();
+            } catch (error) {
+                errors.push(
+                    new Error(
+                        `Failed to dispose in-process runtime host: ${error instanceof Error ? error.message : String(error)}`
                     )
                 );
             }
@@ -1111,6 +1160,16 @@ export class CopilotClient {
                 // Ignore errors
             }
             this.cliProcess = null;
+        }
+
+        // Tear down the in-process FFI host (if any).
+        if (this.ffiHost) {
+            try {
+                this.ffiHost.dispose();
+            } catch {
+                // Ignore errors during force stop
+            }
+            this.ffiHost = null;
         }
 
         if (this.cliStartTimeout) {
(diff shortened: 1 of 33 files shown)


nodejs/src/ffiRuntimeHost.ts
```diff
@@ -0,0 +1,436 @@
...
+     * SDK goes idle mid-turn (e.g. the worker is awaiting a model HTTP response, so
+     * there are no client→server writes) the broker parks and the next outbound frame
+     * (the model response, then everything after it) sits undelivered until the next
+     * koffi call — observed as a permanent 30s stall of otherwise-healthy sessions on
+     * macOS/Windows (libuv services the broker differently on Linux, which is why it
+     * only reproduces off-Linux). Keeping exactly one cheap async FFI call
+     * (`log_dropped_count`) continuously in flight keeps the broker pumping so
+     * foreign-thread frames are always delivered. This is the koffi analogue of the
+     * .NET host's raw function-pointer callback, which runs directly on the runtime
+     * thread and needs no event-loop pumping.
```
SteveSandersonMS: Not sure this is all needed now the runtime threading was fixed. Need to revalidate.

nodejs/src/ffiRuntimeHost.ts
```diff
@@ -0,0 +1,436 @@
...
+        private readonly libraryPath: string,
+        private readonly cliEntrypoint: string,
+        private readonly environment?: Record<string, string | undefined>,
+        private readonly workingDirectory?: string
+    ) {
+        this.lib = loadLibrary(libraryPath);
+        this.receiveStream = new PassThrough();
+        this.sendStream = new Writable({
+            // Write frames with a synchronous FFI call (restored while diagnosing the
+            // macOS/Windows inbound-callback stall; async writes were a no-op for it).
```
SteveSandersonMS: Do not add comments about how the PR was built. The comments must be written for future readers. Also anything related to the inbound-callback stall must be revalidated now as it may no longer apply given the runtime-side fixes.

nodejs/src/ffiRuntimeHost.ts
```diff
@@ -0,0 +1,436 @@
...
+
+        this.startBrokerPump();
+    }
+
+    /**
+     * Keeps exactly one cheap async FFI call in flight at all times so koffi's
+     * threadsafe-function broker stays serviced and foreign-thread outbound callbacks
+     * are delivered without waiting for the next client→server write. Each call
+     * reschedules the next on completion; the in-flight libuv request also keeps the
+     * event loop alive (replacing the old keep-alive timer). Stops when disposed.
```
SteveSandersonMS: Again I'm not sure this is still needed. Revalidate.

nodejs/src/ffiRuntimeHost.ts
```diff
@@ -0,0 +1,436 @@
...
+     *
+     * The native pointer is only valid for the duration of this call, so the bytes are
+     * decoded/copied eagerly; but writing them to {@link receiveStream} (which
+     * synchronously drives frame parsing and JSON-RPC dispatch, and may re-enter the
+     * SDK's write path) is deferred to a `setImmediate` drain on a clean stack. This
+     * keeps the callback minimal and non-reentrant — the koffi analogue of the .NET
+     * host's thread-safe `Channel` callback, which enqueues and returns immediately.
+     * Delivering synchronously here could re-enter a native `connection_write` call
+     * still on the stack and deadlock (observed as hung `session.rpc.*` round-trips
+     * on macOS).
```
SteveSandersonMS: Revalidate - the underlying issues may be gone.

nodejs/src/ffiRuntimeHost.ts
```diff
@@ -0,0 +1,436 @@
...
+        // koffi delivers outbound callbacks from a secondary thread by queuing them onto
+        // the JS event loop; at teardown one such delivery can still be queued after we
+        // stop the native side. Unregistering while koffi still has a queued call makes
+        // koffi invoke a torn-down callback and raise inside its own native code — an
+        // uncaught Node-API callback exception (DEP0168) that no JS try/catch can catch.
+        // The native connection/host are already closed by the time this runs (see
+        // dispose), so no new deliveries originate; a queued delivery fires in libuv's
+        // poll phase and setImmediate (check phase) runs right after it in the same loop
+        // iteration, so the pending delivery (a no-op, since we are disposed) drains
+        // before we free the slot.
```
SteveSandersonMS: Revalidate any hacks like this

nodejs/test/e2e/copilot_request_cancel_error.e2e.test.ts
```diff
@@ -162,23 +162,31 @@ describe("CopilotRequestHandler observes runtime cancellation", async () => {
...
-            await waitFor(() => handler.inferenceEntered, 60_000);
-            await session.abort();
-            await waitFor(() => handler.sawAbort, 30_000);
-        } finally {
-            await session.disconnect();
-        }
+    // A CopilotRequestHandler registers a process-wide LLM inference provider. Over the
+    // in-process transport every client shares the one host process, so a second client
+    // cannot register while another still holds the provider. Covered by the default
+    // (stdio) cell, where each client owns its own process.
```
SteveSandersonMS: Could we make this test run on inproc if it's never run in parallel with other tests?

nodejs/test/e2e/inprocess_ffi.e2e.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        } catch {
+            // Ignore cleanup errors - process may already be stopped
+        }
+    });
+}
+
+describe("In-process FFI transport", () => {
+    // Mirrors the .NET `Should_Start_And_Connect_Over_InProcess_Ffi`. Resolution of the
+    // in-process transport from COPILOT_SDK_DEFAULT_CONNECTION is exercised by the full
+    // E2E suite running under the `inprocess` CI matrix cell, not a dedicated test.
```
SteveSandersonMS: Don't refer to .NET. Just document the Node code.

nodejs/test/e2e/rpc_workspace_checkpoints.e2e.test.ts
```diff
@@ -20,17 +20,25 @@ describe("Session workspace checkpoint RPC", async () => {
...
-            });
-            expect(result.content ?? "").toBe("");
-        } finally {
-            await session.disconnect();
+    // Over the in-process transport session.workspaces.readCheckpoint is answered by the
+    // native runtime, which decodes the checkpoint number as a u32 and rejects the
+    // MAX_SAFE_INTEGER sentinel this test uses. Covered by the default (stdio) cell; the
+    // in-process u32 limitation is a runtime gap tracked in
+    // https://github.com/github/copilot-sdk/issues/1934.
+    it.skipIf(isInProcessTransport)(
```
SteveSandersonMS: Can we just not use MAX_SAFE_INTEGER as a sentinel? Would be better to cover this behavior on inproc than to ignore it.

Also why doesn't MAX_SAFE_INTEGER work over inproc? I don't see any references to numbers in https://github.com/github/copilot-sdk/issues/1934.

nodejs/src/ffiRuntimeHost.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Confirmed no longer needed and removed. This was part of the same runtime-fault compensation; real turns now complete on both transports without it (if delivery still needed pumping, the e2e suites would time out at 30s per turn — they don't).

nodejs/src/ffiRuntimeHost.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Revalidated and removed. With the runtime threading fixed this reduces to standard koffi usage — no hack remains here.

nodejs/src/ffiRuntimeHost.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Revalidated — the underlying issue is gone. Removed. The callback unregister is now synchronous (`koffi.unregister` in `dispose()`), no deferral and no DEP0168 warning.

nodejs/test/e2e/inprocess_ffi.e2e.test.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Done — removed the .NET reference and rewrote the description to document the Node code only as an in-process FFI smoke test.

nodejs/test/e2e/rpc_workspace_checkpoints.e2e.test.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Both done. Replaced the sentinel with `4294967294` (a u32-safe `unknown checkpoint` value) and dropped the skip, so the test now runs on **both** transports (4/4).

On why MAX_SAFE_INTEGER failed over inproc: it's unrelated to #1934. The inproc path goes through the napi-oop wire decoder, and the `workspace_manager_read_checkpoint` #[napi] binding declares `checkpoint_number: u32` (then immediately widens with `u64::from(...)`). The decoder rejects any JSON int > `u32::MAX` before the handler runs (`wire conversion error: ... expected u32`). The stdio path has no typed narrowing (plain JS number) so it tolerated the huge value. It's a vestigial narrowing at the wire boundary, not a real checkpoint bug — real checkpoint numbers are small, so covering `unknown checkpoint` with a u32-safe value is the right fix and now exercises inproc too.

nodejs/test/e2e/copilot_request_cancel_error.e2e.test.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Reasoned about it and then verified empirically: it can't run on inproc. The runtime enforces a **single, process-wide** LLM inference provider. The sibling `Throwing` test in this file already registers a requestHandler provider (and is inproc-enabled) and holds it for the file's lifetime, so a second client here fails deterministically with `Another client is already the LLM inference provider` (reproduced on both runs). It's not a parallelism issue — vitest already runs tests within a file sequentially (`pool: forks`, `isolate: true`). So the `skipIf(isInProcessTransport)` is legitimate; I've sharpened the comment to cite the concrete error rather than hand-wave.

nodejs/src/ffiRuntimeHost.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Done. Rewrote all comments to be timeless and reader-facing (no `how the PR was built` narrative). The inbound-callback stall handling was specifically revalidated: koffi's threadsafe-function delivers the callback on a normal event-loop tick (not nested on the runtime's native stack), so there is no reentrancy to defer. The deferral is removed — `feedInbound` now writes straight to the receive stream. The 30s idle-turn stalls that motivated it were the runtime worker-starvation fault, now fixed.

nodejs/src/ffiRuntimeHost.ts
SteveSandersonMS: _🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

Revalidated against the fixed runtime (1.0.70-0). Most of this block was compensating for the old worker-starvation/threading fault and is now gone — net **-93 lines**. Removed: the broker pump, the deferred inbound delivery, and the deferred callback unregister. Kept only the synchronous frame write (already standard koffi) and a keep-alive timer (see the dedicated reply on that — it is genuinely required, not a fault workaround).

nodejs/src/ffiRuntimeHost.ts
```diff
@@ -0,0 +1,341 @@
...
+        const envJson = buildEnvJson(this.environment);
+
+        // The native host spawns the CLI worker itself and has no cwd parameter, so the
+        // worker inherits this process's cwd. Mirror the stdio child's `cwd: workingDirectory`
+        // by switching cwd for the duration of the blocking host_start, then restoring it.
+        const previousCwd = process.cwd();
+        const shouldSwitchCwd = !!this.workingDirectory && this.workingDirectory !== previousCwd;
+        if (shouldSwitchCwd) {
+            process.chdir(this.workingDirectory!);
+        }
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in e8565675. The in-process transport now rejects a `workingDirectory` in the `CopilotClient` constructor rather than mutating the shared process-global cwd — the `process.chdir()` dance is removed from `FfiRuntimeHost` entirely. Callers that need a specific working directory set it on the host process themselves before constructing the client.

_🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

nodejs/src/ffiRuntimeHost.ts
```diff
@@ -0,0 +1,341 @@
...
+     */
+    private feedInbound(bytesPtr: unknown, bytesLen: number | bigint): void {
+        // An exception thrown across the native→JS (Node-API) boundary cannot propagate
+        // and would surface only as a DEP0168 "uncaught Node-API callback exception"
+        // warning, so catch and log it here instead of letting it escape.
+        try {
+            const length = Number(bytesLen);
+            if (!bytesPtr || length <= 0) {
+                return;
+            }
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in e8565675. `feedInbound()` now returns early when `this.disposed || this.receiveStream.writableEnded`, so a native outbound callback delivered after `dispose()` ends the stream drops the late frame instead of throwing `ERR_STREAM_WRITE_AFTER_END`.

_🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

nodejs/src/client.ts
```diff
@@ -563,6 +566,33 @@ export class CopilotClient {
...
+    /**
+     * Resolves the default {@link RuntimeConnection} for the no-connection case,
+     * honoring {@link CopilotClient.DEFAULT_CONNECTION_ENV_VAR}.
+     */
+    private static resolveDefaultConnection(options: CopilotClientOptions): RuntimeConnection {
+        const env = options.env ?? process.env;
+        const value = env[CopilotClient.DEFAULT_CONNECTION_ENV_VAR];
+        if (!value || value.toLowerCase() === "stdio") {
+            return { kind: "stdio" };
+        }
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in e8565675 — and more strictly than suggested. `resolveDefaultConnection()` now reads `COPILOT_SDK_DEFAULT_CONNECTION` from `process.env` **only**, never `options.env`. `options.env` is a deprecated child-process setting that the in-process transport does not honor, so the host-process transport switch can no longer be silently disabled by a per-client env block.

_🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

nodejs/test/e2e/harness/sdkTestContext.ts
```diff
@@ -102,21 +131,53 @@ export async function createSdkTestContext({
...
+    const isInProcess = connection.kind === "inprocess";
+    const inProcessEnv: Record<string, string> = isInProcess
+        ? {
+              ...(mergedEnv as Record<string, string>),
+              GH_TOKEN: authTokenToUse,
+              GITHUB_TOKEN: authTokenToUse,
+              COPILOT_HMAC_KEY: "",
+              CAPI_HMAC_KEY: "",
+          }
+        : {};
```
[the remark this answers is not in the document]
SteveSandersonMS: Acknowledged, but not changing this. It's test-harness-only code (`sdkTestContext.ts`), and in practice the mirrored env values are concrete strings, so the `undefined` -> `"undefined"` coercion isn't hit. Not worth adding guarding to a test helper for a case the harness doesn't produce.

_🤖 Reply drafted by GitHub Copilot CLI, acting on @SteveSandersonMS's behalf._

PR #1973, opened 2026-07-13
rust: forward intercepted HTTP via hyper+native-tls to drop ring

Shivam60:
reqwest (forward_http) pins rustls/hyper-rustls optionals in Cargo.lock, dragging in ring (CG High, MVS-2022-374v-6mvc) even though only native-TLS is used and ring is never compiled. Replace reqwest with its underlying hyper+hyper-tls native-TLS stack (native-tls ALPN preserves HTTP/2). Removes ring/rustls entirely; all tests pass, clippy clean.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -18,12 +18,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.102"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
-
-[[package]]
 name = "arbitrary"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -57,9 +51,9 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bitflags"
-version = "2.11.1"
+version = "2.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
+checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
 
 [[package]]
 name = "block-buffer"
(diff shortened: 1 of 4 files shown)


stephentoub (commented):
Does this no longer respect proxy configuration, e.g. HTTP_PROXY, HTTPS_PROXY, and NO_PROXY?

SteveSandersonMS:
## Manual validation: HTTP/2 regression reproduced

I independently reproduced a functional regression in `forward_http` while reviewing this PR, by exercising the real request path against a live HTTP/2-only endpoint (`https://api.github.com`) on both branches over the same network path:

- **`main` (reqwest/native-tls):** request succeeds at the protocol level — returns a normal `403` (missing `User-Agent`), i.e. the TLS/HTTP framing negotiation was transparent.
- **This PR (hyper + hyper-tls/native-tls):** the same request returns `400 An HTTP/1.x request was sent to an HTTP/2 only endpoint.` — reproducible on every run.

**Likely cause:** setting ALPN preferences (`h2`, `http/1.1`) on the `native-tls`/`hyper-tls` connector only affects the *TLS handshake*. `hyper_util::client::legacy::Client` does not automatically switch request framing to HTTP/2 based on the connector's negotiated ALPN result — that plumbing (checking `negotiated_alpn()` on the TLS stream and driving an `http2::handshake` on that connection) is something reqwest does internally but isn't replicated here. So despite the doc comment ("native-tls ALPN preserves HTTP/2"), the new client is effectively HTTP/1.1-only in practice.

**Why this is merge-blocking:** `forward_http` is the general-purpose passthrough for intercepted/forwarded HTTP traffic (used as the default of `CopilotRequestHandler::send_request`), so this isn't a theoretical edge case — any upstream that requires HTTP/2 will now hard-fail where it worked before. Neither this PR nor the existing suite (`mcp_oauth.rs`) has any HTTP/2 coverage (the only `http2`/`h2` matches in the whole diff are in the doc comment and the ALPN string list), so nothing caught this before merge.

**Suggested paths forward:**
1. Add explicit HTTP/2 upgrade handling — inspect the negotiated ALPN protocol on the TLS stream post-handshake and drive an `http2::handshake`/HTTP/2 connection when negotiated, instead of relying on the legacy client's default HTTP/1.1 framing, **or**
2. Use a TLS backend that already handles this automatically without pulling in `ring` (e.g. `hyper-rustls` configured with the `aws-lc-rs` or `ring`-free crypto provider), which keeps hyper's automatic ALPN → HTTP/2 upgrade path intact while still resolving the Cargo.lock compliance flag.

Either way, please add an e2e regression test against a real or mocked HTTP/2-only upstream before merging, so this doesn't silently regress again.

I'm converting this to draft for now — please mark it ready for review again once the HTTP/2 handling is fixed and validated against an HTTP/2-only endpoint.

PR #1995, opened 2026-07-15
docs: expand Azure Managed Identity BYOK guidance

scottaddie:
## Summary

- Expand the Azure Managed Identity guide with multi-language samples and new bearer token provider approach introduced via https://github.com/github/copilot-sdk/pull/1748.
- Update the BYOK reference to document the bearer token provider callback.

Directly addresses Patrick's comment: https://github.com/github/copilot-sdk-internal/issues/45#issuecomment-4971913761


docs/setup/azure-managed-identity.md
```diff
@@ -1,37 +1,23 @@
-# Azure managed identity with BYOK
+# Azure Managed Identity with BYOK
```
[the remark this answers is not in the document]
scottaddie: @copilot Managed Identity is a product name and therefore should be treated as a proper noun. Update this PR with the appropriate change in https://github.com/github/copilot-sdk/blob/main/.github/instructions/docs-style.instructions.md#proper-nouns-to-always-capitalize.

.github/instructions/docs-style.instructions.md
```diff
@@ -26,7 +26,7 @@ Do not use `**bold**` or `*italic*` markers inside headings. The heading level p
 * Acronyms: MCP, BYOK, MAF, SDK, CLI, API, HMAC, CI/CD, SaaS, ISV, FAQ, LLM, AI, EMU, ID, UI, PNG
 * Tools (keep canonical casing): npm, npx, stdio
 * Code identifiers in headings: SessionConfig, MessageOptions, TelemetryConfig, ProviderConfig, CopilotClient
-* Multi-word proper names: GitHub App, GitHub Actions, GitHub OAuth, Foundry Local, Azure AD, Container Instances
+* Multi-word proper names: GitHub App, GitHub Actions, GitHub OAuth, Foundry Local, Azure AD, Managed Identity, Container Instances
```
scottaddie: @copilot Update the PR to remove "Azure AD" from this list. Microsoft Entra ID is the new product branding. See https://learn.microsoft.com/entra/fundamentals/how-to-rename-azure-ad.

docs/setup/azure-managed-identity.md
```diff
@@ -65,9 +93,33 @@ npm install @github/copilot-sdk @azure/identity
...
+### Provide an access token you manage
+
+Get a token using your language's Azure Identity SDK credential and pass it as the bearer token in your provider configuration. Entra tokens for Managed Identities are valid for 24 hours. As a recommendation, proactively refresh these tokens after 12 hours. To only refresh when necessary, see the language-specific guidance in each code sample's comments below.
+
+Using Python as an example, the flow is:
+
+1. Configure `DefaultAzureCredential` for your environment.
+1. Invoke `DefaultAzureCredential`'s `get_token_info` method to obtain a token for the `https://ai.azure.com/.default` scope.
+1. Pass the token string in `bearer_token` of the BYOK provider configuration.
+1. Refresh the token before it expires (not shown; your application implements the refresh mechanism).
```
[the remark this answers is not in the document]
scottaddie: @copilot Update the PR to implement this suggestion, but also keep the parenthetical text about the step not being shown in the diagram.

sunbrye (approved):
Mainly reviewed the structure of the content and it looks good to me ✅

docs/setup/azure-managed-identity.md
```diff
@@ -1,37 +1,23 @@
-# Azure managed identity with BYOK
+# Azure Managed Identity with BYOK
 
-The Copilot SDK's [BYOK mode](../auth/byok.md) accepts static API keys, but Azure deployments often use **Managed Identity** (Microsoft Entra ID) instead of long-lived keys. Since the SDK doesn't natively support Microsoft Entra authentication, you can use a short-lived bearer token via the `bearer_token` provider config field.
+The GitHub Copilot SDK's [BYOK mode](../auth/byok.md) supports static API keys, but Azure deployments often use **Managed Identity** (Microsoft Entra ID) instead of long-lived keys. Since the SDK doesn't natively support Microsoft Entra authentication, supply either a bearer token you already acquired or a bearer token provider that can fetch fresh tokens on demand.
```
SteveSandersonMS: Saying "SDK doesn't native support Entra" might sound like the product is deficient, but the layering we've chosen is deliberate - we've designed this to compose with Azure Identity SDK for maximum flexibility.

docs/setup/azure-managed-identity.md
```diff
@@ -1,37 +1,23 @@
...
 ## How it works
 
-Microsoft Foundry's OpenAI-compatible endpoint accepts bearer tokens from Microsoft Entra ID in place of static API keys. The pattern is:
+Microsoft Foundry's OpenAI-compatible endpoint (`https://<resource-name>.openai.azure.com/openai/v1/`) accepts bearer tokens from Microsoft Entra ID in place of static API keys. You can authenticate in either of these ways:
 
-1. Use `DefaultAzureCredential` to obtain a token for the `https://ai.azure.com/.default` scope
-1. Pass the token as the `bearer_token` in the BYOK provider config
-1. Refresh the token before it expires (tokens are typically valid for ~1 hour)
+* [Provide an access token that your application refreshes.](#provide-an-access-token-you-manage)
+* [Specify a token provider callback so the runtime can acquire tokens on demand.](#specify-a-token-provider-for-automatic-token-acquisition)
```
SteveSandersonMS: Can we give some advice for how developers should choose between these two options?

My guess is they should always use the token provider callback since manual token refreshing is complex. But is there some edge-case reason why people would want to do their own token refresh?

docs/setup/azure-managed-identity.md
```diff
@@ -65,9 +93,33 @@ npm install @github/copilot-sdk @azure/identity
 
 </details>
 
-### Basic usage
+### Provide an access token you manage
+
+Get a token using your language's Azure Identity SDK credential and pass it as the bearer token in your provider configuration. Track the token's expiration or refresh metadata so you know when to acquire a fresh token. Because a static bearer token is captured when the session is created, create a new session when you need to use the refreshed token. See the language-specific guidance in each code sample's comments below.
```
SteveSandersonMS: > Because a static bearer token is captured when the session is created, create a new session when you need to use the refreshed token.

This doesn't seem like a viable approach. Doesn't that mean you lose your sessions when tokens expire?

Do you think we should swap the doc around and describe the "automatic token acquisition" method as the only real method? Then we could also comment that if somehow you know you don't need to refresh tokens (e.g., because your sessions are always just one-shot operations not ongoing conversations) then you can pass a static token, but we don't have to document how to do any of the token refresh stuff because that's orthogonal to Copilot SDK.

docs/setup/azure-managed-identity.md
```diff
@@ -1,27 +1,32 @@
...
 
-This guide shows how to use the Azure Identity SDK's `DefaultAzureCredential` API to authenticate with Microsoft Foundry models through the Copilot SDK.
+This guide shows how to use Azure Identity SDK APIs to authenticate with Microsoft Foundry models through the GitHub Copilot SDK. Most languages use `DefaultAzureCredential`; Rust uses `DeveloperToolsCredential` locally and `ManagedIdentityCredential` in Azure.
 
 ## How it works
 
-Microsoft Foundry's OpenAI-compatible endpoint accepts bearer tokens from Microsoft Entra ID in place of static API keys. The pattern is:
+Microsoft Foundry's OpenAI-compatible endpoint (`https://<resource-name>.openai.azure.com/openai/v1/`) accepts bearer tokens from Microsoft Entra ID in place of static API keys. This guide uses a token provider callback so the GitHub Copilot SDK runtime can request fresh tokens on demand.
```
roji, line 9: I've never quite been sure (had similar uncertainty in my previous life working on PostgreSQL and identity): who has the responsibility for caching the token (if at all)? In other words, just to make sure we avoid guidance that would cause an expensive regeneration of a new token on each and every invocation of the callback (as long as e.g. DefaultAzureCredential caches, that's all fine).

May be worth a note here, maybe not...

roji (approved):
Looks great!

One thing I'm not too sure about, is whether the provided bearer token applies just for the inference backend (BYOK mode), or also for MCP servers which require auth... Assuming it's indeed the same bearer token that's supposed to go to both, I'm not quite sure what the flow looks like for such MCP servers, e.g. how do we know which MCP servers do Azure Managed Identity and which don't (I suspect this is something we don't yet support).

docs/setup/azure-managed-identity.md
scottaddie: Caching is handled by MSAL, which Azure Identity is built on top of. I deployed a sample app to Azure, connected it to a Foundry model, and confirmed the caching works as expected.

PR #1996, opened 2026-07-15
Add Rust runtime lock settings

SlowGreek:
## Summary
- add a typed Rust `RuntimeSettings` API with optional model and context-tier locks
- forward `runtimeSettings.modelLock` and `runtimeSettings.contextTierLock` on session create and resume
- preserve wire omission when runtime settings are absent

## Testing
- `cargo test runtime_settings --all-features`
- all non-E2E Rust unit and integration test targets

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1582,6 +1582,38 @@ impl ProviderModelConfig {
     }
 }
 
+/// Runtime settings applied when creating or resuming a session.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct RuntimeSettings {
+    /// Restricts model selection to this model identifier.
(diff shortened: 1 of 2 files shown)


roji (changes requested):
This doesn't seem to be actually supported in the runtime; so while this PR adds the SDK surface and serialization, it won't actually do anything.

In general, locking settings such as the model/context tier seems like something that needs to be handled in the hosting application rather than in the runtime. For example, a UI using the SDK can block the model picker in whatever way it wants. Can you provide some context/justification for actually having a feature like this in the runtime?

Finally, note that changes in the SDK generally need to be done across all languages, not just in a single one (Rust).

SteveSandersonMS:
Thanks for the contribution, @SlowGreek.

After review this doesn't seem to be going in the right direction for Copilot SDK. As per @roji's comment above, an integrator that wants to prevent changes to these settings can achieve that in its own UI; there's no clear reason we know why the SDK should enforce that.

PR #2019, opened 2026-07-17
docs: fix EnableConfigDiscovery summary to accurately describe agent discovery behavior

### LINKED ISSUE #1887: EnableConfigDiscovery - Incorrect Behaviour
The `SessionConfig.EnableConfigDiscovery` property's summary tag reads:

```cs
    /// <summary>
    /// When <see langword="true"/>, automatically discovers MCP server configurations
    /// (e.g. <c>.mcp.json</c>, <c>.vscode/mcp.json</c>) and skill directories from
    /// the working directory and merges them with any explicitly provided
    /// <see cref="McpServers"/> and <see cref="SkillDirectories"/>, with explicit
    /// values taking precedence on name collision.
    /// <para>
    /// Custom instruction files (<c>.github/copilot-instructions.md</c>, <c>AGENTS.md</c>, etc.)
    /// are always loaded from the working directory regardless of this setting.
    /// </para>
    /// </summary>
    public bool? EnableConfigDiscovery { get; set; }
```

Using this simple example, you can discover it's not this the case. Agents will ONLY be loaded when the value is set to true. Which is counter to:
"Custom instruction files (.github/copilot-instructions.md, AGENTS.md, etc. are always loaded from the working directory regardless of this setting"

```cs
using GitHub.Copilot;

// path to .github folder containing agents folder
const string WorkingDirectory = @"C:\SomeWorkspace";

await using var session = await client.CreateSessionAsync(new SessionConfig
{
    WorkingDirectory = WorkingDirectory,
    EnableConfigDiscovery = false, // switch this to true if you want agents to be actually loaded.
});

var agentsList = await session.Rpc.Agent.ListAsync();

foreach (var agent in agentsList.Agents)
{
    Console.WriteLine(agent.Name);
}
```

Agent file I was using in a .github/agents/sample.agent.md file
```markdown
---
name: sample
description: Sample Agent
---

You are an sample agent for this repository, you're just happy to be here!

```

Additionally, are there any plans to support an `AgentsDirectories` property similar to `SkillDirectories`, `PluginDirectories` & `InstructionDirectories`?

smz202000:
## Summary

Fixes #1887

The previous documentation for `EnableConfigDiscovery` incorrectly stated that custom instruction files are "always loaded from the working directory regardless of this setting." In practice, agent files (e.g. `.github/agents/`) are only discovered when `EnableConfigDiscovery` is `true`.

## Changes

Updated the XML doc comment on `SessionConfig.EnableConfigDiscovery` in `dotnet/src/Types.cs` to:

1. Clarify that agent discovery is gated by `EnableConfigDiscovery`
2. Reference `SkipCustomInstructions` as the independent control for instruction file loading

**Before:**
```csharp
/// <para>
/// Custom instruction files (.github/copilot-instructions.md, AGENTS.md, etc.)
/// are always loaded from the working directory regardless of this setting.
/// </para>
```

**After:**
```csharp
/// <para>
/// Agent discovery (e.g. .github/agents/) is only performed when this is
/// true. Use SkipCustomInstructions to control
/// loading of custom instruction files independently.
/// </para>
```

### DIFF
--- a/dotnet/src/Types.cs
+++ b/dotnet/src/Types.cs
@@ -2983,8 +2983,9 @@ public abstract class SessionConfigBase
     /// <see cref="McpServers"/> and <see cref="SkillDirectories"/>, with explicit
     /// values taking precedence on name collision.
     /// <para>
-    /// Custom instruction files (<c>.github/copilot-instructions.md</c>, <c>AGENTS.md</c>, etc.)
-    /// are always loaded from the working directory regardless of this setting.
+    /// Agent discovery (e.g. <c>.github/agents/</c>) is only performed when this is
+    /// <see langword="true"/>. Use <see cref="SkipCustomInstructions"/> to control
+    /// loading of custom instruction files independently.
     /// </para>
     /// </summary>
     public bool? EnableConfigDiscovery { get; set; }
--- a/dotnet/src/Types.cs
+++ b/dotnet/src/Types.cs
@@ -2977,16 +2977,8 @@ public abstract class SessionConfigBase
     public string? ConfigDirectory { get; set; }
 
     /// <summary>
-    /// When <see langword="true"/>, automatically discovers MCP server configurations
-    /// (e.g. <c>.mcp.json</c>, <c>.vscode/mcp.json</c>) and skill directories from
-    /// the working directory and merges them with any explicitly provided
-    /// <see cref="McpServers"/> and <see cref="SkillDirectories"/>, with explicit
-    /// values taking precedence on name collision.
-    /// <para>
-    /// Agent discovery (e.g. <c>.github/agents/</c>) is only performed when this is
-    /// <see langword="true"/>. Use <see cref="SkipCustomInstructions"/> to control
-    /// loading of custom instruction files independently.
-    /// </para>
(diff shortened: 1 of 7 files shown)


SteveSandersonMS (approved):
Reviewed and extended this PR to apply one consistent EnableConfigDiscovery description across all 6 SDKs (dotnet/go/nodejs/python/rust/java), replacing per-language wording that varied and, in 4 languages, incorrectly claimed custom instruction files are always loaded regardless of this setting.

Validation performed:
- Built/typechecked dotnet (net8.0), rust (cargo build), nodejs (tsc --noEmit), python (ruff check) — all clean. Go/Java toolchains unavailable locally but changes are comment-only with balanced braces verified.
- Manually reproduced the underlying behavior against a live copilot CLI runtime: with EnableConfigDiscovery=false, .github/agents/*.agent.md is NOT discovered (agent.list() returns []); with true it is. Separately confirmed AGENTS.md content is loaded into the model's context even when EnableConfigDiscovery=false (model referenced the exact planted marker as a 'custom_instruction').
- Full CI matrix (58 checks) passed across all languages after approving the required workflow runs.

PR #2051, opened 2026-07-22
rust: re-export ModelPolicyState and ModelPickerCategory from crate root

### LINKED ISSUE #1674: Re-export ModelPolicyState and ModelPickerCategory from the crate root alongside Model/ModelPolicy
### Summary

The crate-root prelude (`pub use types::*`) surfaces a curated set of
generated protocol types, but omits two enums that are the field types
of structs already re-exported there. This makes the model-info surface
inconsistent: you can name the container struct at the crate root but
must reach through `copilot::rpc` for the enum that one of its fields
holds.

### Details

`types.rs` curates the public re-export block:

```rust
/// Re-exports of generated protocol types that are part of the SDK's
/// public API surface. ...
pub use crate::generated::api_types::{
    Model, ModelBilling, ModelCapabilities, ModelCapabilitiesLimits, ModelCapabilitiesLimitsVision,
    ModelCapabilitiesSupports, ModelList, ModelPolicy, PermissionDecision,
    PermissionDecisionApproveOnce, PermissionDecisionReject, PermissionDecisionUserNotAvailable,
};
```

Two enums are missing from that list even though they are the declared
field types of structs that *are* in it:

- `ModelPolicy.state: ModelPolicyState` — `ModelPolicy` is re-exported; `ModelPolicyState` is not.
- `Model.model_picker_category: Option<ModelPickerCategory>` — `Model` is re-exported; `ModelPickerCategory` is not.

### Consequence

A consumer that pattern-matches on `policy.state` or reads
`model.model_picker_category` can refer to `copilot::Model` /
`copilot::ModelPolicy` from the crate root, but has to import the field
enums from a different path:

```rust
use copilot::{Model, ModelPolicy};            // crate root
use copilot::rpc::{ModelPolicyState, ModelPickerCategory}; // only here
```

Because the curated list looks complete, it's easy to assume the whole
model-info surface lives at the crate root and be surprised these two
don't. It's a small ergonomics/consistency papercut, not a correctness
issue (both are reachable via `copilot::rpc`).

### Suggested fix

Add `ModelPolicyState` and `ModelPickerCategory` to the curated
`pub use crate::generated::api_types::{ ... }` block in `types.rs`, so
the full model-info type set is reachable from the crate root alongside
`Model` and `ModelPolicy`.

rinceyuan:
Closes #1674

\ModelPolicy\ and \Model\ are re-exported at the crate root, but their field-type enums \ModelPolicyState\ and \ModelPickerCategory\ are not. This forces consumers to reach into \copilot::rpc\ for types that logically belong alongside the already-exported structs.

This adds both enums to the curated \pub use\ block in \	ypes.rs\, in alphabetical order.

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -5388,8 +5388,9 @@ impl InputFormat {
 pub use crate::generated::api_types::{
     Model, ModelBilling, ModelBillingTokenPrices, ModelBillingTokenPricesLongContext,
     ModelCapabilities, ModelCapabilitiesLimits, ModelCapabilitiesLimitsVision,
-    ModelCapabilitiesSupports, ModelList, ModelPolicy, PermissionDecision,
-    PermissionDecisionApproveOnce, PermissionDecisionReject, PermissionDecisionUserNotAvailable,
+    ModelCapabilitiesSupports, ModelList, ModelPickerCategory, ModelPolicy, ModelPolicyState,
+    PermissionDecision, PermissionDecisionApproveOnce, PermissionDecisionReject,
+    PermissionDecisionUserNotAvailable,
 };
 
 /// Permission categories the CLI may request approval for.


rinceyuan:

rinceyuan:
@stephentoub Tiny consistency fix — 2 missing enum re-exports for types whose parent structs are already at the crate root. Thanks!

rinceyuan:
Friendly follow-up — this has been open for 2 days. It's a one-line change (adding two enum re-exports to match the already-exported parent structs). Happy to address any feedback.

rinceyuan:
@stephentoub Second follow-up — this one-liner has been open 5 days. Would appreciate a quick 👍/👎 so I know if the approach is right. Happy to adjust.

SteveSandersonMS:
@rinceyuan A mixture of team events and folks out on vacation last week meant we've got a bit behind. We'll be catching up over the next few days.

SteveSandersonMS:
Thanks for identifying this inconsistency. The current crate-root re-export list is unfortunately already somewhat arbitrary, and adding individual types as omissions are discovered would make that boundary harder to explain and maintain. The same reasoning here also applies to several other generated field types.

We would prefer to define a clear policy for which generated types belong at the crate root, then have codegen produce that set automatically. Since that policy is not yet established, we do not want to extend the manually curated list further. All these types remain available through the intended `github_copilot_sdk::rpc` namespace, so we plan to close this PR rather than merge this particular change.

PR #2065, opened 2026-07-23
Fix Critical Privilege / Permission Bypass in Rust

DrHazemAli:
I discovered and fixed a critical privilege/permission bypass bug in the Rust SDK’s empty-mode session startup flow. Before this fix, `ClientMode::Empty` left config discovery unset during session creation, so the CLI could fall back to its default behavior and discover workspace MCP config before any permission decision was made.

## Flow
1. The app creates a session with `ClientMode::Empty`.
2. Session creation builds the wire payload.
3. `enable_config_discovery` was previously omitted.
4. The CLI applied its default and discovered workspace MCP config.
5. That happened before any tool permission request or approval flow.

```mermaid
flowchart TD
  A[User app / Rust SDK] --> B[Client::create_session]
  B --> C[SessionConfig]
  C --> D[wire payload for session.create]
  D --> E[Copilot CLI runtime]
  E --> F[MCP config discovery]
  F --> G[Workspace MCP servers / tools]

  H[Permission handler] --> I[permission.requested]
  I --> J[App approves / denies tool use]

  E -. before fix .-> F
  H -. normal tool flow .-> I
  ```
  
## Threat
This created an unsafe-defaults startup bypass. A caller expecting empty mode to be conservative could still end up with workspace MCP config discovery happening implicitly, which could expose configured tools or trigger config-driven behavior without the intended approval gate. This is a security issue because starting a malicious workspace-controlled MCP command is itself code execution during initialization. 

## What Fixed
- Empty mode now explicitly disables config discovery during session creation.
- Added a regression test to lock in the behavior.
- Kept the fix scoped to the session startup path that controlled the bypass.

### DIFF
--- a/rust/src/session.rs
+++ b/rust/src/session.rs
@@ -851,6 +851,9 @@ impl Client {
             crate::mode::system_message_for_mode(mode, config.system_message.take());
         config.memory = crate::mode::memory_for_mode(mode, config.memory.take());
         if mode == crate::ClientMode::Empty {
+            if config.enable_config_discovery.is_none() {
+                config.enable_config_discovery = Some(false);
+            }
             if config.enable_session_telemetry.is_none() {
                 config.enable_session_telemetry = Some(false);
             }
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks for raising this. I validated the behavior directly against the runtime version pinned by this repository using a workspace MCP server that writes a marker file when started:

- With `enableConfigDiscovery` omitted, the server was not started.
- With `enableConfigDiscovery: false`, the server was not started.
- With `enableConfigDiscovery: true`, the server was started during `session.create`.

So omission already behaves as `false`. We generally avoid explicitly disabling features that are already disabled by default, so this change is not needed. The added test also expects the field to remain omitted, which contradicts the implementation serializing `false`. Closing this PR.

PR #2099, opened 2026-07-27
Fix the Python list_sessions docstring example to use session_id

### LINKED ISSUE #2098: Python `list_sessions` docstring example uses a nonexistent `sessionId` attribute
### Summary

The `Example:` block in the docstring of `CopilotClient.list_sessions` (Python SDK) prints
`session.sessionId`, an attribute that does not exist on the objects the method returns.
Copying the documented example raises `AttributeError`.

`list_sessions` is annotated `-> list[SessionMetadata]` and builds its results with
`SessionMetadata.from_dict(...)`. `SessionMetadata` is a dataclass that declares `session_id`;
`sessionId` exists only as the wire/JSON key, mapped to `session_id` inside
`SessionMetadata.from_dict` / `to_dict`. There is no `sessionId` property, alias, or
`__getattr__` fallback, so the documented access can never succeed.

### Where

`python/copilot/client.py`, in the `CopilotClient.list_sessions` docstring:

```python
Example:
    >>> sessions = await client.list_sessions()
    >>> for session in sessions:
    ...     print(f"Session: {session.sessionId}")   # <- no such attribute
```

The returned type:

```python
@dataclass
class SessionMetadata:
    """Metadata about a session"""

    session_id: str  # Session identifier
    start_time: datetime
    modified_time: datetime
    is_remote: bool
    summary: str | None = None
    context: SessionContext | None = None
```

### Reproduction

No server, model, or credentials needed. The snippet constructs a `SessionMetadata` exactly the
way `list_sessions` does internally, then performs the documented access.

```bash
pip install github-copilot-sdk
```

```python
from copilot.client import SessionMetadata

# The wire payload shape `list_sessions` feeds into SessionMetadata.from_dict(...)
wire = {
    "sessionId": "abc-123",
    "startTime": "2026-06-27T10:00:00Z",
    "modifiedTime": "2026-06-27T10:05:00Z",
    "isRemote": False,
    "summary": "demo session",
}

session = SessionMetadata.from_dict(wire)

# The attribute that actually exists:
print(f"[ok] session.session_id = {session.session_id}")

# The attribute the documented example tells you to read:
print(f"Session: {session.sessionId}")
```

### Expected

The documented example prints each session's id, e.g. `Session: abc-123`.

### Actual

```
[ok] session.session_id = abc-123
Traceback (most recent call last):
  File "repro.py", line 18, in <module>
    print(f"Session: {session.sessionId}")
                      ^^^^^^^^^^^^^^^^^
AttributeError: 'SessionMetadata' object has no attribute 'sessionId'. Did you mean: 'session_id'?
```

### Notes

`session_id` is the correct spelling for this type, and it is what the rest of the SDK already
uses and asserts:

- the shipped dataclass field itself;
- the end-to-end tests, which assert `session_id` on exactly this return value
  (`python/e2e/test_session_e2e.py`);
- the generated sibling type `SessionMetadataSnapshot`, which also uses `session_id`;
- the sibling docstring for `get_session_metadata`, which documents the same object correctly as
  `metadata.start_time`.

Other bindings document this method with their own idiomatic spelling of the same wire field, so
the Python docstring is the only one affected: Go uses `session.SessionID` (`go/client.go`) and
.NET uses `session.SessionId` (`dotnet/src/Client.cs`).

Nothing in the repository executes these docstring examples (there is no `--doctest-modules`
setting and the docs-validation job only extracts Markdown under `docs/`), which is why lint,
type-checking, and the test suite all pass with the defect present.

The fix is to correct the documented attribute name. It is a docstring-only change with no
public API, behavior, or wire format impact.

examon:
Fixes #2098

### What was wrong

The `Example:` block in the `CopilotClient.list_sessions` docstring printed `session.sessionId`.
`list_sessions` returns `list[SessionMetadata]`, and that dataclass declares `session_id`;
`sessionId` exists only as the wire key, mapped to `session_id` inside
`SessionMetadata.from_dict` / `to_dict`. There is no `sessionId` property, alias, or
`__getattr__` fallback, so anyone copying the documented example hit `AttributeError`.

### The change

```diff
         Example:
             >>> sessions = await client.list_sessions()
             >>> for session in sessions:
-            ...     print(f"Session: {session.sessionId}")
+            ...     print(f"Session: {session.session_id}")
```

One token, in the docstring text of a hand-written binding source file. No public API, behavior,
schema, or wire format change; no generated file is touched.

### Why `session_id` is the right spelling

- `SessionMetadata` declares `session_id` and never `sessionId`.
- The end-to-end tests assert `session_id` on exactly this return value
  (`python/e2e/test_session_e2e.py`).
- The generated sibling type `SessionMetadataSnapshot` also uses `session_id`, and the Python
  code generator converts every property to snake_case.
- The sibling docstring for `get_session_metadata` already documents the same object correctly,
  as `metadata.start_time`.

The alternatives -- adding a `sessionId` alias, or renaming the dataclass field -- would both
change the public Python API, so correcting the example is the minimal correct fix.

### Reproduction

Against the published package, with no server, model, or credentials:

```bash
pip install github-copilot-sdk
```

```python
from copilot.client import SessionMetadata

wire = {
    "sessionId": "abc-123",
    "startTime": "2026-06-27T10:00:00Z",
    "modifiedTime": "2026-06-27T10:05:00Z",
    "isRemote": False,
    "summary": "demo session",
}
session = SessionMetadata.from_dict(wire)

# The documented example:
print(f"Session: {session.sessionId}")
```

**Before:**

```
Traceback (most recent call last):
  File "repro.py", line 13, in <module>
    print(f"Session: {session.sessionId}")
                      ^^^^^^^^^^^^^^^^^
AttributeError: 'SessionMetadata' object has no attribute 'sessionId'. Did you mean: 'session_id'?
```

**After:** the documented example prints `Session: abc-123`.

I also checked the shipped docstring itself, by parsing the example out of the installed
package's `CopilotClient.list_sessions.__doc__` and executing it: it raises `AttributeError`
before the change and prints the session id after.

### Checks run

From `python/`, matching `.github/workflows/python-sdk-tests.yml`:

- `ruff check .` -- passes.
- `ty check copilot` -- passes (2 pre-existing warnings, neither near this line).
- `ruff format --check copilot/client.py` -- already formatted; the changed line is 59 characters
  against a 100-character limit, so `docstring-code-format` has nothing to reflow.
- `pytest` -- per-test results are identical before and after the change, and
  `test_should_list_sessions` passes. The few end-to-end tests that fail in my local environment
  fail the same way on an unmodified tree.

### Scope

Python only. Go already documents this method as `session.SessionID` (`go/client.go`) and .NET as
`session.SessionId` (`dotnet/src/Client.cs`), so the other bindings are correct as they stand and
no cross-binding change is needed.

### DIFF
--- a/python/copilot/client.py
+++ b/python/copilot/client.py
@@ -3386,7 +3386,7 @@ class CopilotClient:
         Example:
             >>> sessions = await client.list_sessions()
             >>> for session in sessions:
-            ...     print(f"Session: {session.sessionId}")
+            ...     print(f"Session: {session.session_id}")
             >>> # Filter sessions by repository
             >>> from copilot.client import SessionListFilter
             >>> filtered = await client.list_sessions(SessionListFilter(repository="owner/repo"))


SteveSandersonMS (approved):
Validated the Python API mapping and reproduced both the broken camelCase access and corrected snake_case access with Python 3.12. The change is necessary, sufficient, and minimal.

PR #2115, opened 2026-07-28
chore: bump @github/copilot to ^1.0.75 and regenerate all SDKs

MRayermannMSFT:
## Why

Split out of #2114 so the Agent Factories authoring surface can be reviewed on its own. This half is almost entirely machine-generated: of ~12.1k added lines, roughly 200 are hand-written, and every one of those exists because the new schemas force it.

Bumping `@github/copilot` from `^1.0.73` to `^1.0.75` is also useful independently of factories — 1.0.75 is the first published runtime carrying the full factory wire surface, but the regeneration picks up everything else that landed alongside it.

## What

**Generated types** — regenerated for Node.js, Python, Go, .NET, and Rust from the 1.0.75 schemas.

**Codegen: narrow the opaque-JSON conversion.** `x-opaque-json` was being applied to every annotated definition, but only 1 of the 14 (`CanvasJsonSchema`) is genuinely unconstrained; the other 13 carry `anyOf` or `type` + `properties`. Blanket-converting them threw away the structure of `McpServerConfig`, `ExternalToolResult`, `EventLogTypes`, and `UIElicitationSchemaProperty`. The conversion now requires the absence of any structural keyword. Genuinely opaque nodes emit `JsonValue` rather than `unknown`, which is a global tightening.

**Fallout the new schemas force:**

- `sessionFs` gained `sqliteTransaction`, so every hand-written provider adapter must implement it. Node forwards to an optional provider hook and classifies failures; Python, Go, and .NET have no transaction hook in their provider interfaces, so they refuse the batch with a classified `fatal` error rather than applying it non-atomically. **This is the deliberate call worth a look** — refusing is safer than silently running a "transaction" that isn't atomic.
- `UIExitPlanModeResponse` gained `deferImplementation`, so a Rust e2e struct literal has to name it.
- `lsp.initialize` and `telemetry.setFeatureOverrides` are declared `Promise<void>` and now resolve with `{}` rather than `null`, so two e2e assertions no longer pin the resolved value.
- The `JsonValue` tightening reaches `ToolTelemetry`, the hooks-invoke response, and `data.arguments` in a session-event type test.

## Verification

Verified standalone on this branch (not just as part of #2114): Node typecheck clean, lint 0 errors, 675 unit tests passing; `cargo check --all-targets` clean; `go build ./...` and `go vet ./...` clean; `dotnet build` clean with 0 warnings.

## Notes for reviewers

The generated files are the bulk of the diff and are reproducible — `cd scripts/codegen && npm run generate` should be a no-op on this branch. The review that matters is the ~200 hand-written lines listed under *Fallout* above, plus the codegen predicate change in `scripts/codegen/typescript.ts`.

#2114 depends on this and will rebase onto it once this merges.

### DIFF
--- a/dotnet/src/Generated/Rpc.cs
+++ b/dotnet/src/Generated/Rpc.cs
@@ -65,7 +65,7 @@ internal sealed class ConnectResult
 [Experimental(Diagnostics.Experimental)]
 internal sealed class ConnectRequest
 {
-    /// <summary>Opt this connection in to GitHub telemetry forwarding for its lifetime. When set, the runtime forwards every internal telemetry event it emits — across all sessions, plus sessionless events — to this connection over the `gitHubTelemetry.event` notification, in addition to the runtime's normal GitHub/CTS emission (dual-write). Intended for first-party hosts that re-emit the events into their own telemetry stores. Both unrestricted and restricted events are forwarded, each tagged with a `restricted` discriminator; a backstop drops restricted events when restricted telemetry is disabled.</summary>
+    /// <summary>Opt this connection in to GitHub telemetry forwarding for its lifetime. When set, the runtime forwards every internal telemetry event it emits — across all sessions, plus sessionless events — to this connection over the `gitHubTelemetry.event` notification. Regular events are also written to the runtime's normal GitHub/CTS path (dual-write); host-only compatibility events are forward-only and intentionally skip that path. Intended for first-party hosts that re-emit the events into their own telemetry stores. Both unrestricted and restricted events are forwarded, each tagged with a `restricted` discriminator; a backstop drops restricted events when restricted telemetry is disabled — using the process-global gate for ordinary events and an explicit session-scoped decision for host-only events.</summary>
     [JsonPropertyName("enableGitHubTelemetryForwarding")]
     public bool? EnableGitHubTelemetryForwarding { get; set; }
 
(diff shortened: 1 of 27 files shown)


MRayermannMSFT:
**Parked for now.** The `copilot-sdk` team already has an update to the latest CLI in flight, so we are holding these rather than racing it. This PR stays open as a draft and should be re-evaluated once `main` carries a newer `@github/copilot`.

**Remaining CI blocker is a runtime regression, not this PR.** Two Python E2E tests fail on **Windows only**, deterministically across two full runs:

- `test_session_fs_e2e.py::test_should_write_workspace_metadata_via_sessionfs`
- `test_session_fs_e2e.py::test_should_persist_plan_md_via_sessionfs`

Both time out waiting for the runtime to persist session state through a SessionFS provider — `workspace.yaml` never reaches the provider. Evidence that it is not the SDK code here:

- Linux and macOS run the *same two tests* and pass, with identical Python code.
- The `session_fs_sqlite` tests pass on Windows, so the new `sqliteTransaction` adapter is never reached.
- The model call succeeds; the failure is the next line, a filesystem wait.
- Captures are byte-identical to `main`.
- These tests pass on `main` (runtime 1.0.73). The only changed variable is the runtime binary, 1.0.73 -> 1.0.75.

Prime suspect, unconfirmed: the SessionFs port to Rust. The test passes `session_state_path = "/session-state"`, a POSIX-style absolute path, and Windows path resolution is the obvious hazard there. This likely affects any SDK consumer using a SessionFS provider on Windows.

Note that the hand-written fixes in this PR are **not** factory-specific — they are what bumping past 1.0.73 forces at all (the `sqliteTransaction` adapters, the `x-opaque-json` codegen narrowing, the `JsonValue` tightening fallout, `deferImplementation`, and the two `Promise<void>` assertions). They should be useful to whoever lands the next bump, even if this PR itself is superseded.

MRayermannMSFT:
Superseded. SDK `main` moved to `@github/copilot@^1.0.76-5` in #2140, which already carries the factory wire types and the `sqliteTransaction` adapters this PR was providing. #2114 now rebases directly onto `main` and carries only the remaining piece that is still needed — the `x-opaque-json` codegen narrowing, without which a factory result generates as an object alone and a non-object or void result cannot satisfy it.

Issue #2116, opened 2026-07-28
[Go][tests] E2E harness picks the first @github/copilot-* package instead of the current platform package


nytron88:
### Summary

The Go test harness resolves the Copilot CLI by globbing `nodejs/node_modules/@github/copilot-*/index.js` and taking `matches[0]` — the alphabetically first result — rather than the package matching the current OS/architecture. This logic lives in `go/internal/e2e/testharness/context.go:36-41` and affects every E2E test that calls `testharness.CLIPath()`.

This is the same defect [#2103](https://github.com/github/copilot-sdk/issues/2103) reports for the Python harness, and the same one PR #2093 fixed for .NET.

```go
// go/internal/e2e/testharness/context.go
base := RepoPath("nodejs", "node_modules", "@github")
matches, _ := filepath.Glob(filepath.Join(base, "copilot-*", "index.js"))
if len(matches) > 0 {
    cliPath = matches[0]
    return
}
```

### Why This Is a Bug

`filepath.Glob` sorts its results, so `matches[0]` is whichever platform package sorts first — `copilot-darwin-arm64` beats `copilot-linux-x64` on a Linux host. The Node SDK resolves this correctly via `getCliPlatformPackageNames()` in `nodejs/src/client.ts:357`, which builds the candidate names from platform and architecture and tries both Linux libc variants (`linux` vs `linuxmusl`).

Two further problems in the same block:

1. **The glob is too loose.** `copilot-*` also matches non-platform packages under `@github`, such as `copilot-language-server`. Any of those shipping an `index.js` becomes a CLI candidate.
2. **The no-match diagnostic is duplicated and uninformative.** `CLIPath()` returns `""` rather than an error, so every caller re-implements the same guard — `NewTestContext` (`context.go:121-124`), plus `client_e2e_test.go:12-15`, `rpc_e2e_test.go:13-16`, `agent_and_compact_rpc_e2e_test.go:15-18` and `inprocess_ffi_e2e_test.go:29-32`, which each repeat `if cliPath == "" { t.Fatal("CLI not found. Run 'npm install' in the nodejs directory first.") }`. Nothing fails silently, but no message says which package was looked for, so "you never ran npm install" and "you have a `node_modules` for the wrong platform" — the situation this bug creates — are indistinguishable. `NewTestContext`'s variant also interpolates the empty path, printing "CLI not found at .".

### Where this actually bites

[...]

PR #2117, opened 2026-07-29
fix(python): select the current platform's CLI package in the E2E harness

### LINKED ISSUE #2103: [Python][tests] E2E harness picks the first @github/copilot-* package instead of the current platform package
## Summary

The Python test harness resolves the Copilot CLI by scanning `nodejs/node_modules/@github` for `copilot-*` directories and returning the first `index.js` it finds:

```py
for platform_pkg in sorted(github_modules.glob("copilot-*")):
    candidate = platform_pkg / "index.js"
    if candidate.exists():
        return str(candidate.resolve())
```

That selection is based on directory order, not the current OS/arch. As a result, Python tests can choose the wrong platform package when multiple `@github/copilot-*` packages are present, or whenever the checked-out `node_modules` tree contains only a mismatched platform package.

This logic lives in `python/e2e/testharness/context.py` and also affects tests that import `CLI_PATH` from that harness, such as `python/test_client.py`.

## Why this is a bug

The Node SDK already resolves platform packages explicitly in `nodejs/src/client.ts` via `getCliPlatformPackageNames()`, using the current platform/arch and handling Linux `linux` vs `linuxmusl` variants.

The repo also already fixed the same class of problem for the .NET E2E harness in #2093, where directory enumeration order was selecting the wrong platform-specific package.

Python still has the older directory-order-based logic.

## Repro

1. Leave `COPILOT_CLI_PATH` unset.
2. Have more than one `@github/copilot-*` package under `nodejs/node_modules/@github`, or otherwise have a checked-out `node_modules` tree whose available `copilot-*` package(s) do not match the current host platform.
3. Run any Python test that imports `e2e.testharness.context` or `CLI_PATH`.
4. `get_cli_path_for_tests()` returns the first matching package instead of the package for the current platform.
5. The Python test process then tries to launch the wrong CLI entrypoint.

## Expected behavior

The Python harness should resolve the platform package the same way the runtime code already does elsewhere in the repo:

- choose `win32` / `darwin` / `linux` based on the current platform
- choose the correct architecture
- on Linux, distinguish `linux` vs `linuxmusl`

## Suggested fix

Mirror the platform-aware package selection used in `nodejs/src/client.ts` instead of scanning `copilot-*` in sorted order.

## Related

- #2093 fixes the analogous .NET test harness issue
- #1739 introduced the newer `@github/copilot` package layout where the runnable `index.js` lives in platform-specific packages

nytron88:
Fixes #2103.

The Python E2E harness resolved the Copilot CLI by taking the first
`@github/copilot-*` directory in alphabetical order, so a tree with more than one
platform package could launch an entrypoint built for another OS/arch — whichever
name sorted first won, regardless of the host. The glob could also match
non-platform packages such as `copilot-language-server`.

It now probes the exact platform package names for the host — reusing
`copilot._cli_version.get_npm_platform()`, which already owns the platform, arch and
musl-vs-glibc mapping — mirroring `getCliPlatformPackageNames()` in
`nodejs/src/client.ts` and the .NET fix in #2093. On Linux both libc variants are
tried, detected one first. When nothing matches, the error names the directory
searched, the packages tried, the packages actually present, and `COPILOT_CLI_PATH`.

New unit tests in `python/test_e2e_harness_cli_path.py` cover this against fixture
directory trees under `tmp_path`, so no real CLI is ever launched.

### Behaviour change on mixed-architecture hosts

`platform.machine()` reflects the **Python interpreter**, but the resolved `index.js` is
executed by **node** (`python/copilot/client.py:3900-3901` prepends `node` for `.js`
paths), and which platform package exists on disk was decided by whichever npm ran the
install. On a host where those differ — x86_64 CPython under Rosetta alongside native
arm64 node — the old loose glob happened to find `copilot-darwin-arm64` and work, and
this change fails fast instead.

That is deliberate: silently launching a package built for a different architecture is
the bug being fixed. `COPILOT_CLI_PATH` keeps top precedence as the escape hatch and is
named in the error, which also lists the packages actually present so the mismatch is
visible rather than mysterious.

Worth naming the divergence from #2093: it keys on the OS prefix only and globs `-*` for
the arch, so it does not pin the architecture — this is the stricter of the two. Both
fail fast on the ordering bug itself (.NET errors on an ambiguous multi-package tree), so
the difference is narrowly about an arch mismatch on a single-package tree.

Note: `go/internal/e2e/testharness/context.go:36-41` has the identical bug — filed
separately as #2116 to keep this PR scoped to the issue.

### DIFF
--- a/python/e2e/testharness/context.py
+++ b/python/e2e/testharness/context.py
@@ -11,6 +11,7 @@ import re
 import shutil
 import tempfile
 import time
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any
 
@@ -38,6 +39,19 @@ def _cli_platform_package_names(npm_platform: str | None = None) -> list[str]:
             if name not in names:
                 names.append(name)
     return names
+
+
+def _find_cli_in_node_modules(github_modules: Path, package_names: Sequence[str]) -> str | None:
+    """Return the resolved ``index.js`` of the first installed candidate package.
+
+    Only exact package names are probed, so unrelated ``copilot-*`` directories
+    (e.g. ``copilot-language-server``) can never be mistaken for the CLI.
+    """
+    for name in package_names:
+        candidate = github_modules / name / "index.js"
+        if candidate.exists():
+            return str(candidate.resolve())
+    return None
 
 
 def get_cli_path_for_tests() -> str:
--- a/python/e2e/testharness/context.py
+++ b/python/e2e/testharness/context.py
@@ -54,6 +54,18 @@ def _find_cli_in_node_modules(github_modules: Path, package_names: Sequence[str]
     return None
 
 
+def _installed_cli_package_names(github_modules: Path) -> list[str]:
+    """Return the ``copilot-*`` directory names present, for error messages only.
+
+    Selection never globs — that was the #2103 bug. This exists so a failure can
+    say what *is* installed, which is the difference between a dead-end "run npm
+    install" and a message that diagnoses itself on a mixed-architecture host.
+    """
+    if not github_modules.is_dir():
+        return []
+    return sorted(path.name for path in github_modules.glob("copilot-*") if path.is_dir())
+
+
 def get_cli_path_for_tests() -> str:
     """Get CLI path for E2E tests.
 
@@ -75,10 +87,12 @@ def get_cli_path_for_tests() -> str:
     if found is not None:
         return found
 
+    installed = _installed_cli_package_names(github_modules)
     raise RuntimeError(
         f"CLI not found for tests under {github_modules} "
-        f"(tried: {', '.join(package_names)}). Run 'npm install' in the nodejs "
-        f"directory, or set COPILOT_CLI_PATH."
+        f"(tried: {', '.join(package_names)}; "
+        f"present: {', '.join(installed) or 'none'}). "
+        "Run 'npm install' in the nodejs directory, or set COPILOT_CLI_PATH."
     )
 
 
--- a/python/e2e/testharness/context.py
+++ b/python/e2e/testharness/context.py
@@ -57,7 +57,8 @@ def _find_cli_in_node_modules(github_modules: Path, package_names: Sequence[str]
 def get_cli_path_for_tests() -> str:
     """Get CLI path for E2E tests.
 
-    Uses COPILOT_CLI_PATH env var if set, otherwise node_modules CLI.
+    Uses COPILOT_CLI_PATH env var if set, otherwise the platform-specific CLI
+    package in the sibling nodejs directory's node_modules.
     """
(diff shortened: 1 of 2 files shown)


syedkazmi14:
@nytron88 Thanks, this is really helpful.

Reusing copilot._cli_version.get_npm_platform() sounds like the right approach to me. I’d prefer to keep this fix scoped to the platform-aware package selection bug itself, so I would leave the lazy CLI_PATH resolution change out of this PR and track that separately if it still seems worth doing after this lands.

The mixed-architecture fail-fast behavior also seems like the right trade-off here to me. If the current interpreter arch does not match the installed CLI package, surfacing that clearly is better than silently picking a different package, and COPILOT_CLI_PATH seems like a reasonable escape hatch for unusual setups.

Also appreciate you filing the Go version separately.

nytron88:
Thanks @syedkazmi14  — done, and agreed on both counts.

**Dropped the lazy `CLI_PATH` change.** Force-pushed; the branch is four commits now:

```
818249f2 fix(python): name the installed CLI packages when resolution fails
5a7105a0 fix(python): select the current platform's CLI package in E2E harness
469847bd test(python): probe exact CLI platform package names on disk
e164fb24 test(python): resolve E2E CLI candidates by host platform
```

The last one stays because it is part of the selection fix rather than the hardening: when
nothing matches, the error now lists the `copilot-*` packages actually present alongside the
ones tried. That matters specifically for the mixed-arch case below — otherwise the message
tells you to run `npm install` while the package npm would install is already sitting in that
directory, which is a dead end with no way to see why.

One consequence worth stating plainly, since it is the reason I had bundled the change: with
that commit out, `CLI_PATH = get_cli_path_for_tests()` runs at module scope again, so on a host
that cannot resolve a matching package the new unit tests fail at *collection* rather than as
test failures — even though they only use `tmp_path` fixtures and need no CLI. CI is unaffected
(`npm ci --ignore-scripts` runs in `nodejs/` before pytest), so this only shows up on a fresh
clone or a stale worktree locally. Happy to leave it exactly as you suggested; I have parked the
work on [`spike/lazy-cli-path-2103`](https://github.com/nytron88/copilot-sdk/tree/spike/lazy-cli-path-2103)
and will open an issue after this lands if it still looks worth doing.

**On the mixed-architecture trade-off** — that matches my read too, so I have reworded that part
of the PR description from an open question into a plain note about the behaviour change, and
kept the detail that the divergence from #2093 is intentional (that one keys on the OS prefix
only and does not pin the arch, so this is the stricter of the two).

Current state on this branch: 17 tests in `python/test_e2e_harness_cli_path.py`, 302 in the flat
unit suite, `e2e/test_client_e2e.py` 14/14 against the real CLI, `ruff format --check` and
`ruff check` clean on each commit individually.

PR #2162, opened 2026-07-30
fix(python): preserve boolean const discriminators in RPC codegen

### LINKED ISSUE #2122: python: boolean-discriminated unions never decode - sessions.list() raises for any non-empty result
## Summary

The Python SDK generates union dispatchers that `match` the discriminator against Python **string** literals `"true"` / `"false"`.
A JSON boolean decodes to Python `True` / `False`, which matches neither, so the dispatcher always falls through to `case _: raise ValueError(...)`.

The API schema has exactly two boolean-discriminated unions, and both are broken in Python:

| Union | Discriminator | Used by | Effect in Python |
| --- | --- | --- | --- |
| `SessionListEntry` | `isRemote` (`{"type": "boolean", "const": false/true}`) | `sessions.list` result | decode raises for any non-empty list |
| `QueuedCommandResult` | `handled` (boolean const) | `session.commands.respondToQueuedCommand` | fails to decode, and serializes the string `"true"` instead of `true` |

Every other discriminated union in the schema uses string consts, which is why this went unnoticed.

`isRemote` is `required` on both variants, so there is no payload for which the dispatcher succeeds. The type cannot even round-trip its own output.

## Impact

`client.rpc.sessions.list()` raises on any non-empty result. It succeeds only when the session list is empty.

The `QueuedCommandResult` symptom is a wire-contract violation rather than a visible failure today. It is worth fixing regardless: both variants serialize to a non-empty string, so `handled: "false"` is indistinguishable from `handled: "true"` to any consumer that reads the field as a boolean.

## Reproduction

```bash
pip install github-copilot-sdk
python repro.py
```

`repro.py` - no runtime or network needed, this is a pure decode:

```python
from copilot.rpc import SessionList

# The wire shape `sessions.list` returns. Per the API schema,
# LocalSessionMetadataValue.isRemote is {"type": "boolean", "const": false}
# and it is a required property.
payload = {
    "sessions": [
        {
            "sessionId": "example-local",
            "startTime": "2026-07-26T10:00:00.000Z",
            "modifiedTime": "2026-07-26T10:05:00.000Z",
            "isRemote": False,
        }
    ]
}

print(SessionList.from_dict(payload))
```

Expected: a `SessionList` holding one `LocalSessionMetadataValue`.

Actual:

```
Traceback (most recent call last):
  File "repro.py", line 17, in <module>
    print(SessionList.from_dict(payload))
  File ".../copilot/generated/rpc.py", line 7857, in from_dict
    sessions = from_list(_load_SessionListEntry, obj.get("sessions"))
  File ".../copilot/generated/rpc.py", line 65, in from_list
    return [f(y) for y in x]
  File ".../copilot/generated/rpc.py", line 27771, in _load_SessionListEntry
    case _: raise ValueError(f"Unknown SessionListEntry isRemote: {kind!r}")
ValueError: Unknown SessionListEntry isRemote: False
```

A remote entry (`"isRemote": True`) fails the same way with `Unknown SessionListEntry isRemote: True`.

### Second symptom: the discriminator is serialized as a string

```python
>>> from copilot.rpc import QueuedCommandHandled
>>> QueuedCommandHandled.from_dict({"handled": True}).to_dict()
{'handled': 'true'}
```

The schema declares `handled` as `{"type": "boolean", "const": true}` with `additionalProperties: false`, so this payload violates the wire contract.

## Generated code

```python
def _load_SessionListEntry(obj: Any) -> "SessionListEntry":
    assert isinstance(obj, dict)
    kind = obj.get("isRemote")
    match kind:
        case "false": return LocalSessionMetadataValue.from_dict(obj)
        case "true": return RemoteSessionMetadataValue.from_dict(obj)
        case _: raise ValueError(f"Unknown SessionListEntry isRemote: {kind!r}")
```

`match True: case "true":` never matches - a literal pattern compares by equality, and a `bool` is never equal to a `str`.

## Root cause

`scripts/codegen/python.ts` coerces the schema `const` to a JavaScript string when it captures a discriminator value (`String(discProp.const)`), and types the dispatch table as `Array<{ value: string; typeName: string }>`. The boolean type is lost before the emitters run, so they can only ever render a Python string literal - both in the `match` arms and in the variant's `handled: ClassVar[str] = "true"`.

`scripts/codegen/go.ts` already models this correctly with `type GoDiscriminatorValue = string | boolean`, which is why Go decodes `switch *raw.IsRemote { case false: ...; case true: ... }` and is unaffected. Node.js, .NET and Rust are unaffected as well; Python is the only binding that loses the type.

## Why CI does not catch it

`python/e2e/test_rpc_server_e2e.py` already calls `client.rpc.sessions.list(...)` and asserts over `listed.sessions`, but never requires the list to be non-empty. An empty list is the only payload that decodes, so the assertions pass vacuously.

## Environment

- `github-copilot-sdk` 1.0.8 (published wheel), Python 3.13, Linux x86-64
- Also present on `main` at `aa4e707f28c16f683965871c6adb47d2896d7aa5`
- Deterministic: fails on every run

syf2211:
## Summary

Fix Python RPC codegen so JSON boolean `const` discriminator values are preserved instead of being coerced to strings. This restores decoding for `SessionListEntry` (`isRemote`) and `QueuedCommandResult` (`handled`).

## Motivation

`sessions.list()` raised `ValueError` for any non-empty session list because the generated dispatcher matched `case "true":` / `case "false":` against Python `bool` values. `QueuedCommandHandled.to_dict()` also serialized `handled` as the string `"true"` instead of JSON boolean `true`.

Fixes #2122

## Changes

- Add `PyDiscriminatorValue` helpers in `scripts/codegen/python.ts` to emit `True`/`False` match patterns and `ClassVar[bool]` defaults
- Regenerate `python/copilot/generated/rpc.py`
- Add unit tests covering boolean union decode/encode and `SessionList.from_dict`

## Tests

```bash
cd python && python3 -m pytest test_rpc_generated.py -q
```

Result: 3 passed

## Notes

- String-discriminated unions are unchanged (`JSON.stringify` path preserved)
- Go codegen already handled booleans correctly; Python was the only affected binding
- composer-2.5 review: APPROVE

### DIFF
--- a/python/copilot/generated/rpc.py
+++ b/python/copilot/generated/rpc.py
@@ -7391,7 +7391,7 @@ class QueuedCommandHandled:
     """Queued-command response indicating the host executed the command, with an optional flag
     to stop queue processing.
     """
-    handled: ClassVar[str] = "true"
+    handled: ClassVar[bool] = True
     """The host actually executed the queued command."""
 
     stop_processing_queue: bool | None = None
@@ -7418,7 +7418,7 @@ class QueuedCommandNotHandled:
     """Queued-command response indicating the host did not execute the command and the queue may
     continue.
     """
-    handled: ClassVar[str] = "false"
+    handled: ClassVar[bool] = False
     """The host did not execute the queued command. Unblocks the queue without claiming the
     command was processed (e.g. when the handler threw before completing).
     """
@@ -31183,8 +31183,8 @@ def _load_QueuedCommandResult(obj: Any) -> "QueuedCommandResult":
     assert isinstance(obj, dict)
     kind = obj.get("handled")
     match kind:
-        case "true": return QueuedCommandHandled.from_dict(obj)
-        case "false": return QueuedCommandNotHandled.from_dict(obj)
+        case True: return QueuedCommandHandled.from_dict(obj)
+        case False: return QueuedCommandNotHandled.from_dict(obj)
         case _: raise ValueError(f"Unknown QueuedCommandResult handled: {kind!r}")
 
 # State of the runtime-managed remote-control singleton.
@@ -31207,8 +31207,8 @@ def _load_SessionListEntry(obj: Any) -> "SessionListEntry":
     assert isinstance(obj, dict)
     kind = obj.get("isRemote")
     match kind:
-        case "false": return LocalSessionMetadataValue.from_dict(obj)
-        case "true": return RemoteSessionMetadataValue.from_dict(obj)
+        case False: return LocalSessionMetadataValue.from_dict(obj)
+        case True: return RemoteSessionMetadataValue.from_dict(obj)
         case _: raise ValueError(f"Unknown SessionListEntry isRemote: {kind!r}")
 
 # Open a session by creating, resuming, attaching, connecting to a remote, or handing off.
--- a/python/test_rpc_generated.py
+++ b/python/test_rpc_generated.py
@@ -4,9 +4,15 @@ from unittest.mock import AsyncMock
 
 import pytest
 
+from copilot.generated.rpc import _load_SessionListEntry
 from copilot.rpc import (
     CommandsApi,
     CommandsInvokeRequest,
+    LocalSessionMetadataValue,
+    QueuedCommandHandled,
+    QueuedCommandNotHandled,
+    RemoteSessionMetadataValue,
+    SessionList,
     SlashCommandTextResult,
 )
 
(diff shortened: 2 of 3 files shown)


SteveSandersonMS:
Thanks for the fix. This duplicates #2123, which addresses the same Python boolean-discriminator codegen bug with broader test coverage and is the version we plan to proceed with, so I’m closing this PR in favor of #2123.

PR #2163, opened 2026-07-30
fix(go): select the host CLI package in E2E tests

### LINKED ISSUE #2116: [Go][tests] E2E harness picks the first @github/copilot-* package instead of the current platform package

Pybsama:
Closes #2116.

## What this changes

The Go E2E harness now derives exact `@github/copilot-<platform>-<arch>`
package candidates from `ffihost.PrebuildsFolder()` instead of globbing every
`copilot-*` directory and taking the first sorted result.

- macOS and Windows probe only their exact host package;
- Linux probes the detected glibc/musl package first and the sibling libc
  variant second;
- unrelated or foreign-platform `copilot-*` packages are ignored;
- `CLIPath(testing.TB)` reports the packages tried and those actually present,
  replacing the duplicated empty-path guards at its call sites; and
- `go/test.sh` applies the same exact host OS/architecture mapping.

The shared `.github/actions/setup-copilot` action is intentionally unchanged,
as called out in the Issue.

## Tests

Added unit coverage for:

- macOS, Windows, Linux glibc, and Linux musl candidate generation;
- ignoring alphabetically earlier foreign and non-platform packages;
- respecting Linux libc candidate order;
- rejecting a package without an `index.js`; and
- listing installed packages for failure diagnostics.

Local validation:

- `GOTOOLCHAIN=go1.24.0 go test ./internal/e2e/testharness`
- `go test ./...`
- `env -u COPILOT_CLI_PATH bash test.sh`
- `COPILOT_SDK_DEFAULT_CONNECTION=inprocess GOFLAGS=-tags=copilot_inprocess env -u COPILOT_CLI_PATH bash test.sh`
- `golangci-lint run ./... --timeout=5m` (`0 issues`)
- `bash -n go/test.sh`
- cross-compiled the testharness tests for Linux, Windows, and macOS on amd64
  and arm64

Both the default and in-process full race-enabled test runs passed.


Pybsama:
Revalidated this PR against current `main` (`53efb3593e65e5f04474099ffd3e00efa7777482`). Upstream has moved 137 commits since the PR's original base, but the affected `CLIPath` and `go/test.sh` discovery blocks still use the loose sorted `copilot-*` glob, and I found no other open PR for #2116.

The existing PR patch applies cleanly on current `main`; its stable patch ID is unchanged (`c65eba0db789bd46b864cadfe3228471100d46f1`). GitHub's current synthetic merge commit `4a58782925d685839fbd94735eebfa6969910ac9` has current `main` and the public PR head as its two parents, and its tree `0ea94a0a517038537a90d8e5741fa77ea9b73bd6` matches the locally validated current-main tree byte-for-byte. Current-main validation:

- The added focused regression is RED on unmodified current main because the exact platform-candidate helpers are absent; with the patch it passes on Go 1.24.13 and for 50 consecutive runs. It also passes for 20 runs on Go 1.26.5.
- With an alphabetically earlier foreign package present, exact `go/test.sh` discovery probes selected `copilot-darwin-arm64`, `copilot-linux-x64`, and `copilot-win32-arm64` for the corresponding simulated hosts.
- Using the workflow's Go 1.24 line (latest patch 1.24.13), official Node 22.23.2, and Copilot CLI 1.0.82, both complete race-enabled matrices passed locally: default transport (322 top-level passes; E2E package 276.743s) and in-process transport (318 top-level passes; E2E package 213.131s). Both selected the Darwin ARM64 package.
- `go vet ./...`, Go formatting, Bash syntax, golangci-lint 2.12.2 (0 issues), Linux/Windows/macOS testharness cross-compilation on amd64 and arm64, default and in-process Go 1.26.5 compile checks, and `git diff --check` all pass.
- Public Semgrep rules reported 0 findings across 134 rules on the two changed Go files and `go/test.sh`.

Both full E2E runs rewrite replay snapshots as part of the harness; those generated changes were restored after each run, and the final source worktree is clean.

I also attempted an exact `--force-with-lease` update from the unchanged public head `33f12d115a1bb67d03a80c250354ca66780c0665`. GitHub rejected it before changing the ref because the intervening upstream history modifies workflows and the current OAuth token lacks `workflow` scope. I did not expand credential scopes. The public head therefore remains unchanged; GitHub currently reports the PR as mergeable, and the patch above was validated on the current-main equivalent tree.

SteveSandersonMS (approved):
I independently reproduced the platform-selection bug and validated the fix. The change is confined to Go E2E discovery and is safe to merge.

PR #2176, opened 2026-07-30
Update .NET extension reload E2E contract

dmytrostruk:
## Summary

- update the loaded-extension E2E contract for `session.extensions.reload`
- verify reload re-reads persisted global settings and clears a session-local disable when the global default remains enabled
- assert the relaunched extension is running with a PID

Depends on github/copilot-agent-runtime#13831. The current SDK CI runtime still implements the previous session-local reload behavior, so this contract test becomes green when that runtime change is available.

## Validation

- `git diff --check`
- .NET test execution unavailable locally because this environment has no `dotnet` toolchain


dmytrostruk:
Closing for sequencing: this contract-test update should be generated and applied with the SDK release after github/copilot-agent-runtime#13831 is finalized, rather than landing as a standalone SDK PR now.

PR #2180, opened 2026-07-31
feat: add support for additional directories in session configuration

DonJayamanne:
- Introduced `additionalDirectories` field in various session-related configurations across SDKs (Java, Go, .NET, Node.js, Python, Rust).
- Updated session creation and resume methods to handle `additionalDirectories`.
- Enhanced tests to verify that additional directories are correctly forwarded during session creation and resumption.
- Updated documentation to reflect the new configuration options.


SteveSandersonMS (approved):
It seems like the underlying runtime feature only works for newly-created sessions, not resumed ones, but that won't affect the SDK code so I'll approve here anyway.

DonJayamanne:
> It seems like the underlying runtime feature only works for newly-created sessions, not resumed ones, but that won't affect the SDK code so I'll approve here anyway.

Hmm, it worked for me, will have a look.

PR #2185, opened 2026-07-31
Test extension canvas forwarding

jmoseley:
A canvas registration regression in Copilot CLI 1.0.76-5 initially appeared to originate in the SDK extension path. The investigation confirmed that `joinSession({ canvases })` forwards canvas providers correctly; the actual failure is an agent-runtime connection identity race.

This adds focused coverage for the SDK boundary so future regressions cannot silently drop canvases before `resumeSessionForExtension`.

This PR is test-only and does not fix the runtime issue. The production fix belongs in `copilot-agent-runtime`, where extension identity must be published before the first JSON-RPC request is dispatched. Runtime PR github/copilot-agent-runtime#13958 implements that ordering.

### DIFF
--- a/nodejs/test/extension.test.ts
+++ b/nodejs/test/extension.test.ts
@@ -57,4 +57,25 @@ describe("joinSession", () => {
 
         expect(canvas.declaration.id).toBe("counter");
     });
+
+    it("forwards canvas providers when joining as an extension", async () => {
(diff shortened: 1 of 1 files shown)


jmoseley:
Closing because this test only covers the SDK forwarding boundary and would not catch the agent-runtime registration race. The regression coverage and fix belong in copilot-agent-runtime.

PR #2198, opened 2026-08-01
fix(go): tolerate numeric ping timestamps

### LINKED ISSUE #1356: Client.Start fails when CLI returns ping `timestamp` as a JSON string (PingResponse.Timestamp int64)
# `Ping`/`Client.Start` fails when the CLI sends `timestamp` as a JSON string

## Summary

`copilot.Client.Start` aborts before any RPC can be issued whenever the
connected Copilot CLI serializes the `ping` response's `timestamp` field as a
JSON string instead of a JSON number.

The Go SDK declares `PingResponse.Timestamp` as a hard `int64`:

```go
// types.go (v0.3.0)
type PingResponse struct {
    Message         string `json:"message"`
    Timestamp       int64  `json:"timestamp"`
    ProtocolVersion *int   `json:"protocolVersion,omitempty"`
}
```

…and `Client.Start` → `verifyProtocolVersion` → `Ping` → `json.Unmarshal`
bubbles every unmarshal failure back to the caller verbatim, with no
compatibility shim.

In practice this means: if the CLI on a given platform returns
`"timestamp":"2026-05-21T08:29:54.042Z"` (ISO 8601 string), the SDK can
never start, never enumerate models, never create a session — full stop —
with this hard-to-debug error message:

```
json: cannot unmarshal string into Go struct field PingResponse.timestamp of type int64
```

I am filing a sibling issue against `copilot-cli` so that the wire format
gets unified across platforms (currently Windows builds emit a number,
Linux builds emit an ISO string at the same CLI version `1.0.51` — see
attached captures). However, even after the CLI is unified, the SDK should
not be one hostile string field away from refusing to start. **Please add a
tolerant `UnmarshalJSON` on `PingResponse` (or change the field to a
shape-tolerant type) so that the SDK can survive both wire shapes.**

## Reproduction (no CLI required)

A minimal, dependency-isolated reproducer is in
[`tmp/sdk-repro` of this gist / branch](#) — two Go files (`go.mod`,
`repro_test.go`) totalling ~60 lines. Both `payload` constants below are
real captures taken via a hand-rolled JSON-RPC harness against a real
`copilot --headless --stdio` process; they are not synthetic.

```go
// payloadIntTimestamp     - Windows host, copilot CLI 1.0.51-2  -> NUMBER
// payloadStringTimestamp  - Linux container, copilot CLI 1.0.51 -> STRING
const payloadStringTimestamp = `{"message":"pong","timestamp":"2026-05-21T08:29:54.042Z","protocolVersion":3}`
const payloadIntTimestamp    = `{"message":"pong","timestamp":1779352370134,"protocolVersion":3}`

func TestPingResponse_StringTimestamp_Reproduces(t *testing.T) {
    var resp copilot.PingResponse
    err := json.Unmarshal([]byte(payloadStringTimestamp), &resp)
    // err != nil; err.Error() contains:
    //   "cannot unmarshal string into Go struct field PingResponse.timestamp of type int64"
}
```

Inside `docker.io/library/golang:latest` (go1.26.3 / linux/amd64):

```
=== RUN   TestPingResponse_NumericTimestamp_OK
    baseline OK: {Message:pong Timestamp:1779352370134 ProtocolVersion:0x...}
--- PASS
=== RUN   TestPingResponse_StringTimestamp_Reproduces
    reproduced error: json: cannot unmarshal string into Go struct field PingResponse.timestamp of type int64
--- PASS
```

Note neither test requires the Copilot CLI to be installed — the bug lives
purely in the exported SDK type.

## Where it bites in `Client.Start`

`client.go` (v0.3.0):

```go
// 302: func (c *Client) Start(ctx context.Context) error
//   ...
//   329: if err := c.verifyProtocolVersion(ctx); err != nil {
//   330:     killErr := c.killProcess()
//   331:     c.state = StateError
//   332:     return errors.Join(err, killErr)
//   333: }

// 1326: func (c *Client) verifyProtocolVersion(ctx context.Context) error {
// 1328:     pingResult, err := c.Ping(ctx, "")
// 1329:     if err != nil { return err }   // <-- unmarshal error reaches here
//   ...

// 1216: func (c *Client) Ping(ctx context.Context, message string) (*PingResponse, error) {
//   ...
// 1226: var response PingResponse
// 1227: if err := json.Unmarshal(result, &response); err != nil {
// 1228:     return nil, err              // <-- error originates here
// 1229: }
```

There is no `ClientOptions` knob (e.g. `SkipPingVerification`,
`PingResponseHook`) that lets a downstream consumer bypass this code path.

## Proposed fix

Any one of the following resolves the immediate breakage; the first is
preferred because it preserves the typed-`int64` ergonomics for existing
callers:

**Option A — tolerant `UnmarshalJSON` on `PingResponse`** (recommended)

```go
func (p *PingResponse) UnmarshalJSON(b []byte) error {
    type alias struct {
        Message         string          `json:"message"`
        Timestamp       json.RawMessage `json:"timestamp"`
        ProtocolVersion *int            `json:"protocolVersion,omitempty"`
    }
    var a alias
    if err := json.Unmarshal(b, &a); err != nil {
        return err
    }
    p.Message = a.Message
    p.ProtocolVersion = a.ProtocolVersion

    raw := bytes.TrimSpace(a.Timestamp)
    switch {
    case len(raw) == 0 || string(raw) == "null":
        // leave zero
    case raw[0] == '"':
        // String form. Support both ISO-8601 and stringified epoch ms.
        var s string
        if err := json.Unmarshal(raw, &s); err != nil {
            return err
        }
        if n, err := strconv.ParseInt(s, 10, 64); err == nil {
            p.Timestamp = n
        } else if t, err := time.Parse(time.RFC3339Nano, s); err == nil {
            p.Timestamp = t.UnixMilli()
        } else {
            return fmt.Errorf("PingResponse.timestamp: unrecognised string %q", s)
        }
    default:
        if err := json.Unmarshal(raw, &p.Timestamp); err != nil {
            return err
        }
    }
    return nil
}
```

**Option B — change `Timestamp` to `json.Number`** and add `Time()` /
`Int64()` helpers. Less invasive but pushes the parse responsibility onto
every caller and silently accepts garbage strings.

**Option C — `ClientOptions.SkipProtocolVerification bool`** as an escape
hatch for consumers stuck on broken CLI builds. Only useful in combination
with (A) or (B); on its own it defers the same crash to the first real RPC
that hits a similarly-typed field.

I'm happy to send the PR for Option A if it's the direction you want.

## Environment

- SDK: `github.com/github/copilot-sdk/go v0.3.0`
- (Also verified still affected on `v1.0.0-beta.4`.)
- Reproducer: `go1.26.3 linux/amd64` inside `docker.io/library/golang:latest`
- Observed wire payloads: see captures attached in sibling
  `github/copilot-cli` issue.

xianjianlf2:
Fixes #1356.

## Summary
- add custom JSON unmarshalling for Go ping responses so both the public `PingResponse` and generated `rpc.PingResult` accept ISO-8601 timestamps, numeric epoch milliseconds, and stringified epoch milliseconds
- cover both ping response shapes with unit tests

## Root cause
The Go SDK models ping timestamps as `time.Time`, which accepts ISO-8601 strings but rejects older numeric epoch-millisecond payloads from some CLI builds. That can make startup fail during protocol verification before callers can issue any SDK requests.

## Validation
- `go test . ./rpc`
- `go test ./...` was also run; non-E2E packages passed, but `go/internal/e2e` failed because this checkout does not have the Copilot CLI package installed under `nodejs/node_modules/@github/copilot-*/index.js`.

### DIFF
--- a/go/ping_response_json.go
+++ b/go/ping_response_json.go
@@ -1,0 +1,76 @@
+package copilot
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"time"
+)
+
+// UnmarshalJSON accepts both current ISO-8601 ping timestamps and older
+// epoch-millisecond timestamps emitted by some CLI builds.
+func (p *PingResponse) UnmarshalJSON(data []byte) error {
+	var wire struct {
+		Message         string          `json:"message"`
+		ProtocolVersion *int            `json:"protocolVersion,omitempty"`
+		Timestamp       json.RawMessage `json:"timestamp"`
+	}
+	if err := json.Unmarshal(data, &wire); err != nil {
+		return err
+	}
+
+	timestamp, err := parsePingTimestamp(wire.Timestamp)
+	if err != nil {
+		return err
+	}
+
+	p.Message = wire.Message
+	p.ProtocolVersion = wire.ProtocolVersion
+	p.Timestamp = timestamp
+	return nil
+}
+
+func parsePingTimestamp(raw json.RawMessage) (time.Time, error) {
+	raw = bytes.TrimSpace(raw)
+	if len(raw) == 0 || bytes.Equal(raw, []byte("null")) {
+		return time.Time{}, nil
+	}
+
+	if raw[0] == '"' {
+		var value string
+		if err := json.Unmarshal(raw, &value); err != nil {
+			return time.Time{}, fmt.Errorf("ping timestamp: %w", err)
+		}
+		return parsePingTimestampString(value)
+	}
+
+	var number json.Number
+	decoder := json.NewDecoder(bytes.NewReader(raw))
+	decoder.UseNumber()
+	if err := decoder.Decode(&number); err != nil {
+		return time.Time{}, fmt.Errorf("ping timestamp: %w", err)
+	}
+	return parsePingTimestampNumber(number)
+}
+
+func parsePingTimestampString(value string) (time.Time, error) {
+	if milliseconds, err := strconv.ParseInt(value, 10, 64); err == nil {
+		return time.UnixMilli(milliseconds), nil
+	}
+	if timestamp, err := time.Parse(time.RFC3339Nano, value); err == nil {
+		return timestamp, nil
+	}
+	return time.Time{}, fmt.Errorf("ping timestamp: unsupported string value %q", value)
+}
+
+func parsePingTimestampNumber(number json.Number) (time.Time, error) {
+	if milliseconds, err := number.Int64(); err == nil {
+		return time.UnixMilli(milliseconds), nil
+	}
+	value, err := strconv.ParseFloat(number.String(), 64)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("ping timestamp: %w", err)
+	}
+	return time.UnixMilli(int64(value)), nil
+}
--- a/go/rpc/generated_rpc_union_test.go
+++ b/go/rpc/generated_rpc_union_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"io"
 	"testing"
+	"time"
 
 	"github.com/github/copilot-sdk/go/internal/jsonrpc2"
 )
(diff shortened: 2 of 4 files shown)


SteveSandersonMS:
Thanks for the contribution, @xianjianlf2. However since we don't support using mismatched SDK and runtime versions (especially not radically mismatches ones as in this issue), it's not something we'd want to add support for.

PR #2200, opened 2026-08-01
fix: support bracketed IPv6 runtime URLs

### LINKED ISSUE #1137: cliUrl / cli_url parser rejects IPv6 addresses (including bracketed [::1]:port form)
## Summary

The `cliUrl` (Node, .NET) / `cli_url` (Python) / `CLIUrl` (Go) URL parser rejects IPv6 addresses in every standard form, including the canonical bracketed `[::1]:8080` notation. The error message ("Invalid cli_url format") doesn't suggest IPv6 isn't supported, so users debugging it have no obvious next step.

This is a feature gap, not a security issue.

## Affected versions

`main` at commit `dd2dcbc439256acfb9feb2cff07c0b9c820091b8` (current HEAD as of 2026-04-26). All language SDKs share the same parsing logic and exhibit the same gap.

## Affected source

- `python/copilot/client.py:944-986` — `_parse_cli_url`
- `nodejs/src/client.ts:351-376` — `parseCliUrl`
- `go/client.go:256-283` — `parseCliUrl`

## Reproduction

```python
from copilot.client import CopilotClient, ExternalServerConfig
c = CopilotClient(ExternalServerConfig(url="localhost:1"))

# IPv6 loopback with port — standard URL bracketed form
c._parse_cli_url("[::1]:8080")
# → ValueError: Invalid cli_url format: [::1]:8080

# IPv6 loopback alone
c._parse_cli_url("::1")
# → ValueError: Invalid cli_url format: ::1

# Link-local IPv6 with port
c._parse_cli_url("fe80::1:8080")
# → ValueError: Invalid cli_url format: fe80::1:8080

# Same with http:// prefix (which the parser otherwise accepts)
c._parse_cli_url("http://[::1]:8080")
# → ValueError: Invalid cli_url format: http://[::1]:8080
```

For comparison, IPv4 works as expected:

```python
c._parse_cli_url("127.0.0.1:8080")
# → ('127.0.0.1', 8080)
```

## Root cause

The Python parser splits on `:` and asserts exactly two parts:

```python
# python/copilot/client.py:973-975
parts = clean_url.split(":")
if len(parts) != 2:
    raise ValueError(f"Invalid cli_url format: {url}")
```

IPv6 addresses contain multiple colons, so any IPv6 form produces `len(parts) > 2` and is rejected. Bracketed-form parsing (the standard way to disambiguate `host:port` when the host itself contains colons, per RFC 3986 §3.2.2) is not implemented.

The Node and Go parsers have the same shape (`split(":")` / `strings.Cut`) and the same gap.

## Impact

- IPv6-only environments cannot use `cliUrl` to connect to a backend CLI server.
- The error message does not hint that IPv6 is unsupported, so users hit a dead end debugging.
- The documented `backend-services` and `scaling` deployment patterns implicitly require IPv4.

## Suggested fix

Use the language's URL parser instead of hand-rolled splitting:

**Python**:
```python
from urllib.parse import urlsplit
def _parse_cli_url(self, url: str) -> tuple[str, int]:
    if "://" not in url:
        url = "tcp://" + url
    parts = urlsplit(url)
    if parts.hostname is None or parts.port is None:
        raise ValueError(f"Invalid cli_url format: {url}")
    return (parts.hostname, parts.port)
```

`urllib.parse.urlsplit` handles bracketed IPv6 correctly, returns `hostname` already unbracketed, and rejects out-of-range ports.

**Node** has `URL` built in; **Go** has `net/url` and `net.SplitHostPort` — both handle IPv6 brackets natively.

## Related minor issues found while investigating

These are smaller and could be folded into the same fix or filed separately:

1. **Leading whitespace in host accepted silently.** `_parse_cli_url(" localhost:8080")` returns `(" localhost", 8080)` — the literal space is preserved as part of the host, then fails at `socket.connect()` with a confusing DNS error. Should `strip()` or reject.

2. **Misleading error category for URL paths.** `_parse_cli_url("http://localhost:8080/path")` returns `"Invalid port in cli_url"` — the actual problem is the trailing `/path`, not the port.

3. **Empty host with colon prefix accepted.** `_parse_cli_url(":8080")` returns `("localhost", 8080)`. Possibly intentional, but undocumented and inconsistent with the spec.

## Verification script

Self-contained reproducer covering all cases above. Save as `probe.py` inside a checkout of `github/copilot-sdk` and run from the repo root:

```python
"""Demonstrates the IPv6 parsing gap in _parse_cli_url across all input forms."""
import os, sys
sys.path.insert(0, os.path.join(os.getcwd(), "python"))

from copilot.client import CopilotClient, ExternalServerConfig

cases = [
    # IPv4 baseline (works)
    ("127.0.0.1:8080",             "IPv4 baseline"),
    ("localhost:8080",             "hostname baseline"),
    # IPv6 — the gap
    ("[::1]:8080",                 "IPv6 loopback, bracketed (RFC 3986 standard form)"),
    ("::1",                        "IPv6 loopback bare"),
    ("fe80::1:8080",               "IPv6 link-local with port"),
    ("http://[::1]:8080",          "IPv6 with http:// prefix"),
    ("[2001:db8::1]:443",          "IPv6 documentation prefix"),
    # Minor adjacent issues
    (" localhost:8080",            "leading space in host (silently accepted, breaks later)"),
    ("http://localhost:8080/path", "URL with path (misleading error)"),
    (":8080",                      "empty host (silently defaults to localhost)"),
]

c = CopilotClient(ExternalServerConfig(url="localhost:1"))
print(f"{'INPUT':<32s} {'RESULT':<55s} NOTE")
print(f"{'-'*32} {'-'*55} ----")
for url, note in cases:
    try:
        h, p = c._parse_cli_url(url)
        result = f"({h!r}, {p})"
    except Exception as e:
        result = f"{type(e).__name__}: {e}"
    print(f"{url!r:<32s} {result:<55s} {note}")
```

Expected output:

```
INPUT                            RESULT                                                  NOTE
-------------------------------- ------------------------------------------------------- ----
'127.0.0.1:8080'                 ('127.0.0.1', 8080)                                     IPv4 baseline
'localhost:8080'                 ('localhost', 8080)                                     hostname baseline
'[::1]:8080'                     ValueError: Invalid cli_url format: [::1]:8080          IPv6 loopback, bracketed (RFC 3986 standard form)
'::1'                            ValueError: Invalid cli_url format: ::1                 IPv6 loopback bare
'fe80::1:8080'                   ValueError: Invalid cli_url format: fe80::1:8080        IPv6 link-local with port
'http://[::1]:8080'              ValueError: Invalid cli_url format: http://[::1]:8080   IPv6 with http:// prefix
'[2001:db8::1]:443'              ValueError: Invalid cli_url format: [2001:db8::1]:443   IPv6 documentation prefix
' localhost:8080'                (' localhost', 8080)                                    leading space in host (silently accepted, breaks later)
'http://localhost:8080/path'     ValueError: Invalid port in cli_url: http://localhost:8080/path URL with path (misleading error)
':8080'                          ('localhost', 8080)                                     empty host (silently defaults to localhost)
```

xianjianlf2:
Fixes #1137

## Summary
- support bracketed IPv6 runtime URLs across Node.js, Python, Go, Java, and .NET
- validate bracketed host payloads as IPv6 in Node.js, Python, and Go, including rejection of malformed values such as [not-ipv6]:1234
- add positive and negative URL parsing regression coverage for the affected SDKs while preserving existing non-IPv6 parser behavior

## Validation
- cd nodejs && npm test -- test/client.test.ts -t "URL parsing" — 13 passed
- cd nodejs && npm run lint && npm run format:check && npm run typecheck — passed; three existing any warnings remain
- cd python && uv run pytest test_client.py -q — 117 passed
- cd go && go test . ./cmd/bundler ./embeddedcli ./internal/embeddedcli ./internal/ffihost ./internal/flock ./internal/jsonrpc2 ./internal/truncbuffer ./rpc — passed
- Java and .NET checks are left to CI because no Java Runtime or dotnet SDK is installed locally


SteveSandersonMS:
The underlying gap is worth fixing, but I think this PR should be narrower so it adds IPv6 without changing the existing endpoint grammar or taking on general URL-validation policy.

A minimal version would be:

- **Node:** after the existing `http(s)://` removal, recognize the canonical `[ipv6]:port` form before falling back to the current `host:port` parser. Avoid moving all inputs through `URL`.
- **Python:** add the same narrow bracketed-form branch and retain the existing parser for all other inputs. Also change `_connect_via_tcp` from an explicitly IPv4 `socket.AF_INET` socket to family-independent connection resolution (for example, `socket.create_connection`). At present this PR parses `[::1]:4321`, but `start()` still cannot connect to it because it constructs an IPv4-only socket.
- **Go:** after removing the existing schemes, use `net.SplitHostPort` for bracket-prefixed inputs and otherwise retain the current behavior. `net/url` is not needed for this fix.
- **.NET:** `Uri` already parses bracketed IPv6; normalize the parsed host to the unbracketed value if necessary and add regression coverage.
- **Java:** confirm its existing `URI` path with a regression test and change production code only if the host passed to `Socket` remains bracketed.
- **Rust:** no change; its external transport already accepts host and port as separate typed fields.

The intended new cases are simply:

```ts
RuntimeConnection.forUri("[::1]:4321");
RuntimeConnection.forUri("http://[2001:db8::1]:4321");
```

Tests should cover those forms in each URL-based SDK, plus one Python connector-level test proving the connection path is not forced to `AF_INET`. Parser-only assertions are insufficient there because they currently pass while the actual connection remains broken.

I would leave path/query/fragment rejection, whitespace normalization, arbitrary-scheme handling, userinfo policy, and unrelated port-validation changes out of this PR. Moving wholesale to general-purpose URL parsers changes which previously accepted inputs succeed or fail and introduces cross-language differences that then require us to define and maintain a much larger URL contract. A narrow bracketed-IPv6 branch fixes the demonstrated user problem, preserves compatibility, and avoids pulling those edge cases into scope.

SteveSandersonMS:
@xianjianlf2 Does that sound reasonable? I'll mark this PR as draft but please remark as ready to review as appropriate.

xianjianlf2:
Updated in 44894dec after maintainer feedback:

- Narrowed Node.js, Python, and Go parsing to add only canonical bracketed IPv6 host:port handling while preserving existing fallback behavior.
- Switched Python TCP connection setup to socket.create_connection for family-independent IPv4/IPv6 resolution and added connector-level coverage.
- Normalized IPv6 hosts before socket connection in .NET and Java.
- Removed the broad path/query/fragment validation changes.
- Rebases the branch onto main and pushed the update.

Local validation passed for the Go unit packages, Node typecheck/format/lint, Python parsing/connection tests, and AIRI display tests. Please re-review.

nodejs/src/client.ts
```diff
@@ -765,22 +765,33 @@ export class CopilotClient {
...
 
+        // Handle the canonical bracketed IPv6 host:port form without changing
+        // the existing parser behavior for other inputs.
+        const ipv6Match = cleanUrl.match(/^\[([^\]]+)\]:(\d+)$/);
+        if (ipv6Match) {
+            const port = parseInt(ipv6Match[2], 10);
+            if (isNaN(port) || port <= 0 || port > 65535) {
+                throw new Error(`Invalid port in cliUrl: ${url}`);
+            }
+            return { host: ipv6Match[1], port };
```
SteveSandersonMS: This regex accepts any non-`]` text, so `[not-ipv6]:1234` becomes host `not-ipv6`. That conflicts with the documented `[ipv6]:port` grammar and with .NET/Java, which reject this input. Validate `ipv6Match[1]` with Node's `net.isIPv6` before returning, and add a negative regression test. Scoped literals such as `fe80::1%eth0` should remain supported.

python/copilot/client.py
```diff
@@ -1654,14 +1651,20 @@ def _parse_cli_url(self, url: str) -> tuple[str, int]:
                 raise ValueError(f"Invalid port in cli_url: {url}")
             return ("localhost", port)
 
-        # Parse host:port format
-        parts = clean_url.split(":")
-        if len(parts) != 2:
-            raise ValueError(f"Invalid cli_url format: {url}")
+        ipv6_match = re.match(r"^\[([^\]]+)\]:(.*)$", clean_url)
+        if ipv6_match:
+            host = ipv6_match.group(1)
```
SteveSandersonMS, line 1865: The bracket payload is treated as IPv6 without validation, so `[not-ipv6]:1234` is accepted. Validate it with Python's standard IPv6 parser (while preserving scoped literals) and add a negative regression test. This should match the .NET/Java behavior for the same public endpoint syntax.

go/client.go
```diff
@@ -372,29 +372,41 @@ func setEnvValue(env []string, key string, value string) []string {
...
+	if strings.HasPrefix(cleanURL, "[") {
+		host, portStr, err := net.SplitHostPort(cleanURL)
+		if err != nil {
+			panic(fmt.Sprintf("Invalid port in URIConnection: %s", url))
+		}
+		port, err := strconv.Atoi(portStr)
+		if err != nil || port <= 0 || port > 65535 {
+			panic(fmt.Sprintf("Invalid port in URIConnection: %s", url))
+		}
+		return host, port
```
SteveSandersonMS, line 427: `net.SplitHostPort` only splits the bracketed authority; it does not prove `host` is IPv6, so `[not-ipv6]:1234` is accepted. Validate the returned host with `net/netip.ParseAddr` and require `Is6()` before returning, preserving IPv6 zones, then add a negative regression test.

SteveSandersonMS (changes requested):
The narrow approach is now appropriate and the Python transport fix is necessary, but the new bracketed syntax is not consistently validated yet. Please address the three inline comments and add a negative regression case for `[not-ipv6]:1234` in Node, Python, and Go.

Also update the PR description before marking it ready again: the summary still says path/query/fragment forms are rejected even though that change was intentionally removed, and it omits Java from the affected SDKs. The branch is currently behind `main` and has no CI results, so update it onto current `main` and let the normal checks complete before requesting re-review.

xianjianlf2:
已处理 @SteveSandersonMS 的 review：Node.js、Python、Go 现在都会验证方括号内确实是 IPv6（保留带 zone 的 IPv6），并新增 `[not-ipv6]:1234` 负例；已同步 `main`，同时更新 PR 描述以包含 Java 和当前实际支持的 URI 形式。

提交：`298f34f8`

本地验证：Node URL parsing 13/13、Node lint/format/typecheck 通过；Python 114/114；Go 非 E2E 单测通过。Java/.NET 留给 CI（本机无 Java Runtime/dotnet SDK）。

xianjianlf2:
@codex review

xianjianlf2:
Updated this branch onto current main and fixed the PR description so it no longer claims the removed path/query/fragment validation behavior.\n\nCurrent local validation:\n- cd nodejs && npm test -- test/client.test.ts -t "URL parsing" (13 passed)\n- cd nodejs && npm run lint (0 errors; 3 existing any warnings)\n- cd nodejs && npm run format:check && npm run typecheck\n- cd python && uv run pytest test_client.py -q (117 passed)\n- cd go && go test . ./cmd/bundler ./embeddedcli ./internal/embeddedcli ./internal/ffihost ./internal/flock ./internal/jsonrpc2 ./internal/truncbuffer ./rpc\n\nThe three requested [not-ipv6]:1234 negative cases remain covered in Node, Python, and Go. Please re-review when CI finishes.

xianjianlf2:
Synced this branch with current main and resolved the GitHub conflict state. The latest upstream Java SDK directory split was merged automatically, preserving the bracketed IPv6 changes across Node, Python, Go, .NET, and Java.

Validation after sync:
- `cd nodejs && npm test -- test/client.test.ts -t "URL parsing"` (13 passed)
- `cd nodejs && npm run format:check && npm run typecheck && npm run lint` (lint completed with the existing 3 `no-explicit-any` warnings, 0 errors)
- `cd python && uv run pytest test_client.py -q` (117 passed)
- `cd go && go test . ./cmd/bundler ./embeddedcli ./internal/embeddedcli ./internal/ffihost ./internal/flock ./internal/jsonrpc2 ./internal/truncbuffer ./rpc`
- `git diff --check`

Local environment note: `dotnet` is not installed and Java Runtime is unavailable on this machine, so .NET/Java validation is left to CI.

xianjianlf2:
Resolved the remaining Python and Go review conversations. The current head keeps the requested `[not-ipv6]:1234` negative coverage and validates bracketed hosts with `ipaddress.IPv6Address` in Python and `netip.ParseAddr(...).Is6()` in Go. Please re-review when convenient.

WH-sa:
> Updated in 44894dec after maintainer feedback:
> 
> - Narrowed Node.js, Python, and Go parsing to add only canonical bracketed IPv6 host:port handling while preserving existing fallback behavior.
> - Switched Python TCP connection setup to socket.create_connection for family-independent IPv4/IPv6 resolution and added connector-level coverage.
> - Normalized IPv6 hosts before socket connection in .NET and Java.
> - Removed the broad path/query/fragment validation changes.
> - Rebases the branch onto main and pushed the update.
> 
> Local validation passed for the Go unit packages, Node typecheck/format/lint, Python parsing/connection tests, and AIRI display tests. Please re-review.

xianjianlf2:
Synced the branch with current `upstream/main` and resolved the remaining conflict in `nodejs/src/client.ts`.

Resolution kept both import requirements: the PR still imports `isIPv6` for bracketed IPv6 validation, while the upstream `isAbsolute` import is preserved.

Validation after sync:
- `cd nodejs && npm test -- test/client.test.ts -t "URL parsing"` (13 passed)
- `cd nodejs && npm run format:check && npm run typecheck && npm run lint` (lint completed with 5 existing `no-explicit-any` warnings, 0 errors)
- `cd python && uv run pytest test_client.py -q` (121 passed)
- `cd go && go test . ./cmd/bundler ./embeddedcli ./internal/embeddedcli ./internal/ffihost ./internal/flock ./internal/jsonrpc2 ./internal/truncbuffer ./rpc`
- `git diff --check`

Current head: `424bb98dee50a27d561f53ba99dbdd7ce3c05c7a`.

Please re-review when convenient.

xianjianlf2:
The requested IPv6 validation and negative regression are present on the current branch: bracket contents are checked with Node `net.isIPv6`, while scoped literals remain supported. I revalidated after the latest upstream merge:

- `npm test -- --run test/client.test.ts` (176 passed)
- `npm run typecheck`
- `npm run lint` (0 errors; 5 existing warnings in unrelated tests)
- `npm run format:check`

The review thread is resolved; could you please re-review when convenient?

xianjianlf2:
Synced this branch with current upstream/main and resolved the nodejs/src/client.ts import conflict, preserving both the IPv6 validation and the upstream path resolution change.\n\nValidation after sync:\n- Node client tests: 191 passed\n- Node typecheck, format check, and lint passed (0 errors; 5 existing warnings)\n- Python client tests: 139 passed\n- Go unit packages passed\n- git diff --check passed\n\nCurrent head: d490cff7. @SteveSandersonMS, please re-review when convenient.

SteveSandersonMS (approved):
The requested changes are addressed, and the trivial Python import-order issue has been fixed. Approving pending successful CI.

PR #2202, opened 2026-08-01
python: make generated package private

### LINKED ISSUE #2048: [v2][Python] Make generated implementation modules private
https://github.com/github/copilot-sdk/blob/main/python/copilot/generated/__init__.py Makes it clear that the `copilot.generated` subpackage is not meant to be public:

https://github.com/github/copilot-sdk/blob/cd2b10b07e8f393b09379dfda6b30f658b2963d2/python/copilot/generated/__init__.py#L3

But the subpackage doesn't start with an underscore to signify it's an implementation detail: `copilot._generated`.

As well, https://github.com/github/copilot-sdk/blob/main/python/copilot/rpc.py and https://github.com/github/copilot-sdk/blob/main/python/copilot/session_events.py which are meant to act as the public API potentially over-expose things. While `copilot.rpc` does export something explicitly:

https://github.com/github/copilot-sdk/blob/cd2b10b07e8f393b09379dfda6b30f658b2963d2/python/copilot/rpc.py#L10-L11

... both files **also** do an `import *` followed by an import of `__all__` from the generated code which lists everything contained in the generated code. E.g.:

https://github.com/github/copilot-sdk/blob/cd2b10b07e8f393b09379dfda6b30f658b2963d2/python/copilot/session_events.py#L9-L10

exposes 330 symbols in the module:

https://github.com/github/copilot-sdk/blob/cd2b10b07e8f393b09379dfda6b30f658b2963d2/python/copilot/generated/session_events.py#L9883-L10215

I'm not sure if all 330 symbols are meant to be exposed, but at least `copilot.generated` should be made private by renaming it to `copilot._generated`. If all of those symbols are meant to be exposed then that one-liner in `copilot.rpc` might as well be removed. If they are not meant to be exposed then the `import __all__` and `import *` should go and you can explicitly set `__all__` in the exporting modules.

omidsaffari:
## Why

`copilot.generated` documents itself as an internal implementation package, but its public-looking name does not follow Python's underscore convention for private modules.

## What

- Rename the generated implementation package to `copilot._generated` and update internal imports.
- Keep the documented `copilot.rpc` and `copilot.session_events` entry points and their current exports unchanged.
- Update the Python generator, codegen workflow path filter, and static-analysis exclusions for the new location.
- Add regression coverage for the private package boundary and preserved public exports.

This intentionally does not curate or otherwise change the existing generated export lists.

The mistakenly exposed `copilot.generated` import path is removed. Consumers should use the documented `copilot.rpc` and `copilot.session_events` modules instead.

Fixes #2048

## Validation

- `uv run pytest --ignore=e2e -q` — 326 passed
- `uv run pytest test_public_api.py test_codegen_type_names.py test_event_forward_compatibility.py test_tools.py test_rpc_generated.py -q` — 64 passed
- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run ty check copilot` — exits successfully with the two diagnostics already present on `main`
- Regenerated Python protocol types and confirmed a second generation produces no diff
- Built the wheel and confirmed it contains `copilot._generated`, `copilot.rpc`, and `copilot.session_events`, with no `copilot.generated` package


SteveSandersonMS:
Thanks for the contribution. We agree that `copilot._generated` would better communicate that this package is an implementation detail. However, removing the existing `copilot.generated` import path would break consumers on a patch release, including imports of types that remain supported through `copilot.rpc` and `copilot.session_events`. We are therefore not going to make this change in the current release line. We will reconsider the package rename and migration strategy as part of the next major version.

PR #2204, opened 2026-08-01
Serialize tool results with Pydantic's JSON mode

### LINKED ISSUE #2203: Python: tool results containing datetime/UUID/Decimal/Enum are reported to the model as a tool failure
### Summary

In the Python SDK, a tool handler that returns a Pydantic model (or a plain `dict`) with a `datetime`, `date`, `UUID`, `Decimal`, `Enum` or `set` field never reaches the model. Instead the model receives:

This happens on every call, and the tool author gets no indication of what went wrong.

### Location

`python/copilot/tools.py:352-360`, in `_normalize_result`:

```python
    # Everything else gets JSON-serialized (with Pydantic model support)
    def default(obj: Any) -> Any:
        if isinstance(obj, BaseModel):
            return obj.model_dump()
        raise TypeError(f"Object of type {type(obj).__name__} is not JSON serializable")

    try:
        json_str = json.dumps(result, default=default)
    except (TypeError, ValueError) as exc:
        raise TypeError(f"Failed to serialize tool result: {exc}") from exc
```

### Root cause

`BaseModel.model_dump()` uses the default `mode="python"`, which leaves `datetime`, `date`, `time`, `UUID`, `Decimal`, `Enum` and `set` as native Python objects. `json.dumps` then hands those objects back to the same `default` hook; they are not `BaseModel` instances, so the hook raises `TypeError`.

`_normalize_result` re-raises, and that raise happens inside `wrapped_handler`'s `try` block (`tools.py:268-280`), which deliberately converts any exception into the redacted generic failure. So the advertised "Pydantic model support" only holds for models whose fields are all JSON primitives.

### Reproduction

```python
import datetime, decimal, uuid
from copilot.tools import _normalize_result

class Issue(BaseModel):
    id: uuid.UUID
    created_at: datetime.datetime
    price: decimal.Decimal

_normalize_result(Issue(
    id=uuid.uuid4(),
    created_at=datetime.datetime.now(),
    price=decimal.Decimal("1.5"),
))
```

Observed:

```
TypeError: Failed to serialize tool result: Object of type UUID is not JSON serializable
```

Through a real `@define_tool` handler the same input yields:

```
result_type: failure
text: Invoking this tool produced an error. Detailed information is not available.
error: Failed to serialize tool result: Object of type UUID is not JSON serializable
```

### Impact

Returning a model with a timestamp or an id is ordinary tool design, so this is easy to hit and hard to diagnose — the redaction is intentional for security, but it also hides the fact that the failure is in the SDK's serializer rather than in the user's handler.

The TypeScript SDK is not affected: `nodejs/src/session.ts:1074-1079` normalizes with `JSON.stringify`, and JS `Date` implements `toJSON()`.

### Suggested fix

Use Pydantic's JSON mode so each field is coerced to a JSON-native type before `json.dumps` sees it:

```diff
-            return obj.model_dump()
+            return obj.model_dump(mode="json")
```

Models whose fields are all JSON primitives serialize exactly as before. Happy to send a PR — I have one ready with a regression test.

For fuller parity with the TypeScript SDK you may also want fallbacks in `default` for `dataclasses.asdict`, `datetime/date/time.isoformat()`, `UUID`/`Decimal` → `str`, `Enum` → `.value` and `set` → `list`, so a plain `dict` carrying those types works too. I left that out of the PR to keep the change focused.

thejesh23:
Fixes #2203

### What's wrong

`_normalize_result`'s `default` hook calls `model_dump()`, which uses the default `mode="python"`. That leaves `datetime`, `date`, `UUID`, `Decimal`, `Enum` and `set` fields as native Python objects, so `json.dumps` passes them straight back into the same hook, which doesn't recognise them and raises `TypeError`.

That raise happens inside `wrapped_handler`'s `try` block, which converts any exception into the deliberately redacted failure result. So a tool returning a perfectly ordinary model:

```python
class Issue(BaseModel):
    id: uuid.UUID
    created_at: datetime.datetime
```

reports `"Invoking this tool produced an error. Detailed information is not available."` to the model on **every** call, and the tool author sees no indication of why.

### The fix

`model_dump(mode="json")` makes Pydantic coerce each field to a JSON-native type before `json.dumps` sees it. Models whose fields are all JSON primitives serialize exactly as before, so this is backwards compatible.

This matches what the TypeScript SDK already does via `JSON.stringify` (JS `Date` implements `toJSON()`).

### Verification

Added `test_pydantic_model_with_non_primitive_fields_is_serialized` to the existing `TestNormalizeResult` class, covering `UUID`, `datetime`, `Decimal` and `StrEnum` in one model.

It **fails on the unfixed code** with the reported error:

```
TypeError: Failed to serialize tool result: Object of type UUID is not JSON serializable
copilot/tools.py:360: TypeError
```

and passes with the change. Full local run:

- `uv run pytest test_tools.py` → **41 passed** (40 before, plus the new test)
- `uv run pytest --ignore=e2e --ignore=test_client.py --ignore=test_commands_and_elicitation.py --ignore=test_e2e_harness_cli_path.py` → **184 passed**
- `uv run ruff check .` → All checks passed
- `uv run ruff format --check` → already formatted

The four ignored modules fail at collection with `RuntimeError: CLI not found for tests` in my environment; I confirmed they fail identically on unmodified `main`, so that is environmental and unrelated to this change.

### Scope

Deliberately minimal — one line of behaviour change plus a test. For fuller parity with the TypeScript SDK you may also want fallbacks in `default` for `dataclasses.asdict`, `datetime/date/time.isoformat()`, `UUID`/`Decimal` → `str`, `Enum` → `.value` and `set` → `list`, so a plain `dict` carrying those types works too. I left that out to keep the diff focused — happy to add it if you'd prefer it here.

### DIFF
--- a/python/copilot/tools.py
+++ b/python/copilot/tools.py
@@ -351,7 +351,11 @@ def _normalize_result(result: Any) -> ToolResult:
     # Everything else gets JSON-serialized (with Pydantic model support)
     def default(obj: Any) -> Any:
         if isinstance(obj, BaseModel):
-            return obj.model_dump()
+            # mode="json" coerces datetime/UUID/Decimal/Enum fields to JSON-native
+            # types. The default mode="python" leaves them as native objects, which
+            # json.dumps then passes back into this hook and we reject as
+            # unserializable -- turning an ordinary tool result into a failure.
+            return obj.model_dump(mode="json")
         raise TypeError(f"Object of type {type(obj).__name__} is not JSON serializable")
 
     try:
--- a/python/test_tools.py
+++ b/python/test_tools.py
@@ -1,6 +1,10 @@
 """Unit tests for define_tool"""
 
+import datetime
+import decimal
+import enum
 import json
+import uuid
 
 import pytest
 from pydantic import BaseModel, ConfigDict, Field, field_validator
@@ -387,6 +391,33 @@ class TestNormalizeResult:
         result = _normalize_result(items)
         parsed = json.loads(result.text_result_for_llm)
         assert parsed == [{"name": "a", "value": 1}, {"name": "b", "value": 2}]
+        assert result.result_type == "success"
+
+    def test_pydantic_model_with_non_primitive_fields_is_serialized(self):
+        class IssueState(enum.StrEnum):
+            OPEN = "open"
+            CLOSED = "closed"
+
+        class Issue(BaseModel):
+            id: uuid.UUID
+            created_at: datetime.datetime
+            price: decimal.Decimal
+            state: IssueState
+
+        issue = Issue(
+            id=uuid.UUID("6b686a99-0000-4000-8000-000000000000"),
+            created_at=datetime.datetime(2026, 8, 1, 1, 49, 10),
+            price=decimal.Decimal("1.5"),
+            state=IssueState.OPEN,
+        )
+        result = _normalize_result(issue)
+        parsed = json.loads(result.text_result_for_llm)
+        assert parsed == {
+            "id": "6b686a99-0000-4000-8000-000000000000",
+            "created_at": "2026-08-01T01:49:10",
+            "price": "1.5",
+            "state": "open",
+        }
         assert result.result_type == "success"
 
     def test_raises_for_unserializable_value(self):


thejesh23:
Correction to the verification section above: I originally ran the suite without the shared test harness, so I reported a partial run (184 passed) and described the four uncollectable modules as environmental.

I've since followed the CONTRIBUTING step I'd skipped — `cd test/harness && npm ci` — and the **full Python suite passes**:

```
uv run pytest
743 passed, 7 skipped in 500.63s
```

Plus the rest of the CI gate from `.github/workflows/python-sdk-tests.yml`:

- `uv run ruff format --check .` → 113 files already formatted
- `uv run ruff check` → All checks passed
- `uv run ty check copilot` → exit 0 (2 pre-existing warnings, in `client.py:4306` and `copilot_request_handler.py:462` — both untouched here, and identical on unmodified `main`)

So there are no outstanding failures on this branch. Apologies for the noise in the original description.

Observed on darwin / Python 3.12.13 / Node v26.5.0, against `6287d1ad`.

SteveSandersonMS:
Thanks for reporting and addressing this. The fix has now been merged in #2225, with expanded coverage for ordinary Enum, date, and set values, so this PR is superseded.

PR #2206, opened 2026-08-01
Stop sendAndWait from emitting an unhandled rejection

### LINKED ISSUE #2205: Node: sendAndWait can emit an unhandled rejection and terminate the host process when session.error arrives mid-send
### Summary

`CopilotSession.sendAndWait` can reject a promise that has no handler attached yet, which Node classifies as an unhandled rejection and — under the default `--unhandled-rejections=throw` — terminates the host application.

A caller cannot defend against it: the rejection is on an internal promise, not on the one `sendAndWait` returns, so even correct `try`/`.catch` handling around the call does not prevent the crash.

### Location

`nodejs/src/session.ts:723-761`:

```ts
        let resolveIdle: () => void;
        let rejectWithError: (error: Error) => void;
        const idlePromise = new Promise<void>((resolve, reject) => {
            resolveIdle = resolve;
            rejectWithError = reject;
        });

        let lastAssistantMessage: AssistantMessageEvent | undefined;

        // Register event handler BEFORE calling send to avoid race condition
        // where session.idle fires before we start listening
        const unsubscribe = this.on((event) => {
            ...
            } else if (event.type === "session.error") {
                const error = new Error(event.data.message);
                error.stack = event.data.stack;
                rejectWithError(error);
            }
        });

        let timeoutId: ReturnType<typeof setTimeout> | undefined;
        try {
            await this.send(options);
            /* … timeoutPromise built here … */
            await Promise.race([idlePromise, timeoutPromise]);
```

### Root cause

`idlePromise` is created before the listener is registered, but its **first consumer** is only attached by the `Promise.race` — which runs after `await this.send(options)`, a full JSON-RPC round trip to the CLI.

The listener is live for that entire window. A `session.error` arriving in it calls `rejectWithError` on a promise with zero handlers. Node's rejection tracker runs at the following checkpoint, well before the `session.send` response lands, and reports it as unhandled.

The existing comment shows the inverse race (subscribing *too late*) was considered; this is the opposite ordering.

### Why the window is reachable in normal use

`session.error` is not reserved for turn-fatal failures. `session.log(message, { level: "error" })` — a public method on this same class, whose own JSDoc example is `await session.log("Connection failed", { level: "error" })` — emits a `session.error` carrying `errorType: "notification"`. The repo's own e2e test asserts exactly that, in `nodejs/test/e2e/session.e2e.test.ts`:

```ts
        expect(byMessage("Error message").type).toBe("session.error");
        expect(byMessage("Error message").data).toEqual({
            errorType: "notification",
            message: "Error message",
        });
```

So a joined client or extension writing one error log line while another caller is mid-`sendAndWait` is enough. MCP servers failing to start, sub-agent errors (`ErrorData` carries `agentId`) and multi-client sessions produce `session.error` at times entirely uncorrelated with the caller's send.

### Reproduction

Mirroring the promise wiring from lines 723-761, on Node v26.5.0:

1. `t=0ms` — caller invokes `sendAndWait({ prompt: "…" })`; listener registered, `send()` RPC dispatched.
2. `t=5ms` — an unrelated `session.error` arrives.
3. The `session.send` response has not returned yet, so `Promise.race` has not been reached.

```
Error: MCP server failed to start
Node.js v26.5.0
$ echo $?
1
```

The process dies. In the repro the caller does `sendAndWait().then(v => …, e => …)` — correct handling — and it makes no difference.

### Impact

An error-level log line from an unrelated participant in the session can take down the embedding application. Since the SDK is meant to be embedded in host apps, this is a liveness issue rather than just a noisy warning.

The other SDKs are structurally immune, for what it's worth: Go (`go/session.go:502-505`) uses a buffered `errCh`; Python (`python/copilot/session.py:1729-1731`) sets a variable plus `idle_event.set()`.

### Suggested fix

Attach a no-op `catch` when the promise is created. It marks the promise handled without consuming the rejection, so the `Promise.race` still observes it and `sendAndWait` rejects with the original error exactly as before:

```diff
         const idlePromise = new Promise<void>((resolve, reject) => {
             resolveIdle = resolve;
             rejectWithError = reject;
         });
+        void idlePromise.catch(() => {});
```

Happy to send a PR — I have one ready with a regression test that fails on the current code.

thejesh23:
Fixes #2205

### What's wrong

`sendAndWait` creates `idlePromise` and registers the event listener that can reject it, but the first consumer is only attached by the `Promise.race` further down — after `await this.send(options)`, a full JSON-RPC round trip to the CLI.

A `session.error` arriving in that window therefore rejects a promise with no handler. Node's rejection tracker runs at the following checkpoint, well before the `session.send` response lands, and classifies it as unhandled — which terminates the process under the default `--unhandled-rejections=throw`.

The existing comment above the listener shows the inverse race (subscribing *too late*) was considered; this is the opposite ordering.

### Why it's reachable in normal use

`session.error` is not reserved for turn-fatal failures. `session.log(msg, { level: "error" })` emits one with `errorType: "notification"` — asserted in `test/e2e/session.e2e.test.ts`. So a joined client or extension writing an error log line while another caller is mid-`sendAndWait` is enough; MCP servers failing to start and sub-agent errors do the same.

A caller cannot defend against this: the rejection is on the internal promise, not on the one `sendAndWait` returns, so even correct `try`/`.catch` handling around the call does not prevent the crash.

### The fix

A no-op `catch` marks the promise handled without consuming the rejection, so the `Promise.race` still observes it and `sendAndWait` rejects with the original error exactly as before. No behaviour change on any other path.

### Verification

Added `test/session-send-and-wait.test.ts`, which holds the `session.send` RPC open, dispatches a `session.error` into the window, and asserts both that no `unhandledRejection` fires and that `sendAndWait` still rejects with the original message.

It **fails on the unfixed code**:

```
- []
+ [ Error { "message": "MCP server failed to start" } ]

 ❯ test/session-send-and-wait.test.ts:56:27
```

and passes with the change. Full local run:

- `npx vitest run test/ --exclude "test/e2e/**" --exclude "test/cjs-compat.test.ts"` → **14 files, 359 passed**
- `npx tsc --noEmit` → clean
- `npm run lint` → 0 errors (3 pre-existing warnings, all in `test/shared-codegen.test.ts`, none in the files touched here)

On the full `npm test` my environment goes from **6 failed / 371 passed** on unmodified `main` to **5 failed / 372 passed** with this change — the delta is exactly the new test. The remaining failures are `test/cjs-compat.test.ts` (needs a built `dist/`, which I didn't produce) and the `test/e2e/**` suites (need the real `copilot` CLI); I verified both fail identically on unmodified `main`.

### Scope

One line of behaviour change plus a test. Note the other SDKs are already structurally immune — Go (`go/session.go:502-505`) uses a buffered `errCh` and Python (`python/copilot/session.py:1729-1731`) sets a variable plus `idle_event.set()` — so this brings Node in line rather than introducing a new pattern.

### DIFF
--- a/nodejs/src/session.ts
+++ b/nodejs/src/session.ts
@@ -726,6 +726,13 @@ export class CopilotSession {
             resolveIdle = resolve;
             rejectWithError = reject;
         });
+        // A `session.error` can arrive while `send()`'s RPC is still in flight —
+        // that is, before the `Promise.race` below attaches the first consumer to
+        // `idlePromise`. Mark the promise handled now so such a rejection can never
+        // surface as an unhandled rejection, which terminates the process under
+        // Node's default `--unhandled-rejections=throw`. This extra `catch` does
+        // not consume the rejection: the `race` below still sees and rethrows it.
+        void idlePromise.catch(() => {});
 
         let lastAssistantMessage: AssistantMessageEvent | undefined;
 
--- a/nodejs/src/session.ts
+++ b/nodejs/src/session.ts
@@ -720,19 +720,11 @@ export class CopilotSession {
             typeof optionsOrPrompt === "string" ? { prompt: optionsOrPrompt } : optionsOrPrompt;
         const effectiveTimeout = timeout ?? 60_000;
 
-        let resolveIdle: () => void;
-        let rejectWithError: (error: Error) => void;
-        const idlePromise = new Promise<void>((resolve, reject) => {
-            resolveIdle = resolve;
-            rejectWithError = reject;
+        type SessionOutcome = { kind: "idle" } | { kind: "error"; error: Error };
+        let resolveOutcome: (outcome: SessionOutcome) => void;
+        const outcomePromise = new Promise<SessionOutcome>((resolve) => {
+            resolveOutcome = resolve;
         });
-        // A `session.error` can arrive while `send()`'s RPC is still in flight —
-        // that is, before the `Promise.race` below attaches the first consumer to
-        // `idlePromise`. Mark the promise handled now so such a rejection can never
-        // surface as an unhandled rejection, which terminates the process under
-        // Node's default `--unhandled-rejections=throw`. This extra `catch` does
-        // not consume the rejection: the `race` below still sees and rethrows it.
-        void idlePromise.catch(() => {});
 
         let lastAssistantMessage: AssistantMessageEvent | undefined;
 
@@ -742,11 +734,11 @@ export class CopilotSession {
             if (event.type === "assistant.message") {
                 lastAssistantMessage = event;
             } else if (event.type === "session.idle") {
-                resolveIdle();
+                resolveOutcome({ kind: "idle" });
             } else if (event.type === "session.error") {
                 const error = new Error(event.data.message);
                 error.stack = event.data.stack;
-                rejectWithError(error);
+                resolveOutcome({ kind: "error", error });
             }
         });
 
@@ -765,7 +757,10 @@ export class CopilotSession {
                     effectiveTimeout
                 );
             });
-            await Promise.race([idlePromise, timeoutPromise]);
+            const outcome = await Promise.race([outcomePromise, timeoutPromise]);
+            if (outcome.kind === "error") {
+                throw outcome.error;
+            }
 
             return lastAssistantMessage;
         } finally {
--- a/nodejs/test/session-send-and-wait.test.ts
+++ b/nodejs/test/session-send-and-wait.test.ts
@@ -7,6 +7,17 @@ import type { MessageConnection } from "vscode-jsonrpc/node.js";
 import { CopilotSession } from "../src/session.js";
 import type { SessionEvent } from "../src/generated/session-events.js";
 
+function sessionEvent(type: "session.idle", data: Record<string, never> = {}): SessionEvent {
+    return {
+        type,
+        id: "00000000-0000-4000-8000-000000000001",
+        parentId: null,
+        timestamp: new Date().toISOString(),
+        ephemeral: true,
+        data,
+    } as SessionEvent;
+}
+
 /** Builds a `session.error` event, the shape `session.log(…, { level: "error" })` produces. */
 function errorEvent(message: string): SessionEvent {
     return {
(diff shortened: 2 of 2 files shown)


thejesh23:
Correction to the verification section above: the "6 failed / 371 passed → 5 failed / 372 passed" numbers were from an incomplete local setup — I had not run `npm run build` (so `test/cjs-compat.test.ts` had no `dist/` to load) nor the shared harness install that CONTRIBUTING asks for (`cd test/harness && npm ci`), so every `test/e2e/**` suite failed at import with `Cannot find package 'openai'`.

With both done, the **full Node.js suite passes on this branch**:

```
npm test
Test Files  86 passed (86)
     Tests  797 passed | 8 skipped (805)
```

The rest of the gate from `.github/workflows/nodejs-sdk-tests.yml` is green too:

- `npm run format:check` → All matched files use Prettier code style
- `npm run lint` → 0 errors (3 pre-existing warnings, all in `test/shared-codegen.test.ts`, none in files touched here)
- `npm run typecheck` → clean
- `npm run build` → clean

So there are no outstanding failures — the earlier numbers reflected my environment, not this change. Apologies for the confusion.

Observed on darwin / Node v26.5.0, against `6287d1ad`.

SteveSandersonMS:
I added a follow-up in [`90be27b`](https://github.com/github/copilot-sdk/pull/2206/commits/90be27b14bbf95a27d29d6941124524f9da27e8f) to avoid the no-op `catch` workaround.

`sendAndWait` now represents the first `session.idle` or `session.error` as a resolved, discriminated outcome. After `session.send` succeeds, it returns for idle or explicitly throws the stored error. This preserves the existing ordering and semantics: `session.send` failures still take precedence, early idle is retained without completing before `send`, the first session outcome wins, and the timeout still starts only after `send` succeeds. The internal event promise never rejects, so an error arriving while the RPC is in flight cannot produce an unhandled rejection.

The focused tests now cover the original crash plus early-idle behavior, send-error precedence, and first-event settlement. I also ran the broader Node unit suite (362 tests), typecheck, lint, formatting, and a standalone consumer-style reproduction.

SteveSandersonMS (approved):

PR #2208, opened 2026-08-01
dotnet: update README attachment examples to current API (fixes #2196)

### LINKED ISSUE #2196: .NET README attachment examples reference types that no longer exist

HindzStark:
…hmentBlob; fix Session example (fixes #2196)

### DIFF
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -296,9 +296,9 @@ The SDK supports image attachments via the `Attachments` parameter. You can atta
 await session.SendAsync(new MessageOptions
 {
     Prompt = "What's in this image?",
-    Attachments = new List<UserMessageDataAttachmentsItem>
+    Attachments = new List<Attachment>
     {
-        new UserMessageDataAttachmentsItemFile
+        new AttachmentFile
         {
             Path = "/path/to/image.jpg",
             DisplayName = "image.jpg",
@@ -310,9 +310,9 @@ await session.SendAsync(new MessageOptions
 await session.SendAsync(new MessageOptions
 {
     Prompt = "What's in this image?",
-    Attachments = new List<UserMessageDataAttachmentsItem>
+    Attachments = new List<Attachment>
     {
-        new UserMessageDataAttachmentsItemBlob
+        new AttachmentBlob
         {
             Data = base64ImageData,
             MimeType = "image/png",
@@ -721,13 +721,12 @@ await session2.SendAsync(new MessageOptions { Prompt = "Hello from session 2" })
 await session.SendAsync(new MessageOptions
 {
     Prompt = "Analyze this file",
-    Attachments = new List<UserMessageDataAttachmentsItem>
+    Attachments = new List<Attachment>
     {
-        new UserMessageDataAttachmentsItem
+        new AttachmentFile
         {
-            Type = UserMessageDataAttachmentsItemType.File,
             Path = "/path/to/file.cs",
-            DisplayName = "My File"
+            DisplayName = "My File",
         }
     }
 });
--- a/dotnet/src/Session.cs
+++ b/dotnet/src/Session.cs
@@ -1,4 +1,4 @@
-﻿/*---------------------------------------------------------------------------------------------
+/*---------------------------------------------------------------------------------------------
  *  Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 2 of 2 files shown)


SteveSandersonMS (approved):
Thanks @HindzStark!

PR #2225, opened 2026-08-03
fix(python): serialize Pydantic models with mode='json' in tool results

rinceyuan:
## Summary

Fix Python tool handlers returning Pydantic models containing `datetime`, `date`, `UUID`, `Decimal`, `Enum`, or `set` fields being silently reported as failures.

## Root cause

`BaseModel.model_dump()` defaults to `mode="python"`, which keeps native Python types. `json.dumps` then fails on those values, and the error is surfaced as a generic tool failure.

## Fix

Use `model_dump(mode="json")` so Pydantic converts model fields to JSON-safe values before `json.dumps` runs.

## Test

Added `test_pydantic_model_with_non_primitive_fields_is_serialized`, covering the exact JSON representations of UUID, datetime, date, Decimal, Enum, and set fields.

This addresses the Pydantic-model case in #2203. Plain dictionaries containing native non-JSON types remain outside this focused change.

### DIFF
--- a/python/copilot/tools.py
+++ b/python/copilot/tools.py
@@ -351,7 +351,7 @@ def _normalize_result(result: Any) -> ToolResult:
     # Everything else gets JSON-serialized (with Pydantic model support)
     def default(obj: Any) -> Any:
         if isinstance(obj, BaseModel):
-            return obj.model_dump()
+            return obj.model_dump(mode="json")
         raise TypeError(f"Object of type {type(obj).__name__} is not JSON serializable")
 
     try:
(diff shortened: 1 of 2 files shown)


rinceyuan:

rinceyuan:
@SteveSandersonMS One-line Python fix + test. Tool results with datetime/UUID/Decimal/Enum were silently failing because model_dump() defaults to mode='python'. Tests pass.

SteveSandersonMS (approved):
Thanks

PR #2248, opened 2026-08-04
chore(ci): bump Go 1.22 to 1.24 in codegen-check and update-copilot-dependency workflows

rinceyuan:
Two workflows still use Go 1.22 while go.mod requires Go 1.24 and other workflows already use 1.24. Changes: codegen-check.yml and update-copilot-dependency.yml go-version 1.22 to 1.24.


SteveSandersonMS:
Thanks, but please leave it for dependabot to do routine version bumps unless it's specifically causing an immediate problem we need to solve.

PR #2247, opened 2026-08-04
docs: fix stale version requirements and add Rust/Java sections to CONTRIBUTING.md

rinceyuan:
### Changes

- **Node.js**: \18+\ → \20.19+ or v22.12+\ (matches \package.json\ engines field)
- **Python**: \3.8+\ → \3.11+\ (matches \pyproject.toml\ requires-python)
- **Rust SDK**: add prerequisites and test command (was missing entirely)
- **Java SDK**: add prerequisites and test command (was missing entirely)

The repo has six SDKs but CONTRIBUTING.md only covered four, and two of those listed outdated version requirements.

PR #2253, opened 2026-08-04
docs: move SDK development guidance to local READMEs

SteveSandersonMS:
## Summary

- keep the top-level contribution guide focused on shared policy and links to each SDK README
- add only the prerequisites and commands needed to install dependencies and run each SDK's tests
- use the existing per-language manifests and READMEs as the source of version requirements

This is a smaller alternative to #2247 that avoids duplicating mutable language-specific instructions in `CONTRIBUTING.md`.

## Validation

- ran the documented Node.js workflow, including E2E tests
- verified Python setup with `uv sync` and ran its formatting, linting, and type checks
- checked every development command block for valid shell syntax and every top-level README link/anchor

### DIFF
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -1054,7 +1054,7 @@ npm ci
 
 ```bash
 cd dotnet
-GITHUB_ACTIONS=true dotnet test
+dotnet test
 ```
 
 ## License
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -1038,6 +1038,22 @@ catch (Exception ex)
 }
 ```
 
+## Development
+
+Development requires [.NET SDK 10+](https://dotnet.microsoft.com/download) and a supported [Node.js version](../nodejs/README.md#prerequisites) for the runtime and shared test harness. From the repository root:
+
+```bash
+cd nodejs
+npm ci --ignore-scripts
+cd ../test/harness
+npm ci --ignore-scripts
+cd ../../dotnet
+dotnet restore
+dotnet format --verify-no-changes
+dotnet build --no-restore
+dotnet test --no-build -v n
+```
+
 ## License
 
 MIT
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -1043,9 +1043,18 @@ catch (Exception ex)
 Development requires [.NET SDK 10+](https://dotnet.microsoft.com/download). From the repository root:
 
 ```bash
-(cd nodejs && npm ci)
-(cd test/harness && npm ci)
-cd dotnet && dotnet test
+cd nodejs
+npm ci
+```
+
+```bash
+cd test/harness
+npm ci
+```
+
+```bash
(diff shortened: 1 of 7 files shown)


dotnet/README.md
```diff
@@ -1038,6 +1038,16 @@ catch (Exception ex)
...
 
+## Development
+
+Development requires [.NET SDK 10+](https://dotnet.microsoft.com/download). From the repository root:
+
+```bash
+(cd nodejs && npm ci)
+(cd test/harness && npm ci)
+cd dotnet && dotnet test
+```
```
SteveSandersonMS, line 1058: I don't know why you've use parens here. Make this simpler, like people normally write commands. No need for && either. Just have 3 separate blocks.

java/README.md
```diff
@@ -420,25 +420,12 @@ The gate also applies to individual methods annotated with `@CopilotExperimental
...
-git config core.hooksPath .githooks
-
-# Build and test with JDK 25
-mvn test-compile jar:jar
-mvn verify -Dskip.test.harness=true
-
-# Set your paths for JDK 17
-# Run the JDK 25 built jar with JDK 17 JVM for tests. Do not re-compile the jar.
-mvn jacoco:prepare-agent@wire-up-coverage-instrumentation antrun:run@print-test-jdk-banner surefire:test failsafe:integration-test failsafe:verify jacoco:report@build-coverage-report-from-tests -Denforcer.skip=true
+cd java && mvn verify
```
SteveSandersonMS: There probably was a reason why all these instructions used to be here. Don't remove things that are likely for a reason.

dotnet/README.md
SteveSandersonMS: Updated the new instructions across all SDKs to use separate, conventional command blocks with no subshells or command chaining.

java/README.md
SteveSandersonMS: Restored the existing Java development guidance exactly and changed the top-level link to its existing development-setup anchor.

dotnet/README.md
```diff
@@ -1038,6 +1038,25 @@ catch (Exception ex)
...
+
+```bash
+cd test/harness
+npm ci
+```
+
+```bash
+cd dotnet
+GITHUB_ACTIONS=true dotnet test
```
SteveSandersonMS: That's not cross-platform and we do not tell people to use GITHUB_ACTIONS=true in local development normally.

SteveSandersonMS:
Addressed the valid review feedback by linking Python, Go, .NET, Rust, and Java development setup to the authoritative Node.js prerequisite. The Rust `test-support` command was already corrected in the preceding review-fix commit.

dotnet/README.md
SteveSandersonMS: Removed the inline environment assignment from Node, Python, Go, and .NET so all documented local test commands remain conventional and cross-platform.

PR #2262, opened 2026-08-04
Update multiple dependencies including flatted, lodash, and rollup

iAlien369-design:


SteveSandersonMS:
Thanks but we'll leave it to dependabot to update these when needed.

PR #2290, opened 2026-08-07
Rust: add race-safe canvas availability waiter

ellismg:
## Summary

- add `Session::wait_for_canvas(WaitForCanvasOptions)` for waiting on a known canvas declaration without sending a model turn
- subscribe before the initial `session.canvas.list` snapshot, then re-list on registry changes or subscription lag
- support provider-specific matching, total timeouts, and deterministic session-shutdown errors
- cover existing declarations, registration races, lag recovery, timeout/no-turn behavior, and shutdown during both event waiting and an in-flight list request

## Motivation

github/github-app#11029 currently has to coordinate `session.canvas.list` and `session.canvas.registry_changed` itself while asynchronously loaded extensions register their canvases. This helper provides the smallest SDK-level primitive that removes the list/subscription race when the caller already knows the target canvas.

The API deliberately guarantees only target availability. It does **not** claim that extension initialization or the complete canvas registry has settled. A handshake-backed bounded initialization snapshot remains a separate copilot-agent-runtime protocol follow-up for github/github-app#7544 and github/github-app#10224.

## Compatibility

This is an additive Rust API implemented entirely over existing protocol-v3 methods and events. It does not change `session.create`, add wire fields, block session creation on extension startup, or require a model turn. Older runtimes therefore need no protocol changes.

`CHANGELOG.md` is release-generated and has no unreleased section, so this PR relies on public rustdoc rather than editing it manually.

## Validation

- `cargo +nightly-2026-04-14 fmt --all -- --config-path .rustfmt.nightly.toml --check`
- `cargo clippy --all-targets --features test-support -- --no-deps -D warnings -D clippy::unwrap_used -D clippy::disallowed_macros -D clippy::await_holding_invalid_type`
- `cargo test --features test-support`

### DIFF
--- a/rust/src/errors.rs
+++ b/rust/src/errors.rs
@@ -131,6 +131,24 @@ pub enum SessionErrorKind {
     /// The session event loop exited before a pending `send_and_wait` completed.
     EventLoopClosed,
 
+    /// A requested canvas did not become available before the wait timeout.
+    CanvasWaitTimeout {
+        /// Provider-local canvas identifier.
+        canvas_id: String,
+        /// Owning provider identifier, when the wait was provider-specific.
+        extension_id: Option<String>,
+        /// Configured wait timeout.
+        timeout: Duration,
+    },
+
+    /// The session shut down while waiting for a canvas.
+    CanvasWaitSessionClosed {
+        /// Provider-local canvas identifier.
+        canvas_id: String,
+        /// Owning provider identifier, when the wait was provider-specific.
+        extension_id: Option<String>,
+    },
+
     /// Elicitation is not supported by the host.
     /// Check `session.capabilities().ui.elicitation` before calling UI methods.
     ElicitationNotSupported,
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
I’m moving this PR to draft for tracking. Please address these before marking it ready for review:

1. Fix the shutdown path so an in-flight `wait_for_canvas` returns `CanvasWaitSessionClosed` promptly when `Client::stop()` or `Client::force_stop()` tears down the runtime; it currently waits until its configured timeout. Add an integration/E2E test that starts a wait, performs client-level shutdown (not only `Session::stop_event_loop()`), and asserts the typed shutdown error is returned promptly.
2. This is a language-neutral capability: the same `session.canvas.list` and `session.canvas.registry_changed` primitives exist in Node, Python, Go, .NET, Java, and Rust. Please add equivalent support across the SDKs, or document and coordinate a concrete cross-language rollout plan with consistent semantics.

For tracking, I've moved this to draft - please mark as ready to review when appropriate.

PR #2294, opened 2026-08-07
sdk: Forward decisionContext on permission replies across languages

aymenfurter:
## Why

The runtime emits `auto_approval_decision` telemetry only when the client sends an explicit `decisionContext` with its permission reply. The wire schema, runtime, and generated SDK types already support the field.

The hand-written permission reply paths in these SDKs did not forward it. Hosts that answered permission requests through an SDK therefore could not tell the runtime whether a decision came from a person, host policy, or an automated recommendation.

## What changed

Permission handlers can now attach optional decision context. The SDK sends it as a top-level sibling of `result` in `session.permissions.handlePendingPermissionRequest`:

- Node: `createAttributedPermissionResult(result, context)`
- Python: `copilot.create_attributed_permission_result(result, context)`
- Go: `copilot.NewAttributedPermissionResult(result, context)`
- .NET: set `DecisionContext` on the permission decision
- Java: `PermissionRequestResult.approveOnce().setDecisionContext(context)`
- Rust: `PermissionResult::approve_once().with_context(context)`

Each SDK follows its existing language conventions. Applying context twice replaces the previous context instead of nesting it. A no-result response remains suppressed.

When a handler does not supply context, the SDK sends the same legacy JSON shape with only `sessionId`, `requestId`, and `result`. It does not send `decisionContext: null`.

No schema or generated code changed. This PR only fills the gap in the hand-written permission reply paths.

### Rust compatibility

Rust keeps `PermissionHandler` as the single handler API. `PermissionResult::Decision` now contains the decision and an optional context:

```rust
PermissionResult::Decision {
    decision: PermissionDecision,
    context: Option<PermissionDecisionContext>,
}
```

This is an intentional Rust source break. Clients that construct or match `PermissionResult::Decision` directly must use the new struct variant. The compiler identifies these sites during migration, including wildcard matches that could otherwise mistake a contextual decision for `NoResult`. Existing handler implementations that use the result helpers and existing session registration remain unchanged. This avoids a parallel attributed handler trait and keeps all decisions in one semantic variant.

The other five SDK changes are additive.

## Testing

- Unit tests in all six SDKs cover context forwarding, omission when absent, no-result behavior, and replacement when context is applied twice.
- The Node E2E test runs the real permission flow and checks the exact params passed to the CLI.
- The Rust fake-server test captures the actual outbound JSON-RPC request and verifies that `decisionContext` is beside `result`, not inside it.
- Rust library and session tests pass. Clippy and formatting are clean.
- The targeted Node, Python, Go, .NET, and Java test suites pass.

## Release note

Permission handlers can attach `decisionContext` so the runtime can attribute permission decisions. This is additive for Node, Python, Go, .NET, and Java. Rust clients that construct or match `PermissionResult::Decision` directly must migrate from the tuple variant to the new struct variant.


java/sdk/src/main/java/com/github/copilot/rpc/PermissionRequestResult.java
```diff
@@ -168,4 +179,35 @@ public PermissionRequestResult setFeedback(String feedback) {
...
+     * @param context
+     *            the decision context, or {@code null} to clear it
+     * @return this result for method chaining
+     * @since 1.3.0
+     */
+    public PermissionRequestResult withContext(PermissionDecisionContext context) {
+        this.decisionContext = context;
+        return this;
+    }
 }
```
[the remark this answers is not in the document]
aymenfurter: Good catch — fixed in the follow-up commit.

Agreed this should be uniform. Worth adding that null is Java's default, so an uninitialized variable would have silently dropped the context and produced exactly the unattributed telemetry this feature exists to remove — a quiet failure rather than a loud one. `Objects.requireNonNull` (already the convention in this package) now rejects it, the Javadoc documents `@throws NullPointerException`, and a `withContextRejectsNull` test covers it.

I did not add a `clearContext()`: no caller needs it today, and adding it to all six SDKs would be speculative API surface.

> Above text was produced by **Copilot** Agent using **Claude Opus** without my direct review

dotnet/src/PermissionDecision.cs
```diff
@@ -43,4 +43,30 @@ public static PermissionDecision Reject(string? feedback = null) =>
...
+    /// Attaches provenance to this decision so the runtime can attribute
+    /// auto-approval telemetry. Returns the same instance mutated in place;
+    /// because the static factories (<see cref="ApproveOnce"/>, <see cref="Reject"/>,
+    /// etc.) return a fresh instance on every call, mutating is safe and keeps
+    /// the fluent call site concise. Calling this more than once replaces the
+    /// previously attached context rather than nesting it.
+    /// </summary>
+    /// <param name="context">The provenance to attach.</param>
+    /// <returns>This decision, for fluent chaining.</returns>
+    public PermissionDecision WithContext(PermissionDecisionContext context)
```
stephentoub: Why is this needed? We don't have that fluent style elsewhere here, do we? And the setter is public.

dotnet/src/PermissionDecision.cs
```diff
@@ -43,4 +43,30 @@ public static PermissionDecision Reject(string? feedback = null) =>
...
     /// </summary>
     public static PermissionDecision NoResult() => new PermissionDecisionNoResult();
+
+    /// <summary>
+    /// Optional provenance describing how and where this decision was made.
+    /// This is never serialized as part of the decision itself: the SDK forwards
+    /// it to the runtime as a sibling of <c>result</c> so that auto-approval
+    /// telemetry can be attributed correctly.
+    /// </summary>
+    [JsonIgnore]
```
stephentoub, line 53: Why is this JsonIgnore? Doesn't that defeat the point?

stephentoub: Oh, I see, we're passing the data in separately. Ok.

dotnet/src/PermissionDecision.cs
stephentoub: Similar question for the other languages. If they already expose other things with the fluent helpers, then ok, but we shouldn't introduce a new pattern.

go/session.go
```diff
@@ -1646,6 +1646,14 @@ func (s *Session) executePermissionAndRespond(requestID string, permissionReques
...
 		return
 	}
+	// Unwrap any attribution so decisionContext travels as a sibling of result,
+	// not nested inside it. The suppression and send logic below operates on the
+	// underlying decision.
+	var decisionContext *rpc.PermissionDecisionContext
+	if attributed, ok := decision.(*AttributedPermissionResult); ok {
+		decisionContext = attributed.DecisionContext
+		decision = attributed.PermissionDecision
+	}
```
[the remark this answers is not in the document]
aymenfurter: Confirmed and fixed — this was a real bug, thank you.

I verified the promotion behaviour in isolation before changing anything: a value-form `Attributed` satisfies the interface, `.(*Attributed)` returns false, and `.(Attributed)` returns true. So `return *WithDecisionContext(...)` would have sent the wrapper as `result` and dropped the context entirely — silently, in the one path whose purpose is telemetry.

Both the session unwrap and the replace-not-nest check in `WithDecisionContext` now handle either form via a type switch. Added `TestValueFormAttributedResultIsUnwrapped`, which I mutation-tested: it fails against the previous pointer-only assertion and passes with the fix.

> Above text was produced by **Copilot** Agent using **Claude Opus 4.5** without my direct review

go/permissions.go
```diff
@@ -6,6 +6,46 @@ import (
...
+//
+// The returned value satisfies [rpc.PermissionDecision], so a
+// [PermissionHandlerFunc] can return it directly. Applying WithDecisionContext
+// to an already-attributed result replaces the previous context rather than
+// nesting it. If result is a [rpc.PermissionDecisionNoResult] (attributed or
+// not), the SDK still suppresses the response.
+//
+// Experimental: WithDecisionContext is part of an experimental API and may
+// change or be removed.
+func WithDecisionContext(result rpc.PermissionDecision, decisionContext *rpc.PermissionDecisionContext) *AttributedPermissionResult {
```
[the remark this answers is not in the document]
aymenfurter: You're right, and the PR body was wrong — the exported API is `copilot.WithDecisionContext`, not `permissions.WithContext`. The code is correct; my description wasn't. PR body updated.

> Above text was produced by **Copilot** Agent using **Claude Opus 4.5** without my direct review

python/copilot/session.py
```diff
@@ -367,14 +368,53 @@ class PermissionNoResult:
...
+    """The underlying permission decision (or :class:`PermissionNoResult`)."""
+
+    decision_context: PermissionDecisionContext
+    """Context describing how and where the decision was reached."""
+
+
+def with_decision_context(
+    result: PermissionRequestResult | AttributedPermissionResult,
+    decision_context: PermissionDecisionContext,
+) -> AttributedPermissionResult:
```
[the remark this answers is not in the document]
aymenfurter: Correct — Python exposes a module-level `with_decision_context(result, ctx)`, not a method on the result. The PR body was wrong; the code matches the Node.js free-function pattern deliberately. PR body updated.

> Above text was produced by **Copilot** Agent using **Claude Opus 4.5** without my direct review

rust/src/handler.rs
```diff
@@ -42,6 +42,13 @@ use crate::types::{
 pub enum PermissionResult {
     /// Send a permission decision on the wire.
     Decision(PermissionDecision),
+    /// Send a permission decision annotated with the context describing how
+    /// and where it was reached, so the runtime can attribute
+    /// auto-approval telemetry to the responding surface.
+    ///
+    /// The context is informational only — it never changes permission
+    /// behavior.
+    AttributedDecision(PermissionDecision, PermissionDecisionContext),
```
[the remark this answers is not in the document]
aymenfurter: Fair point — added `#[non_exhaustive]` to `PermissionResult`, which is the convention already used throughout this crate (41 occurrences in `types.rs`, plus `AutoModeSwitchResponse` in this same file). Downstream exhaustive matches now need a wildcard arm, and future variants won't break them again.

Note the crate is `0.0.0-dev` and the type is marked experimental, so this is the right moment to seal it. `cargo build`, 221 lib tests, clippy with the CI flags and `fmt --check` are all clean.

> Above text was produced by **Copilot** Agent using **Claude Opus 4.5** without my direct review

java/sdk/src/main/java/com/github/copilot/CopilotSession.java
```diff
@@ -1026,7 +1026,7 @@ private void executePermissionAndRespondAsync(String requestId, PermissionReques
                         }
                         getRpc().permissions.handlePendingPermissionRequest(
                                 new SessionPermissionsHandlePendingPermissionRequestParams(sessionId, requestId, result,
-                                        null));
+                                        result.getDecisionContext()));
```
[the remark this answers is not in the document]
aymenfurter: Correct that this specific line isn't covered — the Java tests construct the params record directly, so reverting the argument to `null` would leave them green. I've left it as-is rather than papering over it: the Java suite has no session-level permission-dispatch harness today, and building one is a larger piece of work than this PR should absorb. Tracking it as a follow-up.

Worth noting the equivalent line *is* covered elsewhere — Go, Rust, Node.js and Python all assert the captured RPC params through their session paths, so the wire contract itself is well protected.

> Above text was produced by **Copilot** Agent using **Claude Opus 4.5** without my direct review

dotnet/src/PermissionDecision.cs
aymenfurter: Checked each language against its own existing surface and fixed three:

- **.NET**: removed; callers set the `DecisionContext` property in an object initializer.
- **Go**: now `NewAttributedPermissionResult`, matching `NewCanvasError`/`NewToolSet`. `With*` conventionally builds a functional option in Go.
- **Node**: now `createAttributedPermissionResult`, carrying `kind: "attributed"` so it narrows like every other union here.
- **Python**: now `create_attributed_permission_result`, matching `create_session_fs_adapter`.

Java and Rust left as-is: `PermissionRequestResult` already had four fluent setters, and `with_*` is the dominant Rust idiom.

go/permissions.go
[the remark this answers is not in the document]
aymenfurter: Renamed again since this reply, to `copilot.NewAttributedPermissionResult`, matching the existing `New*` constructors. PR body updated.

python/copilot/session.py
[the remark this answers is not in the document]
aymenfurter: Renamed again since this reply, to `copilot.create_attributed_permission_result`, matching `create_session_fs_adapter`. PR body updated.

java/src/main/java/com/github/copilot/rpc/PermissionRequestResult.java
```diff
@@ -168,4 +180,37 @@ public PermissionRequestResult setFeedback(String feedback) {
...
+     * sibling of {@code result}.
+     *
+     * @param context
+     *            the decision context; must not be {@code null}
+     * @return this result for method chaining
+     * @throws NullPointerException
+     *             if {@code context} is {@code null}
+     * @since 1.3.0
+     */
+    public PermissionRequestResult withContext(PermissionDecisionContext context) {
```
stephentoub: Is this necessary?

aymenfurter: Renamed to `setDecisionContext` to match the getter and the other setters here, and dropped the `requireNonNull` since null now means "no context" everywhere else.

aymenfurter:
@stephentoub I moved this PR back to draft after finding that the current Rust API change breaks existing clients. Adding `AttributedDecision` to `PermissionResult`, or marking the enum `#[non_exhaustive]`, breaks clients that match its existing variants exhaustively.

I tested a backwards-compatible alternative that keeps `PermissionHandler` and `PermissionResult` unchanged. It adds a separate `AttributedPermissionHandler` for clients that need to provide `decisionContext`.

Existing clients continue to use:

```rust
impl PermissionHandler for MyHandler {
    async fn handle(...) -> PermissionResult {
        PermissionResult::approve_once()
    }
}

SessionConfig::default()
    .with_permission_handler(Arc::new(MyHandler))
```

Copilot App would use:

```rust
impl AttributedPermissionHandler for CopilotAppPermissionHandler {
    async fn handle(...) -> AttributedPermissionResult {
        PermissionResult::approve_once().with_context(
            PermissionDecisionContext {
                outcome: PermissionDecisionOutcome::PromptedUser,
                source: PermissionDecisionSource::HumanResponse,
                surface: PermissionDecisionSurface::CopilotApp,
            },
        )
    }
}

let client = Client::start(ClientOptions::default()).await?;

let session = client
    .create_session(
        SessionConfig::default().with_attributed_permission_handler(
            Arc::new(CopilotAppPermissionHandler),
        ),
    )
    .await?;
```

Existing clients compile without changes, and context-aware clients implement only one handler method. The cost is one additional handler trait and configuration method.

stephentoub:
> Adding AttributedDecision to PermissionResult, or marking the enum #[non_exhaustive], breaks clients that match its existing variants exhaustively.

In the past, I believe @tclem has suggested we shouldn't care about such breaking changes for rust consumers. Tim?

aymenfurter:
@tclem What do you think? 👀

rust/tests/e2e/permissions.rs
```diff
@@ -120,6 +124,70 @@ async fn should_deny_permission_when_handler_returns_denied() {
     .await;
 }
 
+#[tokio::test]
+async fn should_honor_a_decision_annotated_with_decisioncontext() {
```
tclem: This test passes even if `decisionContext` forwarding is removed; it only observes the rejection. Please test the outbound JSON-RPC request with the existing fake server in `session_test.rs`, asserting `decisionContext` beside `result`.

<sub><em>&nbsp;&nbsp;Generated via Copilot (GPT-5.6 Sol) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/handler.rs
```diff
@@ -75,6 +97,88 @@ impl PermissionResult {
...
+        Self {
+            result,
+            context: None,
+        }
+    }
+}
+
+/// Handler for permission requests that also reports how the decision was made.
+#[async_trait]
+pub trait AttributedPermissionHandler: Send + Sync + 'static {
```
tclem: Can we avoid introducing a second handler hierarchy? Attribution changes the response metadata, not the dispatch contract. I’d rather evolve `PermissionResult` and keep `PermissionHandler` as the single API. If source compatibility is mandatory, a provided `PermissionHandler::handle_with_context` defaulting to the existing `handle` is still less permanent surface than parallel traits and configuration.

<sub><em>&nbsp;&nbsp;Generated via Copilot (GPT-5.6 Sol) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/handler.rs
```diff
@@ -75,6 +97,88 @@ impl PermissionResult {
...
+    /// #     PermissionDecisionSurface,
+    /// # };
+    ///
+    /// let result = PermissionResult::approve_once().with_context(PermissionDecisionContext {
+    ///     outcome: PermissionDecisionOutcome::AutoApproved,
+    ///     source: PermissionDecisionSource::HostPolicy,
+    ///     surface: PermissionDecisionSurface::Sdk,
+    /// });
+    /// ```
+    pub fn with_context(self, context: PermissionDecisionContext) -> AttributedPermissionResult {
```
tclem: This API is misleading: `PermissionResult::approve_once().with_context(ctx)` no longer returns `PermissionResult`, so it cannot be returned from `PermissionHandler::handle`. Please keep `with_context` returning `Self`; the obvious usage should compile.

<sub><em>&nbsp;&nbsp;Generated via Copilot (GPT-5.6 Sol) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/types.rs
```diff
@@ -2166,6 +2166,8 @@ pub struct SessionConfig {
     /// `requestPermission: false` on the wire so the runtime does not
     /// emit `permission.requested` broadcasts to this client.
     pub permission_handler: Option<Arc<dyn PermissionHandler>>,
+    /// Optional context-aware permission-request handler.
+    pub attributed_permission_handler: Option<Arc<dyn AttributedPermissionHandler>>,
```
tclem: These public fields can both be set through direct assignment, while `resolve_handler` silently prefers the attributed handler. Since direct field assignment is explicitly supported, this introduces an invalid and ambiguous state. A single handler field avoids it.

<sub><em>&nbsp;&nbsp;Generated via Copilot (GPT-5.6 Sol) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

rust/src/handler.rs
```diff
@@ -75,6 +97,88 @@ impl PermissionResult {
...
+        request_id: RequestId,
+        data: PermissionRequestData,
+    ) -> AttributedPermissionResult {
+        PermissionHandler::handle(&*self.0, session_id, request_id, data)
+            .await
+            .into()
+    }
+}
+
+pub(crate) fn attributed(
```
tclem: strange to have a function called `attributed` that returns UnattributedHandler...

aymenfurter:
We decided to make the breaking change for Rust. `PermissionResult` now has an `AttributedDecision` variant, so clients with exhaustive matches must handle the new variant. Existing `PermissionHandler` implementations and session registration remain unchanged. This lets us keep one handler API instead of introducing a parallel attributed handler hierarchy. Tim confirmed that this kind of Rust source break is acceptable.

rust/src/handler.rs
aymenfurter: Agreed. I removed `AttributedPermissionHandler` and kept `PermissionHandler` as the single dispatch API. `PermissionResult` now carries the optional attribution through an `AttributedDecision` variant. This intentionally creates one Rust source break: exhaustive matches on `PermissionResult` must handle the new variant. Existing handler implementations and session registration remain unchanged.

rust/tests/e2e/permissions.rs
aymenfurter: Agreed. I removed the E2E test because it only proved that the rejection was honored. The replacement test uses the existing fake server in `session_test.rs`, captures the outbound JSON-RPC request, and verifies that `decisionContext` is beside `result`.

rust/src/handler.rs
aymenfurter: Removed. The `attributed()` adapter and `UnattributedHandler` type no longer exist.

rust/src/handler.rs
aymenfurter: Fixed. `with_context` now returns `Self`, so `PermissionResult::approve_once().with_context(context)` can be returned directly from `PermissionHandler::handle`. Calling it again replaces the previous context.

rust/src/types.rs
aymenfurter: Fixed. `SessionConfig` and `ResumeSessionConfig` now have only one permission handler field. The attributed field and its precedence rules are gone.

rust/src/handler.rs
```diff
@@ -35,13 +35,33 @@ use crate::types::{
...
+///         PermissionResult::NoResult => false,
+///     }
+/// }
+/// ```
 #[derive(Debug, Clone)]
 pub enum PermissionResult {
     /// Send a permission decision on the wire.
     Decision(PermissionDecision),
+    /// Send a permission decision with context describing how it was reached.
+    AttributedDecision {
```
tclem: Could we extend `Decision` instead of adding a second semantic decision variant?

```rust
Decision {
    decision: PermissionDecision,
    context: Option<PermissionDecisionContext>,
}
```

`AttributedDecision` is still a decision, so every consumer must remember to handle both variants. More importantly, downstream code matching `Decision(_)` with a wildcard arm keeps compiling and may silently treat an attributed decision like `NoResult`. Since we're already taking a breaking change here, I'd prefer the compiler-guided break and the coherent model.

<sub><em>&nbsp;&nbsp;Generated via Copilot (GPT-5.6 Sol) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

tclem (approved):

rust/src/handler.rs
aymenfurter: Agreed. I folded the optional context into `Decision { decision, context }`. Downstream construction and matches now get a compiler-guided migration, while `NoResult` remains the only non-decision state.

PR #2295, opened 2026-08-07
[java] Add linux-x64 implementation of in process Copilot CLI

edburns:
Superceded by #2301

This PR is the roll up of the agentic work done in the subtasks of #2166 . At each step of those subtasks, the CI was clean and all reviews were applied as appropriate.

# PR 2295 — Reviewer's guide: In-process FFI runtime for the Java SDK

## TL;DR

This PR does for the Java SDK what [#1901](https://github.com/github/copilot-sdk/pull/1901) did for .NET and [#1915](https://github.com/github/copilot-sdk/pull/1915) did for Rust: it adds an **in-process connection mode** that loads the Copilot runtime (`runtime.node` cdylib) as a native library via JNA, eliminating the need for a separate CLI child process. Currently scoped to **linux-x64 only**; the entire in-process API surface is marked `@CopilotExperimental`.

The PR also restructures the Java Maven project from a single module into a **multi-module reactor** to support publishing the native runtime binaries as separate classifier JARs alongside the existing SDK JAR.

---

### The binary: `runtime.node`

Despite the `.node` extension (a napi-rs naming convention), `runtime.node` is an ordinary platform-specific shared library (`.so` on Linux). It is a Rust `cdylib` produced by the `src/runtime` crate in `github/copilot-agent-runtime`. It exposes two front doors:

- **napi front door** — loaded by Node.js as a native addon (existing CLI path).
- **C ABI front door** — 5 `extern "C"` lifecycle/transport entry points callable by any language via FFI without Node.js.

The 5 C ABI entry points are:

| Entry point | Purpose |
|---|---|
| `copilot_runtime_host_start` | Start the runtime host. Blocks up to ~30s while the worker boots. Returns a server handle (0 = failure). |
| `copilot_runtime_host_shutdown` | Shut down a runtime host by server handle. |
| `copilot_runtime_connection_open` | Open a bidirectional connection; registers an `on_outbound` callback for runtime→SDK data delivery. |
| `copilot_runtime_connection_write` | Write a JSON-RPC frame from the SDK into the runtime. |
| `copilot_runtime_connection_close` | Close a connection. |

All JSON-RPC methods travel as data through this fixed 5-function transport; the export surface never changes as the method set grows.

### Where it comes from (build-time)

The `copilot-native` Maven module's build fetches the binary from npm during `generate-resources`:

1. **`fetch-native.mjs`** reads the pinned version and SHA-512 integrity hash for `@github/copilot-linux-x64` from `nodejs/package-lock.json`.
[...]

Issue #1917, opened 2026-07-06
[Java] Embed Rust-based Copilot CLI Runtime and cease requiring Node.js


edburns:
As of 2026-07, the Copilot SDK for Java does not embed a Copilot runtime. Rather, it depends on one being available externally. This design decision was made before 2026-02. Now that the Rust Copilot runtime cdylib is nearing the point where it can be consumed as a native shared library, this decision must be revisited.

Work items in this Epic will make it so the Copilot SDK for Java embeds N platform-specific instances of the Rust Copilot runtime native library (`runtime.node`, a Rust cdylib), dynamically selecting and loading the correct library for the current platform at runtime via JNA.

The relevant platform dimensions are:

- OS: Windows, macOS, GNU/Linux
- CPU architecture: x86_64 (x64), aarch64 (arm64)
- Linux only — C runtime variant: glibc (mainstream distros: Debian 10+, Ubuntu 20.04+, RHEL 8+) vs. musl (Alpine Linux)

This yields 6 binaries for the common case (Windows × 2 + macOS × 2 + GNU/Linux × 2), or 8 binaries if Alpine/musl Linux support is also required. The correct binary is selected 100% deterministically at runtime — no heuristics are needed — using `System.getProperty("os.name")`, `System.getProperty("os.arch")`, and (on Linux) inspection of the ELF PT_INTERP segment of `/proc/self/exe` to distinguish glibc from musl.

The library is loaded in-process via JNA using approximately 12 fixed C ABI entry points (`copilot_runtime_server_create`, `copilot_runtime_connection_open`, etc.). All API methods are dispatched as JSON-RPC over this in-process connection; no Node.js process is required by the host. (Note: during the current transitional period while the Rust port is in progress, the runtime library may internally spawn a short-lived Node.js child process to service method bodies not yet ported to Rust. This internal dependency shrinks with each port PR and is expected to disappear entirely once the migration is complete.)

The `cli-native.node` addon — which provides ICU4X text segmentation, Win32 APIs, and terminal UI helpers — is a CLI-only artifact and is not needed by the Java SDK.

This decision is based on the assumption that embedding N platform-specific native libraries remains acceptably compact. Current measured sizes (release `cli-1.0.69-2`): each `runtime.node` binary ranges from 48 to 68 MB uncompressed. The 6-platform set compresses to approximately 132 MB within a JAR (deflate ~40%); the 8-platform set to approximately 180 MB.

[...]

PR #2301, opened 2026-08-09
[java] Add linux-x64 implementation of in process Copilot CLI

edburns:
Supercedes #2295 .

This PR is the roll up of the agentic work done in the subtasks of #2166 . At each step of those subtasks, the CI was clean and all reviews were applied as appropriate.

# PR 2295 — Reviewer's guide: In-process FFI runtime for the Java SDK

## TL;DR

This PR does for the Java SDK what [#1901](https://github.com/github/copilot-sdk/pull/1901) did for .NET and [#1915](https://github.com/github/copilot-sdk/pull/1915) did for Rust: it adds an **in-process connection mode** that loads the Copilot runtime (`runtime.node` cdylib) as a native library via JNA, eliminating the need for a separate CLI child process. Currently scoped to **linux-x64 only**; the entire in-process API surface is marked `@CopilotExperimental`.

The PR also restructures the Java Maven project from a single module into a **multi-module reactor** to support publishing the native runtime binaries as separate classifier JARs alongside the existing SDK JAR.

---

### The binary: `runtime.node`

Despite the `.node` extension (a napi-rs naming convention), `runtime.node` is an ordinary platform-specific shared library (`.so` on Linux). It is a Rust `cdylib` produced by the `src/runtime` crate in `github/copilot-agent-runtime`. It exposes two front doors:

- **napi front door** — loaded by Node.js as a native addon (existing CLI path).
- **C ABI front door** — 5 `extern "C"` lifecycle/transport entry points callable by any language via FFI without Node.js.

The 5 C ABI entry points are:

| Entry point | Purpose |
|---|---|
| `copilot_runtime_host_start` | Start the runtime host. Blocks up to ~30s while the worker boots. Returns a server handle (0 = failure). |
| `copilot_runtime_host_shutdown` | Shut down a runtime host by server handle. |
| `copilot_runtime_connection_open` | Open a bidirectional connection; registers an `on_outbound` callback for runtime→SDK data delivery. |
| `copilot_runtime_connection_write` | Write a JSON-RPC frame from the SDK into the runtime. |
| `copilot_runtime_connection_close` | Close a connection. |

All JSON-RPC methods travel as data through this fixed 5-function transport; the export surface never changes as the method set grows.

### Where it comes from (build-time)

The `copilot-native` Maven module's build fetches the binary from npm during `generate-resources`:

1. **`fetch-native.mjs`** reads the pinned version and SHA-512 integrity hash for `@github/copilot-linux-x64` from `nodejs/package-lock.json`.
2. Runs `npm pack` to download the exact tarball, verifies it against the integrity hash.
3. Extracts `runtime.node` and the `copilot` CLI executable into a staging directory.
4. `maven-jar-plugin` packages them into a **classifier JAR** (`copilot-sdk-java-runtime-<version>-linux-x64.jar`) with the layout `native/linux-x64/runtime.node`.

### How it's loaded (runtime)

1. **`PlatformDetector`** (303 lines) determines the classifier using `os.name`, `os.arch`, and on Linux, ELF PT_INTERP parsing to distinguish glibc vs musl — no subprocesses, no heuristics.
2. **`NativeRuntimeLoader`** (466 lines) resolves the binary in this order:
   - `COPILOT_CLI_PATH` env var → checks for `runtime.node` alongside the CLI.
   - Classpath resource `native/<classifier>/runtime.node` → extracts atomically to `~/.copilot/runtime-cache/<version>/<classifier>/runtime.node`.
   - Falls back to `runtime.node` alongside the bundled `copilot` executable.
3. **`JnaNativeBinding`** (253 lines) loads the library by absolute path via JNA and maps each C ABI export. Enforces a one-library-per-process invariant (library handle is `static`, never unloaded). Duplicate loads from the same path are silently accepted; different paths are rejected.
4. **`FfiRuntimeHost`** (349 lines) orchestrates the lifecycle: starts the host, opens a connection, bridges the bidirectional JSON-RPC transport. The `on_outbound` callback (invoked by native threads) feeds received data into a `QueueInputStream` that the SDK's existing `JsonRpcClient` reads from.

---

### Multi-module Maven reactor

The single-module `java/pom.xml` is now a parent POM (`pom` packaging) with two submodules:

| Module | Artifact ID | Purpose |
|---|---|---|
| `java/pom.xml` | `copilot-sdk-java-parent` | Reactor parent. Not published to Maven Central (`maven.deploy.skip=true`). Holds the `release` profile (GPG signing) inherited by all submodules. |
| `java/sdk/` | `copilot-sdk-java` | The existing SDK JAR (~1.5 MB). All existing source moved here from `java/src/` → `java/sdk/src/`. |
| `java/copilot-native/` | `copilot-sdk-java-runtime` | Native runtime module. Produces classifier JARs (currently `linux-x64` only, ~20-26 MB). |

### Consumer dependency declaration

```xml
<dependencies>
    <!-- Pure-Java SDK (~1.5 MB) -->
    <dependency>
        <groupId>com.github</groupId>
        <artifactId>copilot-sdk-java</artifactId>
        <version>${copilot.version}</version>
    </dependency>
    <!-- Native runtime for linux-x64 (~20-26 MB) — needed only for in-process mode -->
    <dependency>
        <groupId>com.github</groupId>
        <artifactId>copilot-sdk-java-runtime</artifactId>
        <version>${copilot.version}</version>
        <classifier>linux-x64</classifier>
    </dependency>
</dependencies>
```

### Consumer usage

```java
CopilotClientOptions options = new CopilotClientOptions()
    .setConnection(RuntimeConnection.forInProcess());

CopilotClient client = new CopilotClient(options);
client.start().get();
```

---

## New public API surface (all `@CopilotExperimental`)

| Type | Description |
|---|---|
| `RuntimeConnection` (sealed class) | Base type for connection configuration. Factory methods: `forStdio()`, `forTcp()`, `forUri(String)`, `forInProcess()`. |
| `StdioRuntimeConnection` | Spawns a runtime child process, communicates over stdin/stdout (the default). |
| `TcpRuntimeConnection` | Spawns a runtime child process listening on a TCP socket. |
| `UriRuntimeConnection` | Connects to an already-running runtime at a URL. |
| `InProcessRuntimeConnection` | Loads the native library in-process — no child process spawned. |
| `CopilotClientOptions.setConnection()` / `getConnection()` | Entry point for selecting a connection type. |

The `RuntimeConnection` API replaces the previous pattern of setting `cliUrl`, `cliPath`, `useStdio`, `port`, and `tcpConnectionToken` individually. When a `RuntimeConnection` is set, it takes precedence; conflicting legacy options cause `IllegalArgumentException`.

---

### `com.github.copilot.ffi` (9 classes, ~1,752 lines)

| Class | Lines | Role |
|---|---|---|
| `FfiRuntimeHost` | 349 | Lifecycle manager: start host → open connection → bridge I/O → shutdown. |
| `JnaNativeBinding` | 253 | JNA bindings for the 5 C ABI exports. Static library handle, one-per-process guard. |
| `NativeBinding` | 131 | Abstract contract for native operations (enables testing without real native library). |
| `NativeRuntimeLoader` | 466 | Locates `runtime.node`: env var → classpath → cache. Atomic extraction with file locking. |
| `PlatformDetector` | 303 | Determines platform classifier. ELF PT_INTERP parsing for glibc/musl detection on Linux. |
| `QueueInputStream` | 119 | Thread-safe bridge: native callback thread writes → SDK reader thread reads. |
| `FfiOutputStream` | 63 | Writes JSON-RPC frames from the SDK into the native runtime via `connection_write`. |
| `OutboundCallback` | 46 | JNA callback implementation for `on_outbound`. |
| `ReaderThreadFactory` | 22 | Named daemon thread factory for the reader executor. |

### Tests for FFI (6 files, ~2,054 lines)

| Test class | What it covers |
|---|---|
| `FfiRuntimeHostTest` | Lifecycle, error handling, concurrent shutdown, callback drain. |
| `JnaNativeBindingTest` | Load guard, duplicate-path acceptance, different-path rejection, active callback tracking. |
| `NativeRuntimeLoaderTest` | Resolution order, atomic extraction, COPILOT_CLI_PATH override, cache reuse. |
| `PlatformDetectorTest` | All 8 platform classifiers, ELF parsing, edge cases. |
| `QueueInputStreamTest` | Thread-safe read/write, close semantics. |
| `InProcessTransportIT` | End-to-end integration test using the replay proxy with in-process transport. |

---

## CI/workflow changes

- **New job** `java-sdk-inprocess` in `java-sdk-tests.yml`: runs `mvn clean verify -Pinprocess` on ubuntu-latest (linux-x64). Uses `continue-on-error: true` while experimental.
- **Path updates** in existing jobs: `java/target/` → `java/sdk/target/` for surefire/failsafe reports and coverage data.
- **JDK 17 cross-test**: added `-pl sdk` to restrict to the SDK module (the native module requires JDK 25 build tools).
- **Codegen workflows**: adjusted working directories for the `java/sdk/` module layout.

✅ Note that the existing java publishing jobs will continue to work as currently written.

---

## Key design decisions (from ADR-007)

1. **JNA over Panama FFM**: JNA supports the Java 17 baseline with zero consumer configuration. Panama FFM requires Java 22+ and `--enable-native-access` flags. Performance difference is irrelevant (JSON-RPC I/O dominates).

2. **Per-platform classifier JARs over monolithic JAR**: A monolithic JAR with all 6 common platforms would be ~132 MB. Classifier JARs let consumers pull only their target platform (~20-26 MB each). An uber-JAR can be assembled via `maven-assembly-plugin` if needed.

3. **Library-never-unloads pattern**: The loaded native library is held in a `static` field and never released. Native worker threads outlive any individual `FfiRuntimeHost` instance; unloading would crash.

4. **One library per process**: Enforced by a process-wide guard, consistent with Rust, .NET, Go, and Python SDK implementations.

---

## Diff statistics

- **107 commits**, 1,582 files changed (mostly renames from `java/src/` → `java/sdk/src/`)
- **~6,624 insertions**, ~825 deletions
- New production code: ~2,117 lines (FFI + RuntimeConnection API)
- New test code: ~2,054 lines
- New build infrastructure: `copilot-native/pom.xml` (214 lines), `fetch-native.mjs` (114 lines)

## Recommended review order

1. **ADR-007**: `java/docs/adr/adr-007-native-bundling-strategy.md` — context, options considered, decision rationale.
2. **RuntimeConnection API**: `rpc/RuntimeConnection.java`, `rpc/InProcessRuntimeConnection.java`, and `rpc/CopilotClientOptions.java` (the `setConnection`/`getConnection` methods).
3. **FFI bridge** (bottom-up): `NativeBinding.java` → `JnaNativeBinding.java` → `FfiRuntimeHost.java` → `NativeRuntimeLoader.java` → `PlatformDetector.java`.
4. **Native module build**: `copilot-native/pom.xml` and `copilot-native/scripts/fetch-native.mjs`.
5. **Multi-module restructure**: `java/pom.xml` (parent) and `java/sdk/pom.xml` (child).
6. **CI**: `.github/workflows/java-sdk-tests.yml` (new inprocess job, path updates).
7. **Tests**: `ffi/` test package and `e2e/InProcessTransportIT.java`.

## Implementation details.

Implemented agentically using https://aka.ms/coreai/shepherd-task/slides .

- Issues, including post-mortems: https://github.com/github/copilot-sdk/issues/1917 .


java/pom.xml
```diff
@@ -88,54 +63,29 @@
             DO NOT EDIT MANUALLY. Updated by the update-copilot-dependency
             workflow.
         -->
-        <readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>^1.0.79-9</readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>
-
+        <readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>^1.0.79-6</readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>
+        <!-- The parent POM is not published to Maven Central. -->
+        <maven.deploy.skip>true</maven.deploy.skip>
```
roji, line 68: **Blocker:** Both published child POMs inherit from `copilot-sdk-java-parent`, but this prevents that parent from being deployed and there is no flattening step. Maven consumers resolving `copilot-sdk-java` or `copilot-sdk-java-runtime` will then fail to resolve their parent POM. Please publish the parent or deploy flattened child POMs.

java/pom.xml
```diff
@@ -88,54 +63,29 @@
             DO NOT EDIT MANUALLY. Updated by the update-copilot-dependency
             workflow.
         -->
-        <readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>^1.0.79-9</readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>
-
+        <readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>^1.0.79-6</readonly-copilot-sdk-ref-impl-version-from-lastmerge-file-updated-by-reference-impl-sync>
```
roji: **High:** This pin is being downgraded from `^1.0.79-9` to `^1.0.79-6`, and the regenerated public API consequently loses current types and fields such as `SandboxConfigAuth`, factory-agent options, `SubagentCompletedEvent.cancelled`, and paged `listRuns`. This is unrelated to the FFI work and rolls `main` backwards. Please restore the current pin and regenerate.

java/README.md
```diff
@@ -70,6 +70,46 @@ Replace `${copilot.sdk.version}` with the latest release from Maven Central.
...
+    <!-- Pure-Java SDK (~1.5 MB) -->
+    <dependency>
+        <groupId>com.github</groupId>
+        <artifactId>copilot-sdk-java</artifactId>
+        <version>${copilot.version}</version>
+    </dependency>
+    <!-- Native runtime for linux-x64 (~20-26 MB) -->
+    <dependency>
+        <groupId>com.github</groupId>
+        <artifactId>copilot-sdk-java-runtime</artifactId>
```
roji, line 94: **Blocker:** These two artifacts are not sufficient for in-process mode. The SDK declares JNA as optional, and the runtime artifact does not depend on JNA, so Maven will not bring it transitively. Users following this snippet will hit `NoClassDefFoundError` when selecting the in-process transport. Please make the runtime artifact bring JNA transitively or document an explicit third dependency.

java/sdk/src/main/java/com/github/copilot/ffi/JnaNativeBinding.java
```diff
@@ -0,0 +1,253 @@
...
+    public boolean connectionWrite(int connectionId, byte[] data, int dataLen) {
+        return lib.copilot_runtime_connection_write(connectionId, data, dataLen) != 0;
+    }
+
+    @Override
+    public boolean connectionClose(int connectionId) {
+        try {
+            return lib.copilot_runtime_connection_close(connectionId) != 0;
+        } finally {
+            trackedCallbacks.remove(connectionId);
```
roji: **High:** Removing the wrapper in `finally` releases its only GC root even when native `connection_close` fails or throws. Native code may still retain and invoke that function pointer, which can crash the JVM after JNA collects the callback. Keep the wrapper rooted through callback draining and host shutdown, and only release it when native ownership has definitely ended.

java/copilot-native/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+                                </condition>
+                                <fail if="placeholder.jar.polluted" message="Placeholder primary JAR must not contain native binaries" />
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <!--
+                Required by Maven Central: sources and javadoc artifacts. This
+                module has no Java sources, so both produce empty archives.
```
roji: **High:** This claim does not match the release output. Building this module with `-Prelease` emits the main and sources JARs but logs `No Javadoc in project. Archive not created`; Maven Central requires a javadoc JAR for every non-POM artifact. Please attach an explicit placeholder javadoc JAR for this source-less module.

java/sdk/src/main/java/com/github/copilot/ffi/OutboundCallback.java
```diff
@@ -0,0 +1,46 @@
...
+     *            opaque cookie passed through unchanged from
+     *            {@code copilot_runtime_connection_open}; always
+     *            {@code Pointer.NULL} in this SDK
+     * @param data
+     *            pointer to the outbound byte buffer; valid only for the duration
+     *            of this invocation
+     * @param len
+     *            byte length of the buffer pointed to by {@code data}
+     */
+    void invoke(Pointer userData, Pointer data, int len);
```
roji: **Medium:** The C ABI declares this length as `size_t`, but Java `int` is always 32-bit and therefore mismatches the 64-bit ABI. The other SDKs use `c_size_t`, `size_t`, `nuint`, or `usize`. Please use a JNA `IntegerType` sized with `Native.SIZE_T_SIZE` here and for every native length parameter.

java/sdk/src/main/java/com/github/copilot/ffi/FfiRuntimeHost.java
```diff
@@ -0,0 +1,349 @@
...
+    FfiRuntimeHost(NativeBinding nativeBinding, String libraryPath) {
+        this(nativeBinding, libraryPath, new QueueInputStream());
+    }
+
+    FfiRuntimeHost(NativeBinding nativeBinding, String libraryPath, QueueInputStream receiveStream) {
+        this.nativeBinding = Objects.requireNonNull(nativeBinding, "nativeBinding must not be null");
+        this.receiveStream = Objects.requireNonNull(receiveStream, "receiveStream must not be null");
+        this.sendStream = new FfiOutputStream(this.nativeBinding, this.connectionId, this.closing, this.operationLock);
+        this.libraryPath = libraryPath;
+        Native.setCallbackExceptionHandler((Callback callback, Throwable throwable) -> LOG.log(Level.WARNING,
```
roji: **Medium:** This replaces JNA's process-wide callback exception handler for the entire hosting application and never restores it, changing the behavior of unrelated JNA callbacks. The local outbound callback already catches and logs failures, so please avoid this global mutation or preserve and restore the prior handler.

.github/workflows/java-sdk-tests.yml
```diff
@@ -17,6 +17,52 @@ permissions:
   contents: read
 
 jobs:
+  java-sdk-inprocess:
+    name: "Java SDK InProcess Tests"
+    if: github.event.repository.fork == false
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
roji: **Medium:** Making the only real in-process test job `continue-on-error` means FFI regressions can never block this PR or later merges. Since this change introduces the transport, this job should be required once it is added.

java/README.md
```diff
@@ -156,7 +196,7 @@ PermissionHandler handler = (request, invocation) -> {
 
 You can run the SDK without setting up a full Java project, by using [JBang](https://www.jbang.dev/).
 
-See the full source of [`jbang-example.java`](jbang-example.java) for a complete example with more features like session idle handling and usage info events.
+See the full source of [`jbang-example.java`](sdk/jbang-example.java) for a complete example with more features like session idle handling and usage info events.
```
roji, line 205: **Medium:** This updates the source link after the move, but the runnable JBang URL immediately below still points to `java/jbang-example.java`. The release workflow also retains that old path. Please update both remaining references to `java/sdk/jbang-example.java`.

java/README.md
```diff
@@ -262,7 +302,7 @@ Chain fluent modifiers to set tool options:
 - `.defer(ToolDefer)` — control deferred execution (`AUTO`, `NEVER`)
 - `.overridesBuiltInTool(boolean)` — shadow built-in tools
 
-For design context and decision rationale, see [ADR-006](docs/adr/adr-006-tool-definition-inline.md).
+For design context and decision rationale, see [ADR-006](sdk/docs/adr/adr-006-tool-definition-inline.md).
```
roji: **Medium:** The ADRs remain under `java/docs/adr/`, so `sdk/docs/adr/...` does not exist. This link should remain `docs/adr/adr-006-tool-definition-inline.md`; the ADR-004 link later in the README needs the same correction.

roji (dismissed):
@edburns here are some comments from my review agent, hope these make sense. Happy to take another human look afterwards!

---

Requesting changes. The overall transport architecture broadly aligns with the other SDKs, and the large file count is mostly explainable: 1,522 of 1,596 files are byte-identical moves into `java/sdk/`. The module split is reasonable, but the published dependency graph, native ABI/lifecycle, and release validation still have blocking issues.

GitHub cannot attach inline review comments to unchanged files, so these relocation omissions are called out here:

- `.github/workflows/java-publish-maven.yml:204,207` still references `java/jbang-example.java`, so release preparation will fail after the move to `java/sdk/jbang-example.java`.
- `scripts/docs-validation/validate.ts:388-394` searches the parent POM for artifact `copilot-sdk-java`; it now falls back to `1.0.0-SNAPSHOT` instead of validating the reactor's `1.0.11-preview.0-SNAPSHOT` artifact.
- `.github/actions/java-test-report/action.yml:7,11,15` still searches `java/target/**`; current CI logs report that no test reports were found even though results are under `java/sdk/target/**`.
- `.github/workflows/java-smoke-test.yml:66,139` still points to the pre-move prompt path.

Please address the inline findings and these unchanged-file omissions before merging.

.github/workflows/java-sdk-tests.yml
```diff
@@ -17,6 +17,52 @@ permissions:
   contents: read
 
 jobs:
+  java-sdk-inprocess:
```
roji, line 20: Does it make sense to integrate this in the regular job as a matrix (`[default, inprocess]`) rather than having a completely separate leg for it?

java/copilot-native/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+            nodejs/package-lock.json, which pins both the runtime version and
+            its SHA-512 integrity hash.
+        -->
+        <copilot.sdk.root>${project.basedir}/../..</copilot.sdk.root>
+        <!--
+            Phase 4 hard invariant: linux-x64 is the only supported platform.
+            Additional classifiers are added in a later phase, each with its own
+            fetch execution and maven-jar-plugin execution.
+        -->
+        <copilot.native.classifier>linux-x64</copilot.native.classifier>
```
roji, line 43: Does this mean we don't support any other platform (linux-arm64, Windows/Mac)? Of course totally fine if you want to do this incrementally in separate PRs, just pointing it out.

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
```diff
@@ -0,0 +1,466 @@
...
+        Path cliOverride = resolveFromCliPath(cliPathEnv);
+        if (cliOverride != null) {
+            return cliOverride;
+        }
+
+        ClassLoader loader = NativeRuntimeLoader.class.getClassLoader();
+        String classifier = PlatformDetector.detectClassifier();
+        String version = readVersion(loader);
+        Path cacheBase = defaultCacheBase();
+        return resolve(null, findRuntimeOnPath(), cacheBase, loader, classifier, version);
```
roji, line 122: Copilot review comment:

Could we avoid automatically falling back to an arbitrary `copilot` found on `PATH` for in-process mode? The embedded native ABI must stay compatible with the CLI/runtime pair, while a PATH installation can be any version and may produce difficult-to-diagnose ABI skew. The other SDKs use bundled/pinned assets or an explicit `COPILOT_CLI_PATH`; requiring one of those here would keep Java aligned and make runtime selection deterministic.

roji (commented):
Here are a few more comments.

Another thing I noticed is that while all other language SDKs automatically download the correct platform package with the native binary, the current approach in this PR requires users to manually take a dependency on e.g. the linux-x64 package, in addition to the platform-agnostic SDK package.

I don't know anything about how this kind of thing works with Java/Maven; is it impossible/not "the right way" to offer something that does this automatically (as all the other SDKs do)? Or maybe you're planning to look at that separately in a later PR (obviously completely fine too). Just raising the question.

java/pom.xml
edburns: Fixed in e4de619f .

java/pom.xml
edburns: Fixed in c943d2de .

java/sdk/src/main/java/com/github/copilot/ffi/JnaNativeBinding.java
edburns: Fixed in 9c7b1022 and a3f85c05

java/README.md
edburns: Fixed in e9ab4fef .

java/copilot-native/pom.xml
edburns: Fixed in 2f0a5325 .

java/sdk/src/main/java/com/github/copilot/ffi/OutboundCallback.java
edburns: Fixed in 04e39fbc .

java/sdk/src/main/java/com/github/copilot/ffi/FfiRuntimeHost.java
edburns: Fixed in 8bdc8d03 .

.github/workflows/java-sdk-tests.yml
edburns: Fixed in 3baa6735 .

java/README.md
edburns: Fixed in 2bd4abd2 .

java/README.md
edburns: Fixed in c125e9dd

java/sdk/src/main/java/com/github/copilot/ffi/JnaNativeBinding.java
```diff
@@ -0,0 +1,254 @@
...
+final class JnaNativeBinding implements NativeBinding {
+
+    private static final Logger LOG = Logger.getLogger(JnaNativeBinding.class.getName());
+
+    /**
+     * JNA inner interface mapping the five {@code copilot_runtime_*} C ABI exports.
+     */
+    interface CopilotRuntimeLibrary extends Library {
+        /** Corresponds to {@code copilot_runtime_host_start}. */
+        int copilot_runtime_host_start(byte[] argvJson, SizeT argvJsonLen, byte[] envJson, SizeT envJsonLen);
```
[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

java/sdk/src/main/java/com/github/copilot/ffi/JnaNativeBinding.java
```diff
@@ -0,0 +1,254 @@
...
+         * <p>
+         * Returns {@code byte} (not Java {@code boolean}) because the Rust ABI exports
+         * a one-byte {@code bool}. JNA maps Java {@code boolean} as a 32-bit C
+         * {@code int}, which would read three extra bytes.
+         */
+        byte copilot_runtime_host_shutdown(int serverId);
+
+        /** Corresponds to {@code copilot_runtime_connection_open}. */
+        int copilot_runtime_connection_open(int serverId, OutboundCallback callback, Pointer userData, byte[] extSource,
+                SizeT extSourceLen, byte[] extName, SizeT extNameLen, byte[] connToken, SizeT connTokenLen);
```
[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

java/sdk/src/main/java/com/github/copilot/ffi/JnaNativeBinding.java
```diff
@@ -0,0 +1,254 @@
...
+
+        /**
+         * Corresponds to {@code copilot_runtime_connection_write}.
+         *
+         * @see #copilot_runtime_host_shutdown for why this returns {@code byte}
+         */
+        byte copilot_runtime_connection_write(int connectionId, byte[] data, SizeT dataLen);
```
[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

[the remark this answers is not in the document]
edburns: **False positive.** These parameters are required — they are declared on a JNA `Library` interface whose methods map 1:1 to native C function signatures. JNA marshals every declared parameter to the native call via dynamic proxies; the parameters are never referenced in Java source because JNA handles marshalling reflectively. Removing any parameter would break the native ABI binding and cause crashes or data corruption at runtime.

.github/workflows/java-sdk-tests.yml
edburns: Will do this as a successor work item. See #2308.

java/copilot-native/pom.xml
edburns: Yes. I am pursuing a depth-first approach. I already mooted this approach in a previous triage meeting and @SteveSandersonMS approved it.

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
edburns: Thanks for flagging this — ABI skew is a real concern and worth scrutinizing.

Looking at the actual resolution chain, I believe the current design avoids the scenario you describe, but I want to walk through it to make sure I'm not missing something:

1. **`COPILOT_CLI_PATH` env var** — explicit user override; looks for `runtime.node` alongside the configured CLI. User takes responsibility for version alignment.
2. **Classpath resource** (`native/<classifier>/runtime.node`) — extracted from the classifier JAR, which is version-locked to the SDK at publish time. This is the primary path.
3. **PATH-discovered `copilot`** — reached only if both (1) and (2) miss. Crucially, this doesn't load an arbitrary `runtime.node` from somewhere else; it looks for `runtime.node` *in the same directory* as the `copilot` binary found on PATH. Since both ship together in the same `@github/copilot-<platform>` npm package, they're from the same build.

So the PATH fallback is "find the co-located native library next to a PATH-installed CLI," not "find any `runtime.node` anywhere." The version pairing is maintained by co-location.

That said — if you see a gap I'm not seeing, or if the concern is that co-location is too fragile a guarantee (e.g., someone could copy `copilot` without `runtime.node`, or vice versa), I'm happy to tighten this further. One option would be to fail rather than fall back when no classifier JAR is present and `COPILOT_CLI_PATH` isn't set, making the PATH lookup purely a dev convenience behind a flag.

edburns:
> Here are a few more comments.
> 
> Another thing I noticed is that while all other language SDKs automatically download the correct platform package with the native binary, the current approach in this PR requires users to manually take a dependency on e.g. the linux-x64 package, in addition to the platform-agnostic SDK package.
> 
> I don't know anything about how this kind of thing works with Java/Maven; is it impossible/not "the right way" to offer something that does this automatically (as all the other SDKs do)? Or maybe you're planning to look at that separately in a later PR (obviously completely fine too). Just raising the question.

We followed prior art:

> This is the same pattern used by DJL's PyTorch native artifacts (pytorch-native-cpu-2.5.1-linux-x86_64.jar, pytorch-native-cpu-2.5.1-osx-aarch64.jar, etc.), Netty's netty-tcnative-boringssl-static per-platform JARs, and others.

See the [ADR](https://github.com/github/copilot-sdk/blob/main/java/docs/adr/adr-007-native-bundling-strategy.md).

java/sdk/src/main/java/com/github/copilot/ffi/JnaNativeBinding.java
```diff
@@ -0,0 +1,254 @@
...
+    public boolean connectionWrite(int connectionId, byte[] data, int dataLen) {
+        return lib.copilot_runtime_connection_write(connectionId, data, new SizeT(dataLen)) != 0;
+    }
+
+    @Override
+    public boolean connectionClose(int connectionId) {
+        boolean closed = lib.copilot_runtime_connection_close(connectionId) != 0;
+        if (closed) {
+            // Native side has released ownership; safe to drop our GC root.
+            trackedCallbacks.remove(connectionId);
```
roji: This still drops the only strong reference to the JNA wrapper too early. Rust's outbound sink runs on a separate queue/thread, and the runtime's own C ABI tests explicitly allow callbacks after `connection_close` and server removal return. `FfiRuntimeHost.callbackRef` only roots the original callback, not this wrapper whose trampoline native code holds, and its separate callback counter can observe zero before another queued callback starts. That leaves a possible native call into a collected JNA callback, which can crash the JVM. Please keep this wrapper rooted for at least as long as native can possibly invoke it; with the current native contract that may need to be process lifetime unless shutdown is changed to join/drain all callback producers.

java/sdk/src/main/java/com/github/copilot/CopilotClient.java
```diff
@@ -131,7 +143,30 @@ public CopilotClient() {
...
+            validateEnvironmentOptions(this.options, requestedConnection);
+            validateConnectionConflicts(this.options, requestedConnection);
+            applyConnection(this.options, requestedConnection);
+        } else {
+            requestedConnection = resolveDefaultConnection(this.options);
+            validateEnvironmentOptions(this.options, requestedConnection);
+            // When the env var overrides inference (e.g. inprocess), validate that
+            // no legacy transport options conflict with the resolved connection.
+            if (requestedConnection != null) {
+                validateConnectionConflicts(this.options, requestedConnection);
```
roji, line 166: This validation also runs for connections synthesized from the legacy options, not only for explicit `setConnection()` or an actual environment override. That breaks previously accepted out-of-process configurations—for example default stdio with a nonzero unused port, or `cliUrl` with legacy CLI args/port values that were previously ignored. Please bypass conflict validation for the ordinary inferred compatibility path and retain it only for explicit connections or a genuine env-var override.

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
```diff
@@ -0,0 +1,466 @@
...
+            // CLI not on classpath — this is allowed for the COPILOT_CLI_PATH fallback
+            // path but will fail later in resolveEntrypoint() if InProcess is selected.
+            return;
+        }
+
+        Files.createDirectories(cacheDir);
+        Path temp = Files.createTempFile(cacheDir, "cli-tmp-", "");
+        try {
+            copyResourceToTemp(cliResource, cliResourcePath, temp);
+            publisher.publish(temp, cachedCli);
```
roji, line 382: Please make the temp CLI executable—and verify that it is executable—before atomically publishing it. As written, another thread/process can accept the nonempty final file between the move and the later chmod and then fail to launch it with `EACCES`. `File.setExecutable()` returning `false` is also currently ignored, so the cache can be considered valid even when it cannot run.

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
roji, line 122: The automatic PATH fallback still does not handle the normal installed layouts. Unlike explicit `COPILOT_CLI_PATH`, this passes only the executable's parent and later checks a flat sibling `runtime.node`. npm's `copilot` normally resolves to the loader package while `runtime.node` lives in the platform package's `prebuilds/<classifier>` directory, and Homebrew's standalone executable has no adjacent `runtime.node`. Please either resolve the actual platform-package layout here or remove/document the fallback rather than implying an arbitrary PATH installation enables in-process mode.

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
```diff
@@ -0,0 +1,466 @@
...
+     * @return path to the extracted {@code runtime.node} binary
+     * @throws IOException
+     *             if I/O or the atomic rename fails
+     * @throws IllegalStateException
+     *             if the classpath resource is missing or empty
+     */
+    static Path extractToCache(Path cacheBase, ClassLoader loader, String classifier, String version,
+            AtomicPublisher publisher) throws IOException {
+        String resourcePath = "native/" + classifier + "/" + RUNTIME_FILENAME;
+        Path cacheDir = cacheBase.resolve(version).resolve(classifier);
```
roji: Using only the Maven project version as the cache identity makes mutable `-SNAPSHOT` builds reuse old native artifacts indefinitely. Rebuilding the same SDK version with a newer bundled runtime still returns the previous cached pair. Please include a content digest or native package version in the cache identity/validation.

java/copilot-native/scripts/fetch-native.mjs
```diff
@@ -0,0 +1,114 @@
...
+// Idempotence: skip the download when the staged binary already matches.
+// The stamp stores version + integrity + binary digest to ensure a corrupted
+// binary or lockfile integrity change is detected.
+if (fs.existsSync(runtimePath) && fs.existsSync(stampPath)) {
+  const stampLines = fs.readFileSync(stampPath, 'utf8').trim().split('\n');
+  const stampVersion = stampLines[0] || '';
+  const stampIntegrity = stampLines[1] || '';
+  const stampBinaryDigest = stampLines[2] || '';
+  const currentBinaryDigest = `sha512-${createHash('sha512').update(fs.readFileSync(runtimePath)).digest('base64')}`;
+  if (stampVersion === version && stampIntegrity === integrity && stampBinaryDigest === currentBinaryDigest) {
```
roji: The incremental fast path validates only `runtime.node`; it can return success when the required `copilot` executable is missing or stale. Since `host_start` requires both artifacts from the same package, the stamp should include and validate the CLI digest as well before skipping extraction.

java/copilot-native/pom.xml
```diff
@@ -0,0 +1,239 @@
...
+                                        <zipentry zipfile="${project.build.directory}/${project.build.finalName}-${copilot.native.classifier}.jar" name="native/${copilot.native.classifier}/runtime.node" />
+                                    </resourceexists>
+                                </condition>
+                                <fail unless="native.binary.present" message="Classifier JAR is missing native/${copilot.native.classifier}/runtime.node" />
+                                <condition property="native.properties.present">
+                                    <resourceexists>
+                                        <zipentry zipfile="${project.build.directory}/${project.build.finalName}-${copilot.native.classifier}.jar" name="native/${copilot.native.classifier}/platform.properties" />
+                                    </resourceexists>
+                                </condition>
+                                <fail unless="native.properties.present" message="Classifier JAR is missing native/${copilot.native.classifier}/platform.properties" />
```
roji, line 193: The classifier-JAR guard should also assert that `native/${copilot.native.classifier}/copilot` (or `copilot.exe`) exists. The runtime cannot start without that executable, but a malformed release currently passes these structural checks as long as `runtime.node` and `platform.properties` are present.

java/pom.xml
```diff
@@ -36,50 +36,25 @@
...
     </scm>
 
-    <distributionManagement>
-        <snapshotRepository>
-            <id>central</id>
-            <url>https://central.sonatype.com/repository/maven-snapshots/</url>
-        </snapshotRepository>
-    </distributionManagement>
+    <modules>
+        <module>sdk</module>
```
roji, line 40: Moving the SDK into this module leaves several repository consumers on the old paths. Please update them as part of this move:

- `.github/actions/java-test-report/action.yml` still reads `java/target/**`, so CI silently misses test and JaCoCo reports under `java/sdk/target/**`.
- `.github/workflows/java-smoke-test.yml` still references `src/test/prompts/PROMPT-smoke-test.md` from `java/` instead of `sdk/src/test/prompts/...`.
- `scripts/docs-validation/validate.ts` still tries to extract the SDK artifact/version from `java/pom.xml` and falls back to `1.0.0-SNAPSHOT`.
- `.gitignore` ignores only `java/target`, so both new module targets remain untracked after a build.
- `.githooks/pre-commit` watches only `^java/src/`, so Spotless no longer runs for the moved sources.
- `.github/copilot-instructions.md` and the Java E2E skill still direct contributors to obsolete `java/src/**` and `java/target/**` paths.

java/sdk/pom.xml
```diff
@@ -0,0 +1,790 @@
...
+
+    <parent>
+        <groupId>com.github</groupId>
+        <artifactId>copilot-sdk-java-parent</artifactId>
+        <version>1.0.11-preview.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <groupId>com.github</groupId>
+    <artifactId>copilot-sdk-java</artifactId>
```
roji, line 16: Please override the inherited project URL and SCM metadata in each published child POM. Maven currently appends the child artifact ID to the inherited values, so the flattened artifacts publish malformed URLs such as `https://github.com/github/copilot-sdk/copilot-sdk-java` and `scm:git:https://github.com/github/copilot-sdk.git/copilot-sdk-java`.

java/sdk/pom.xml
```diff
@@ -0,0 +1,790 @@
...
+        <!-- JNA version — pin deliberately; a JNA upgrade must rerun the callback spike. -->
+        <jna.version>5.19.1</jna.version>
+    </properties>
+
+    <dependencies>
+        <!-- JSON-RPC -->
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.22.0</version>
```
roji: This restructuring also downgrades several unrelated dependencies/plugins: Jackson `2.22.1 -> 2.22.0`, SpotBugs annotations/plugin `4.10.3 -> 4.10.2`, Maven JAR plugin `3.5.1 -> 3.5.0`, and Central publishing plugin `0.11.0 -> 0.10.0`. Please restore the pre-PR versions or explain why each downgrade is required for the FFI/module change.

roji (approved):

java/sdk/src/main/java/com/github/copilot/CopilotClient.java
edburns: Fixed in a598f8b3bb6e691326f1b00961299f01d0bf1f9c.

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
edburns: Fixed in 13ad15a2e8619aaf3f805138106d98d8b43bfebf .

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
edburns: Fixed in 83a56a8ff46a681c8571aef59f86f713542b1222 .

java/sdk/src/main/java/com/github/copilot/ffi/NativeRuntimeLoader.java
edburns: Fixed in 6567869f5127ace65e972c529c492e9298877b57 .

java/copilot-native/scripts/fetch-native.mjs
edburns: Fixed in 1f0a24e18dc46a4468580ca0b3f5791993b09489 .

java/copilot-native/pom.xml
edburns: Fixed in ee14d45b537e5f87f8cc648f0efc8dc544f4e01a .

java/pom.xml
edburns: Fixed in ee14d45b537e5f87f8cc648f0efc8dc544f4e01a .

java/sdk/pom.xml
edburns: Fixed in ee14d45b537e5f87f8cc648f0efc8dc544f4e01a .

java/sdk/pom.xml
edburns: Fixed in ee14d45b537e5f87f8cc648f0efc8dc544f4e01a .

java/sdk/src/main/java/com/github/copilot/ffi/JnaNativeBinding.java
edburns: Fixed in a598f8b3bb6e691326f1b00961299f01d0bf1f9c .

PR #2310, opened 2026-08-11
fix(java): stop logging the legacy 'connect' probe failure as a warning

### LINKED ISSUE #2291: [java] Warning logged "Unhandled method connect" during the execution of the tutorial
I followed the java tutorial and it worked but gives ugly warning in the log what makes one think something is wrong.
As I can't find any connection to the client code itself it seems something inside the SDK itself the user can not influence.

```
Aug. 07, 2026 7:10:49 AM com.github.copilot.LoggingHelpers logTiming
WARNUNG: JsonRpc.invoke JSON-RPC request finished. Elapsed=PT0.793S, Method=connect, RequestId=1, Status=Failed
java.util.concurrent.CompletionException: com.github.copilot.JsonRpcException: Unhandled method connect
	at java.base/java.util.concurrent.CompletableFuture.wrapInCompletionException(CompletableFuture.java:323)
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:376)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:391)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:657)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:531)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2221)
	at com.github.copilot.JsonRpcClient.handleMessage(JsonRpcClient.java:282)
	at com.github.copilot.JsonRpcClient.lambda$startReader$0(JsonRpcClient.java:257)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
	at java.base/java.lang.Thread.run(Thread.java:1474)
Caused by: com.github.copilot.JsonRpcException: Unhandled method connect
	... 7 more
```

rinceyuan:
## Summary

`CopilotClient.verifyProtocolVersion` probes the `connect` RPC and falls back to `ping` when the server does not implement it. That fallback works, but `JsonRpcClient.invoke` logged *every* failed request at `WARNING` with a full stack trace, so the recovered probe printed this on a plain tutorial run (JUL sends `WARNING` to the console by default):

```
WARNUNG: JsonRpc.invoke JSON-RPC request finished. Elapsed=PT0.793S, Method=connect, RequestId=1, Status=Failed
java.util.concurrent.CompletionException: com.github.copilot.JsonRpcException: Unhandled method connect
```

## Change

- `JsonRpcClient.invoke` gets an internal (package-private) overload that takes the level used for failure logging; the existing 3-arg overload still defaults to `WARNING`.
- The protocol-negotiation `connect` probe passes `Level.FINE`, since `CopilotClient` already catches and recovers from that exact failure.

No public API change - `JsonRpcClient` is package-private. Genuinely unexpected RPC failures still log at `WARNING`.

## Tests

Two cases added to `JsonRpcClientTest`, both driving a real socket pair and a `-32601 Unhandled method connect` error response while capturing the JUL records:

- `testInvokeLogsFailureAtWarningByDefault` - unexpected failures are still `WARNING`.
- `testInvokeHonorsCustomFailureLogLevel` - an expected/recovered failure produces nothing at `WARNING` or above.

Verified `testInvokeHonorsCustomFailureLogLevel` fails when the level argument is ignored.

```
mvnw.cmd test -Dtest=JsonRpcClientTest -Denforcer.skip=true -Pskip-test-harness   # 25/25 pass
mvnw.cmd spotless:check -DspotlessFiles=... ; mvnw.cmd checkstyle:check           # clean
```

Fixes #2291.

### DIFF
--- a/java/sdk/src/main/java/com/github/copilot/CopilotClient.java
+++ b/java/sdk/src/main/java/com/github/copilot/CopilotClient.java
@@ -647,8 +647,12 @@ public final class CopilotClient implements AutoCloseable {
             if (this.options.getOnGitHubTelemetry() != null) {
                 connectParams.put("enableGitHubTelemetryForwarding", true);
             }
-            var connectResponse = connection.rpc.invoke("connect", connectParams, ConnectResult.class).get(30,
-                    TimeUnit.SECONDS);
+            // A legacy server rejects 'connect' and we fall back to 'ping' below, so
+            // only that rejection is expected; anything else stays a warning.
+            var connectResponse = connection.rpc
+                    .invoke("connect", connectParams, ConnectResult.class,
+                            cause -> cause instanceof JsonRpcException rpcEx && isUnsupportedConnectMethod(rpcEx))
+                    .get(30, TimeUnit.SECONDS);
             serverVersion = connectResponse.protocolVersion() != null
                     ? connectResponse.protocolVersion().intValue()
                     : null;
--- a/java/sdk/src/main/java/com/github/copilot/JsonRpcClient.java
+++ b/java/sdk/src/main/java/com/github/copilot/JsonRpcClient.java
@@ -19,6 +19,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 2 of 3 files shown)


rinceyuan:
Good catch - fixed.

The level is no longer decided up front. `JsonRpcClient.invoke` now takes a `Predicate<Throwable>` describing the failure the caller recovers from, evaluates it against the unwrapped cause inside `exceptionally`, and logs `FINE` only on a match. Everything else - `-32603`, deserialization failures, I/O failures, timeouts - still logs at `WARNING`.

The probe passes exactly the condition `verifyProtocolVersion` recovers from, so the two can no longer drift:

```java
cause -> cause instanceof JsonRpcException rpcEx && isUnsupportedConnectMethod(rpcEx)
```

Added `testInvokeKeepsWarningForUnexpectedFailure`, which drives a `-32603` through the same predicate and asserts a `WARNING` is still emitted. It fails against the previous revision.

`mvnw test -Dtest=JsonRpcClientTest` 26/26, spotless and checkstyle clean.

SteveSandersonMS:
Thanks for the thorough repro, tests, and cross-checking here — the code itself is correct, but on review this only fires when the CLI predates the SDK's expected protocol version, which isn't a supported configuration: we only support the SDK paired with its bundled CLI version. The warning is correctly telling you to update your CLI rather than something we should quiet down, so I'm closing this PR and the linked issue. I'll separately follow up on our compatibility docs, which currently describe this version-mismatch path more favorably than our current policy allows.

PR #2311, opened 2026-08-11
fix(rust): answer the request id when a handler panics

### LINKED ISSUE #2053: Rust SDK: follow-ups from #2034 — spawned request handlers can outlive stop_event_loop, and a panicking handler leaves the request unanswered
## Context

Follow-up items from the review of #2034 ([Fix `ask_user` starving the Rust SDK per-session event loop](https://github.com/github/copilot-sdk/pull/2034)).

That PR moved concurrency to the request-dispatch boundary: each inbound JSON-RPC request is now `tokio::spawn`ed from the `requests.recv()` branch of the per-session `tokio::select!` loop (matching the other five SDKs), rather than being awaited inline. This fixed the starvation but surfaced two lower-priority issues, called out by @stephentoub in [this review comment](https://github.com/github/copilot-sdk/pull/2034#issuecomment-5040423345). Both were accepted as out of scope for that PR (LGTM/approved) and are tracked here.

### 1. Spawned request handlers can outlive `stop_event_loop` — docs are stale
`Session::stop_event_loop` (in `rust/src/session.rs`) cancels the shutdown token and awaits the event-loop task handle. Its docs currently claim:

> Any in-flight handler (permission callback, tool call, elicitation response) completes before the loop exits, so the CLI never sees a half-handled request. See RFD-400 review finding #3.

Since request handlers are now detached `tokio::spawn`ed tasks, `stop_event_loop` only awaits the loop task itself — not the in-flight spawned request handlers. Those can outlive teardown, which may drop a response during shutdown. This **matches sibling SDK behavior and does not affect active sessions**, but the doc comment is now inaccurate.

**Action:** update the `stop_event_loop` (and any related `Drop`/`destroy`) documentation to reflect that spawned request handlers may not complete before the loop exits, and note the intentional parity with the other SDKs.

### 2. A panicking request handler leaves the request unanswered
Most sibling SDKs translate a panicking/throwing handler into a JSON-RPC error response. In the Rust SDK, a panic inside a spawned `handle_request` task simply aborts that task, leaving the single request unanswered (the caller hangs until its own timeout).

This is still an **improvement** over the pre-#2034 behavior, where a panic in an inline handler could take down the entire per-session event loop — now it's isolated to the one request. But for parity we should catch the panic and reply with a JSON-RPC error.

**Action:** wrap the spawned `handle_request` so a panic (or handler error) is converted into a JSON-RPC error response for that request id, matching the sibling SDKs.

## Notes
- Both are lower priority; neither affects active/healthy sessions.
- Reference implementation for the panic→JSON-RPC-error mapping exists in the dotnet/go/python/nodejs bindings.
- Related earlier flag from #2034: the other bindings' `userInput.request` handlers are also worth an audit for the original inline-await asymmetry.

rinceyuan:
## Summary

Follow-up items from #2034, tracked in #2053.

**1. A panicking request handler left the request unanswered.**
Since #2034 each inbound JSON-RPC request is dispatched to its own `tokio::spawn`ed task. A panic inside a handler is isolated to that task, so the request id is simply never answered and the caller waits out its own timeout. This catches the unwind at the dispatch boundary and replies with `-32603` for that id. The panic payload is not exposed; the method name is logged at `error!`.

**2. Stale lifecycle documentation.**
`Session::stop_event_loop` and the `Drop` impl still claimed that in-flight handlers complete before the loop exits. That has not been true since request dispatch moved to spawned tasks. The rustdoc now says handlers may outlive the loop and that their response can be lost if the connection closes first. The comment on `stop_event_loop_completes_in_flight_handler` claimed the loop was parked inside `handle_request`, which is likewise no longer how the test passes.

## Tests

New `panicking_request_handler_responds_with_internal_error` in `session_test`:

- drives a `userInput.request` into a handler that panics and asserts a `-32603` response for that id,
- then sends a second request and asserts the loop is still serving it, so the panic stays isolated.

Verified it times out (the reported symptom) with the source change reverted.

```
cargo +nightly-2026-04-14 fmt --check                        # clean
cargo clippy --all-features --all-targets -- -D warnings     # clean
cargo test --all-features --lib                              # 226 passed
cargo test --all-features --test session_test                # 118 passed
cargo test --all-features --test jsonrpc_test --test api_types_test --test protocol_version_test
```

The `e2e` target and `cli_resolution_test::stale_env_override_falls_through` need the bundled CLI, which cannot be downloaded from this machine (TLS interception on `registry.npmjs.org`), so they were run with `COPILOT_SKIP_CLI_DOWNLOAD=1` and fail identically on a clean checkout.

Fixes #2053.

### DIFF
--- a/rust/src/session.rs
+++ b/rust/src/session.rs
@@ -1,14 +1,16 @@
 use std::collections::HashMap;
+use std::panic::AssertUnwindSafe;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
+use futures_util::FutureExt;
 use parking_lot::Mutex as ParkingLotMutex;
 use serde_json::Value;
 use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
-use tracing::{Instrument, warn};
+use tracing::{Instrument, error, warn};
 
 use crate::canvas::CanvasHandler;
 use crate::generated::api_types::{
@@ -321,10 +323,15 @@ impl Session {
     /// Stop the internal event loop. Called automatically on [`destroy`](Self::destroy).
     ///
     /// Cooperative: signals shutdown via the session's [`CancellationToken`]
-    /// and awaits the loop's natural exit rather than aborting the task.
-    /// Any in-flight handler (permission callback, tool call, elicitation
-    /// response) completes before the loop exits, so the CLI never sees a
-    /// half-handled request. See RFD-400 review finding #3.
+    /// and awaits the loop's natural exit rather than aborting the task, so
+    /// the loop always stops between iterations instead of at an arbitrary
+    /// await point. See RFD-400 review finding #3.
+    ///
+    /// Inbound requests are dispatched to their own spawned tasks, which this
+    /// call does not await. A handler (permission callback, tool call,
+    /// elicitation response) still running at teardown may therefore outlive
+    /// the loop, and its response can be lost if the connection closes first.
+    /// Await your own handler work before calling this if it must complete.
     pub async fn stop_event_loop(&self) {
         self.shutdown.cancel();
         let handle = self.event_loop.lock().take();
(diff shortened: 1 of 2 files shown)


SteveSandersonMS (approved):
Verified locally: reverting only the source-side fix (keeping the new regression test) reproduces the reported hang — the request times out waiting for a response. With the fix applied, `cargo test --all-features --test session_test` passes 118/118 and `cargo fmt --check` is clean. The panic-recovery pattern here matches the existing `catch_unwind` usage already established in `router.rs` and `jsonrpc.rs`. Small, surgical, well-tested fix — approving.

PR #2328, opened 2026-08-13
Document persistent-file MCP OAuth storage in Rust

dy-hu:
## Why

The Copilot agent runtime is adding a `persistent-file` MCP OAuth token storage mode. Rust SDK consumers need to be able to select and understand this mode when creating or resuming sessions, including github/github-app#11730.

## What changed

- Document `persistent-file` on Rust create and resume session configuration APIs and builders.
- Explain its cross-process persistence, secure file-backed storage, and avoidance of steady-state OS Keychain access.
- Add focused serialization coverage proving create and resume wire requests preserve the value unchanged.
- Keep the existing `Option<String>` API shape so the change remains source compatible and matches the existing wire contract.

## Testing

- `cd rust && cargo +nightly-2026-04-14 fmt --check`
- `cd rust && cargo clippy --all-features --lib --tests -- -D warnings`
- `cd rust && cargo test --all-features session_configs_serialize_persistent_file_oauth_storage`

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1966,6 +1966,9 @@ pub struct SessionConfig {
     /// Controls how MCP OAuth tokens are stored for this session.
     ///
     /// - `"persistent"` — tokens are stored in the OS keychain (shared across sessions).
+    /// - `"persistent-file"` — tokens are stored in the secure file store, persist across
+    ///   processes, and avoid steady-state OS keychain access. The runtime imports legacy
+    ///   keychain credentials.
     /// - `"in-memory"` — tokens are stored in memory and discarded when the session ends.
     ///
     /// Defaults to `"in-memory"` when the client is in [`crate::ClientMode::Empty`],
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks for documenting the new `persistent-file` MCP OAuth token storage mode for Rust, and for adding the wire serialization test — that part looks good.

Since this is a runtime-wide capability, could you extend the same support/documentation to the other SDKs (Node, C#, Python, Go, and Java) so all 6 stay consistent? Specifically:

- **C# and Python**: these currently model the field as a closed type (`McpOAuthTokenStorageMode` enum in C#, `Literal["persistent", "in-memory"]` in Python), so `persistent-file` can't even be passed today — please widen those types to accept the new value.
- **Node, Go, Java**: these already accept arbitrary strings, so this is a docs-only update (mirroring what you did for Rust).

For test coverage, please stick to focused unit-level serialization/wiring tests per language (like the one you added for Rust) rather than new E2E tests — that keeps this change proportional to its scope.

Documenting/enabling this only in Rust leaves an avoidable capability and documentation gap for users of the other 5 SDKs. Once the above is in, please mark this ready for review again. Thanks!

PR #2336, opened 2026-08-14
fix(node): reject sendAndWait on session shutdown

connor4312:
## Summary

- treat `session.shutdown` as a terminal outcome in the Node SDK `sendAndWait()` helper
- reject even when shutdown arrives while the underlying `session.send` RPC is still pending
- add regression coverage for the lost-turn behavior reported in microsoft/vscode#330888

The runtime now emits `session.shutdown` when a session is torn down during an accepted turn, but `sendAndWait()` previously only observed `session.idle` and `session.error`, leaving callers waiting until timeout.

POC: [index.js](https://github.com/user-attachments/files/31084043/index.js)

## Validation

- `npm test -- session-send-and-wait.test.ts`
- `npm run typecheck`
- `npm run lint -- --quiet`
- `npm run format:check`
- deterministic resumed-session POC against the local SDK and current runtime main

### DIFF
--- a/nodejs/src/session.ts
+++ b/nodejs/src/session.ts
@@ -754,6 +754,17 @@ export class CopilotSession {
         const shutdownPromise = new Promise<Error>((resolve) => {
             resolveShutdown = resolve;
         });
+        let outcomeSettled = false;
+        const settleOutcome = (outcome: SessionOutcome, shutdown = false): void => {
+            if (outcomeSettled) {
+                return;
+            }
+            outcomeSettled = true;
+            resolveOutcome(outcome);
+            if (shutdown && outcome.kind === "error") {
+                resolveShutdown(outcome.error);
+            }
+        };
 
         let lastAssistantMessage: AssistantMessageEvent | undefined;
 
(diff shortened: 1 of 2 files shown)


rotationtv1-crypto:
Click link for telegram and come join the future with (Rotationtvnetwork LLC) Everything from AI automation, too Web3 tokenization technology, NFT'S. Rotationtvnetwork has AI NATIVE telegram mini apps for personal and enterprise grade companies. Our motto at ( Rotationtvnetwork) is we keep buiness Rotating globally 🌎 ✨ and we quantum precision and quality ratings. From building your Brand and buiness portfolio to building your company website, Rotationtvnetwork LLC is here too keep your Enterprise running even while your sleep. Click the link in Bio and Let's get you in Rotation ♾️ 🌍

SteveSandersonMS:
Thanks for tracking down this bug and fixing it for Node — nice catch on the shutdown-during-pending-send race too, and the regression test is a good addition.

Since `sendAndWait` exists in all six SDKs, we checked and confirmed the same gap exists in C#, Python, Go, Rust, and Java: none of them currently treat `session.shutdown` as a terminal outcome, so they'd hang the same way until timeout. Could you apply the equivalent fix to those five SDKs as well, with regression coverage in each following that language's existing test conventions (unit test where feasible, E2E where that's the norm for the SDK)?

We want to keep behavior aligned across all six SDKs — a Node-only fix here would leave an avoidable behavioral gap in the others. Once the other languages are updated, please mark this ready for review again. Thanks!

PR #2348, opened 2026-08-18
[Node] Let Extensions Request Sensitive Environment Variables

MRayermannMSFT:
## What

`joinSession()` accepts a new `env` option that names the sensitive environment variables an extension needs. The SDK sends those names to the CLI on the `session.resume` payload as `requestedEnvironmentVariables`. The CLI prompts the user with the extension name and the exact list. If the user approves, the SDK writes the granted values into the extension process's `process.env` before `joinSession()` resolves. If the user denies, `joinSession()` rejects and the extension does not load.

## Why

The CLI strips sensitive variables from every extension process before it starts, so an extension can never read one. The host half of the gate shipped in github/copilot-agent-runtime#15144, but the SDK builds its `session.resume` payload from an explicit field list, so an `env` option could not reach the wire. Without this change no extension can trigger the prompt, and the feature is unusable. The runtime PR left the changelog entry for this side, because the feature becomes usable here.

## A note on the E2E coverage

The released CLI predates the host half, so it ignores the request and grants nothing. The E2E therefore runs the real built SDK in a real forked extension process against a stub host that speaks the extension side of the wire, which is what the CLI itself sees. It proves the requested names reach the wire, that an approved grant lands in the extension's `process.env`, and that a denied request grants nothing. A second case drives the real CLI and proves that asking for variables does not break the join today. That case can assert the grant once the `@github/copilot` dependency carries the host half.


nodejs/src/extension.ts
```diff
@@ -107,6 +134,7 @@ export async function joinSession(config: JoinSessionConfig = {}): Promise<Copil
             onPermissionRequest: config.onPermissionRequest ?? defaultJoinSessionPermissionHandler,
             suppressResumeEvent: config.suppressResumeEvent ?? true,
         },
-        factories
+        factories,
+        env ? { requestedEnvironmentVariables: env } : undefined
```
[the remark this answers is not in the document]
MRayermannMSFT: Good catch, fixed in d4750e8a. `joinSession` now guards on `env?.length`, so an empty list is treated exactly like omitting the option and never reaches the wire. The JSDoc says so explicitly now.

On the coverage gap: you're right that the fixture omitted `env` for an empty parsed list, so the public path was untested. The fixture takes an `EXTENSION_ENV_REQUEST_EMPTY` flag now, and the "asks for nothing" E2E case runs both ways — option omitted and an explicit `env: []` — asserting neither puts `requestedEnvironmentVariables` on the resume payload. There's a unit case for `joinSession({ env: [] })` too.

> [!NOTE]

nodejs/src/client.ts
```diff
@@ -1884,8 +1887,26 @@ export class CopilotClient {
...
+            // The host answers an approved environment request with the resolved
+            // values, and this method consumes the response, so the grant has to be
+            // applied here — no caller ever sees it.
+            if (extensionOptions?.requestedEnvironmentVariables) {
+                const { grantedEnvironmentVariables } = response as {
+                    grantedEnvironmentVariables?: Record<string, string>;
+                };
+                for (const [name, value] of Object.entries(grantedEnvironmentVariables ?? {})) {
+                    process.env[name] = value;
+                }
```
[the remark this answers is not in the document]
MRayermannMSFT: Agreed, fixed in d4750e8a. The grant is intersected with the requested names before anything is written, so only names the user actually approved reach `process.env`.

A correct host never returns an extra entry, so this changes no behavior today — but the client is the last place the approved set is known, and it's cheap to enforce there rather than trust the answer. Covered by a unit test (a host returning one approved and one unrequested name) and an E2E case where the extension samples a name it never asked for and sees it empty after the join.

> [!NOTE]

nodejs/src/client.ts
```diff
@@ -1884,8 +1887,31 @@ export class CopilotClient {
...
+            // approved may reach the extension, so a host that answers with
+            // anything extra cannot widen the grant.
+            if (extensionOptions?.requestedEnvironmentVariables) {
+                const requested = new Set(extensionOptions.requestedEnvironmentVariables);
+                const { grantedEnvironmentVariables } = response as {
+                    grantedEnvironmentVariables?: Record<string, string>;
+                };
+                for (const [name, value] of Object.entries(grantedEnvironmentVariables ?? {})) {
+                    if (requested.has(name)) {
+                        process.env[name] = value;
```
MackinnonBuck, line 1910: Curious if you've considered returning the granted environment variables to the caller instead of automatically writing them into `process.env`? Since `session.resume` already returns `grantedEnvironmentVariables`, `joinSession()` could expose those values and let the extension decide whether to put them in its environment, pass them to a specific child process, or use them directly. That would avoid making the values ambiently available to all dependencies and inherited by child processes by default, although in fairness maybe that's the appropriate behavior if we expect that to be the 90%+ case.

nodejs/docs/extensions.md
```diff
@@ -53,6 +53,25 @@ const session = await joinSession({
...
 The `session` object provides methods for sending messages, logging to the timeline, listening to events, and accessing the RPC API. See the `.d.ts` files in the SDK package for full type information.
 
+## Requesting Sensitive Environment Variables
+
+The CLI strips sensitive environment variables (for example `GITHUB_TOKEN`) from every extension process before it starts. An extension that needs one asks for it by name:
+
+```js
+import { joinSession } from "@github/copilot-sdk/extension";
+
+const session = await joinSession({ env: ["GITHUB_TOKEN"] });
```
MackinnonBuck: Nit: `env` usually implies a map of values being supplied to a process, so it might be ambiguous for a list of names the extension is requesting access to. Should we use `requestedEnvironmentVariables` instead? It's a bit more explicit, albeit more verbose as well. Maybe the extra verbosity is worthwhile for a security-sensitive API. I don't have a strong preference here but just bringing it up as food for the thought.

MackinnonBuck (approved):
Looks great! Just some small thoughts that came to mind I was reviewing this. Nothing blocking, feel free to address these however you feel is appropriate!

nodejs/src/client.ts
MRayermannMSFT: Renamed to `requestedEnvironmentVariables` in 823aad4e. I think this is stronger than a nit, and I'd missed it.

The general convention point is right, but there's a concrete collision inside this SDK: `types.ts` already uses `env` three times — lines 219, 329, and 1756 — and every one is a `Record<string, string>` meaning "values to supply to a process". My `env: string[]` meant "names I'm asking for": same word, opposite direction of data flow, same file I was editing.

On the verbosity tradeoff — it's nearly free here, since an extension writes this once at startup. And the new name matches the wire field exactly, so it greps cleanly across this repo and the runtime when someone is staring at a `session.resume` payload.

The only cost is that the runtime's pending E2E fixtures currently call `joinSession({ env: [...] })`, but that scaffolding is written to be replaced when this lands anyway.

> [!NOTE]

nodejs/docs/extensions.md
MRayermannMSFT: (Reposting here — my previous reply about this landed on the `client.ts` thread by mistake.)

Renamed to `requestedEnvironmentVariables` in 823aad4e. I think this is stronger than a nit, and I'd missed it.

The general convention point is right, but there's a concrete collision inside this SDK: `types.ts` already uses `env` three times — lines 219, 329, and 1756 — and every one is a `Record<string, string>` meaning "values to supply to a process". My `env: string[]` meant "names I'm asking for": same word, opposite direction of data flow, same file I was editing.

On the verbosity tradeoff — it's nearly free here, since an extension writes this once at startup. And the new name matches the wire field exactly, so it greps cleanly across this repo and the runtime when someone is staring at a `session.resume` payload.

The only cost is that the runtime's pending E2E fixtures currently call `joinSession({ env: [...] })`, but that scaffolding is written to be replaced when this lands anyway.

> [!NOTE]

nodejs/src/client.ts
MRayermannMSFT: Apologies for the noise above — that reply was meant for the naming thread. Here's the actual answer to this one.

Honest answer: I didn't consider it. I took the shape from the runtime's own test scaffolding in `setupEnvAccessSdkDrop`, which writes to `process.env`. That explains why the shapes match; it isn't an argument that the shape is right. Your least-privilege point stands on its own.

Having thought it through, I'd still keep the `process.env` write as the default, for two reasons:

**It restores the state the extension would have had anyway.** The feature exists to undo a strip. If the name weren't filtered, it would be in `process.env`. Writing it back there makes the strip invisible to correct code, so `process.env.GITHUB_TOKEN` works whether the variable was blocked and then granted, or never blocked at all.

**Library convention.** Octokit, `gh`, and the Azure SDKs read `process.env` by convention. If the SDK only returns the values, every extension has to plumb them into each library by hand, and the common case gets worse to serve the uncommon one.

On the ambient-spread risk: the CLI prompt already tells the user that an approved extension can pass the value to anything it starts, so the consent model accounts for it. That doesn't make the risk zero, but it does mean the user isn't surprised by it.

The reason I'm comfortable deferring rather than deciding both halves now is the asymmetry. Exposing the granted values later — on the session, or as an opt-out flag — is additive and breaks nobody. Removing the `process.env` write later breaks every extension that relies on it. So the default is the only expensive thing to get wrong, and I believe this is the right default. If you'd like the values exposed as well, I'm happy to add it — say the word and I'll open a follow-up rather than widen this PR.

> [!NOTE]

PR #2360, opened 2026-08-19
fix(go): close failed session event loops

### LINKED ISSUE #2320: Go SDK: CreateSession failure leaks the pre-registered session's processEvents goroutine (non-cloud path)
Created with AI assistance - Human reviewed.

## Summary

On the **non-cloud session path**, `Client.CreateSession` leaks the session's `processEvents` goroutine whenever the `session.create` RPC (or its response handling) fails. The pre-registered session is removed from the client's session map on the error path **without** closing its event channel, so neither `Client.Stop()` nor a caller's own teardown can ever reap the goroutine. Each failed `CreateSession` call (and each retry) permanently leaks one goroutine plus its `Session` (which holds a 128-entry buffered channel) for the lifetime of the process.

Verified in **v1.0.6** and still present in **v1.0.9** (latest stable) — the relevant code is identical.

## Root cause (Go SDK, v1.0.9 line numbers)

1. `newSession` eagerly starts the consumer goroutine: `go s.processEvents()` (`go/session.go:389`). `processEvents` runs `for event := range s.eventCh` and only returns once `eventCh` is closed.
2. `eventCh` is closed **only** by `Session.Disconnect()` → `s.closeOnce.Do(func() { close(s.eventCh) })` (`go/session.go:1724`).
3. For non-cloud sessions, `CreateSession` generates a client-side `localSessionID` and **pre-registers** the session in `c.sessions` before issuing the RPC (`initializeSession` at `go/client.go:915`, invoked ~`go/client.go:990`) so that session-scoped requests emitted during `session.create` processing can be routed.
4. Every `CreateSession` error path after pre-registration then does only `delete(c.sessions, registeredSessionID)` and returns — it never calls `Disconnect()` / closes `eventCh`:
   - RPC failure: `go/client.go:1029-1034`
   - response unmarshal failure: `go/client.go:1039-1044`
   - sessionId mismatch: `go/client.go:1051-1055`
   - (the `SessionFS` validation paths inside `initializeSession`, `go/client.go:964` and `:972`, have the same shape)
5. `Client.Stop()` disconnects only sessions still present in `c.sessions`. Because step 4 already removed the session from the map, `Stop()` cannot see it, and the goroutine blocks forever on `eventCh`.

Equivalent locations in v1.0.6: `session.go:376`, `session.go:1662`, `client.go:920-946`, `client.go:436-445`.

## Impact

A long-running client that constructs a client/session per request accumulates one orphaned goroutine + `Session` per failed `CreateSession`. During a model/config or runtime outage — exactly when `session.create` fails and callers retry — goroutine and memory usage grow without bound until the process restarts. Successful sessions are unaffected (they stay in the map and are disconnected by `Stop()`).

The leaked resource is the in-process goroutine/`Session`; the CLI child process is still reaped by `Stop()`. So this is distinct from the child-process leaks in #1804 / #1381, and from the "disconnected sessions linger in the map" case that the (closed, unmerged) PR #1130 targeted.

## Reproduction

1. Build a `Client` over stdio (non-cloud).
2. Cause `session.create` to fail (e.g. point at an unreachable/invalid model or induce a runtime error) so `CreateSession` returns an error.
3. Call `Client.Stop()`.
4. Observe via `runtime.NumGoroutine()` / a goroutine dump that a `processEvents` goroutine remains parked on `chan receive (s.eventCh)`. Repeat N times → N parked goroutines.

## Suggested fix

On every `CreateSession` failure that occurs after the session has been pre-registered, close/disconnect the pre-registered session before returning (e.g. call `session.Disconnect()`, or at minimum `s.closeOnce.Do(func() { close(s.eventCh) })`) in addition to the existing `delete(c.sessions, …)`. A `defer`-based cleanup that runs on any non-nil error return would cover all of the paths above (including the `SessionFS` ones) uniformly. PR #1130's `OnDisposed` wiring is a related, more general approach but was closed unmerged.

## Environment

- SDK: `github.com/github/copilot-sdk/go` v1.0.6 (repro confirmed in source through v1.0.9)
- Language: Go
- Transport: stdio (non-cloud / local session)

green3sf:
## Summary

- stop the eagerly started session event consumer when `CreateSession` fails
- centralize local event-channel shutdown so failure cleanup does not issue a `session.destroy` RPC
- cover RPC errors, malformed responses, session ID mismatches, and SessionFS initialization failures

## Root cause

`newSession` starts `processEvents` immediately. The affected error paths removed the pre-registered session from `Client.sessions` but left its event channel open. Because the failed session was neither returned to the caller nor retained for `Client.Stop`, its goroutine remained blocked on the channel for the lifetime of the process.

The cleanup now removes only the session instance created by the failing call and closes its event channel without making another RPC.

## Impact

Failed non-cloud session creation no longer leaks one goroutine, session object, and buffered event channel per attempt.

## Validation

- `go test .`
- `go test -race . -run 'TestClient_CreateSession(Failure|InitializationFailure)ClosesRegisteredSession' -count=1`
- `git diff --check`

Fixes #2320

### DIFF
--- a/dotnet/src/Client.cs
+++ b/dotnet/src/Client.cs
@@ -1356,7 +1356,7 @@ public sealed partial class CopilotClient : IDisposable, IAsyncDisposable
         }
         catch (Exception ex)
         {
-            session?.RemoveFromClient();
+            session?.Unregister();
 
             if (ex is not OperationCanceledException)
             {
@@ -1561,7 +1561,7 @@ public sealed partial class CopilotClient : IDisposable, IAsyncDisposable
         }
         catch (Exception ex)
         {
-            session?.RemoveFromClient();
+            session?.Unregister();
             if (ex is not OperationCanceledException)
             {
                 LoggingHelpers.LogTiming(_logger, LogLevel.Warning, ex,
--- a/dotnet/src/Session.cs
+++ b/dotnet/src/Session.cs
@@ -204,6 +204,31 @@ public sealed partial class CopilotSession : IAsyncDisposable
         ((ICollection<KeyValuePair<string, CopilotSession>>)_parentClient._sessions).Remove(new(SessionId, this));
     }
 
+    /// <summary>
+    /// Stops the session's event consumer (<see cref="ProcessEventsAsync"/>) without
+    /// making an RPC. <see cref="CopilotClient.CreateSessionAsync"/> and
+    /// <see cref="CopilotClient.ResumeSessionAsync"/> use this on error paths where a
+    /// locally registered session fails before it can be returned to the caller:
+    /// <see cref="StartProcessingEvents"/> starts the consumer eagerly, and no caller
+    /// ever receives the failed session to dispose it. Safe to call more than once —
+    /// <see cref="ChannelWriter{T}.TryComplete"/> is idempotent.
+    /// </summary>
+    internal void CloseEventChannel()
+    {
+        _eventChannel.Writer.TryComplete();
+    }
+
+    /// <summary>
+    /// Removes the session from its parent client and stops its event consumer.
+    /// Used on session-creation/resume error paths where the session was registered
+    /// (and its event loop started) but never returned to the caller.
+    /// </summary>
+    internal void Unregister()
+    {
+        CloseEventChannel();
+        RemoveFromClient();
+    }
+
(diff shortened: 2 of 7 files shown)


qmuntal (approved):
Thanks!

PR #2366, opened 2026-08-20
Add reportedClientSessionId to Rust session create/resume wire

em-gong:
## What

Threads a new optional `reported_client_session_id` through the **Rust** SDK's `SessionConfig` and `ResumeSessionConfig`, serialized as a flat top-level `reportedClientSessionId` on both the `session.create` and `session.resume` JSON-RPC wire.

Part of a 4-repo fix for CCA (Cloud Copilot Coding Agent) per-session AI-credit attribution. On steering/resume runs the wrong session id lands on `X-Client-Session-Id`, so AI-credit spend bills to the task session instead of the steering session. This field lets callers seed the runtime's client-session scalar (→ `X-Client-Session-Id` + telemetry) **without** touching the logical/registry `session_id`, which keys resume-reattach, permission hooks, and the workspace filesystem.

## Semantics

- Wire key (camelCase JSON): `reportedClientSessionId` — **flat top-level, sibling of `sessionId`** (not nested), on both create and resume.
- Rust field: `reported_client_session_id: Option<String>`; builder: `with_reported_client_session_id(impl Into<String>)`.
- `#[serde(skip_serializing_if = "Option::is_none")]` ⇒ **absent/`None` emits nothing and the wire is byte-for-byte identical to today**; the engine then falls back to `session_id`. Non-empty wins.

Flat top-level is required because the runtime consumer (`copilot-agent-runtime#16469`) forwards this via its standalone-jsonrpc `UPDATE_OPTIONS_FORWARDED_KEYS` allowlist, which reads a flat top-level param on both `create_session` and `resume_session`. Nesting it under a sub-object would break pickup.

## Changes (Rust only)

- `SessionConfig` + `ResumeSessionConfig`: new `reported_client_session_id` field (docs; `Default`/`new` initialize to `None`; included in the custom `Debug` impls).
- `with_reported_client_session_id(..)` builder on each (mirrors `with_client_name`).
- `SessionCreateWire` + `SessionResumeWire`: new field, `skip_serializing_if = Option::is_none`, camelCase, flat top-level sibling of `sessionId`.
- Mapped through both `into_wire()`.
- Tests: absent ⇒ `reportedClientSessionId` omitted; present ⇒ appears top-level on **create and resume** (with `sessionId` unchanged).

## Scope: Rust-only (TS intentionally out of scope)

CCA drives copilotd via **copilotd bootstrap env vars** → copilot-host (Rust) → this Rust SDK builder → closed wire. It does **not** convey this identity through the TS `CopilotClient.createSession`/`resumeSession` JSON-RPC param path, so a TS `reportedClientSessionId` option would be permanent dead surface for this use case. Any non-CCA TS parity is a separate future PR.

## Consumer / release ordering

- Runtime consumer: `copilot-agent-runtime#16469`.
- **This SDK must release before copilot-host can bump its pin** and call `.with_reported_client_session_id(..)`.

## Validation

- `cargo test --all-features` — new tests pass.
- `cargo +nightly-2026-04-14 fmt --check` — clean.
- `cargo clippy --all-features --all-targets -- -D warnings` — clean.

### DIFF
--- a/rust/src/types.rs
+++ b/rust/src/types.rs
@@ -1909,6 +1909,14 @@ impl ManagedSettings {
 pub struct SessionConfig {
     /// Custom session ID. When unset, the CLI generates one.
     pub session_id: Option<SessionId>,
+    /// Client session ID reported for AI-credit attribution and telemetry.
+    ///
+    /// When set to a non-empty value, this seeds the runtime's client-session
+    /// scalar (which drives `X-Client-Session-Id` and telemetry) without
+    /// affecting the logical [`session_id`](Self::session_id) used for
+    /// resume-reattach, permission hooks, and the workspace filesystem. When
+    /// unset or empty, the runtime falls back to `session_id`.
+    pub reported_client_session_id: Option<String>,
     /// Model to use (e.g. `"gpt-4"`, `"claude-sonnet-4"`).
     pub model: Option<String>,
     /// Application name sent as `User-Agent` context.
@@ -2222,6 +2230,10 @@ impl std::fmt::Debug for SessionConfig {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("SessionConfig")
             .field("session_id", &self.session_id)
+            .field(
+                "reported_client_session_id",
+                &self.reported_client_session_id,
+            )
             .field("model", &self.model)
             .field("client_name", &self.client_name)
             .field("reasoning_effort", &self.reasoning_effort)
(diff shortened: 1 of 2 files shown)


em-gong:
Closing: superseded. The per-session AI-credit attribution fix is being solved via the capital proxy instead of threading a `reportedClientSessionId` through the CCA bootstrap chain (sweagentd ccav3 -> copilot-host -> copilot-sdk -> copilot-agent-runtime). That cross-repo approach is abandoned; no part of it should be revived. Closing unmerged.

Issue #2376, opened 2026-08-22
Add searchMessages() method to query inside session history


Navanit-git:
## Summary
Implement a searchMessages() method across all SDK languages to enable 
in-session message filtering and search.

## Problem
Users currently has to manually filter through all 
session events to find specific messages. This is cumbersome for:
- Debugging long conversations
- Finding specific agent responses  
- Building audit/search features on top of the SDK

## Proposed Solution
Add async searchMessages(query, options) method that:
- Accepts string or regex patterns
- Filters by event type (optional)
- Supports case-sensitive/insensitive search
- Returns filtered SessionEvent[] matching criteria

## Example Usage
```typescript
// TypeScript
const results = await session.searchMessages("authentication");
const assistantMsgs = await session.searchMessages("deploy", { 
  eventType: "assistant.message" 
});

Anvith433:
I'd like to work on this. 👋

I've opened github/copilot-sdk#2378 with a first implementation for the **Node.js/TypeScript** SDK — an async `searchMessages(query, options?)` on `CopilotSession` supporting `string`/`RegExp` queries, optional `eventType` filtering, and case-sensitivity, with unit tests.

I understand from CONTRIBUTING that feature work should be aligned with the team first and kept in sync across all supported languages, so I'm treating that PR as a concrete starting point for the discussion rather than a finished proposal. If you're open to this feature, I'm happy to extend it to the other SDKs (Python, Go, .NET, Java, Rust) and adjust the API shape/semantics to match your preferences.

Navanit-git:
I opened draft PR #2380 with an alternative cross-language implementation for Node.js, Python, Go, .NET, and Java. It intentionally overlaps with #2378, but proposes a narrower contract: search only persisted `user.message` and `assistant.message` content instead of recursively searching every string field in arbitrary event data.  Also **@maintainers** could you please advise which search semantics and cross-language API shape you prefer? 
I am happy to coordinate with @Anvith433  and revise or consolidate the work based on maintainer direction.

Anvith433:
Hey @Navanit-git, thanks for the heads up, and for pinging me instead of just
going ahead on your own.

Happy to fold my work into yours. Java's the language I actually know well, so
I'd like to take the Java part of #2380 - the implementation and the tests.

Honestly your version makes more sense than mine. In #2378 I searched every
string field, but just searching `user.message` and `assistant.message` is a
lot cleaner and easier to reason about. So no attachment to my approach, I'm
[...]

PR #2378, opened 2026-08-22
feat(nodejs): add Session.searchMessages() to query session history (#2376)

Anvith433:
## Summary

Implements the feature requested in #2376 — an async `searchMessages()` method on `CopilotSession` for querying inside a session's history, for the **Node.js/TypeScript** SDK.

```ts
async searchMessages(
    query: string | RegExp,
    options?: SearchMessagesOptions,
): Promise<SessionEvent[]>
```

- **`query`** — a `string` (case-insensitive substring by default; exact case via `{ caseSensitive: true }`) or a `RegExp` (case sensitivity governed by its own `i` flag).
- **`options.eventType`** — restrict results to a single event type (e.g. `"assistant.message"`). Typed as `SessionEventType | (string & {})` for autocomplete while still accepting any string.
- Returns the matching `SessionEvent[]` in history order.

## Design notes

- Fetches history via `getEvents()` and filters in memory.
- Matching runs only over the **string leaf values nested within `event.data`** (message text, tool output, error messages, nested attachment fields). Structural fields (`type`, `id`, `timestamp`) and object keys are intentionally not searched, so a query matches content rather than field names or UUIDs.
- For a `RegExp` query, the pattern is cloned with the stateful `g`/`y` flags stripped, so repeated `.test()` calls never depend on a shared `lastIndex` and the caller's `RegExp` is never mutated.
- An empty-string query matches every event (still subject to the `eventType` filter).

## Testing

- `npm test -- test/session-search-messages.test.ts` → 20/20 pass (string/RegExp matching, case sensitivity, `eventType` filtering, nested-`data` matching, `lastIndex` safety, caller's `RegExp` not mutated, order preservation, edge cases).
- `npm run typecheck` → clean. `npm run lint` → 0 errors.

## Scope / alignment

I read [CONTRIBUTING.md](CONTRIBUTING.md) and understand that (a) feature work should be discussed with the team first, and (b) features need to be maintained in sync across all supported languages. This PR is a **Node.js-first implementation** opened against the existing open enhancement issue #2376 as a concrete starting point for that discussion.

If the team is open to this feature, I'm happy to:
- extend it to the other SDKs (Python, Go, .NET, Java, Rust) for parity, and
- adjust the API shape or search semantics to match your preferences.

If it's not aligned with the roadmap, feel free to close — no worries at all, and thanks for the great SDK.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nodejs/src/index.ts
+++ b/nodejs/src/index.ts
@@ -135,6 +135,7 @@ export type {
     ProviderTokenArgs,
     RemoteSessionMode,
     ResumeSessionConfig,
+    SearchMessagesOptions,
     SectionOverride,
     SectionOverrideAction,
     SectionTransformFn,
--- a/nodejs/src/session.ts
+++ b/nodejs/src/session.ts
@@ -47,6 +47,7 @@ import type {
     ReasoningEffort,
     ReasoningSummary,
     ModelCapabilitiesOverride,
+    SearchMessagesOptions,
     SectionTransformFn,
     SessionCapabilities,
     SessionEvent,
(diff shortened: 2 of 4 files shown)


SteveSandersonMS:
Thanks for the contribution, but per our [CONTRIBUTING.md guidance](https://github.com/github/copilot-sdk/blob/main/CONTRIBUTING.md#before-you-submit-a-pr), feature work needs to be discussed and agreed with the maintainers before a PR is opened, and we haven't committed to adding `searchMessages()` yet, so I'm closing this for now — please continue the discussion on #2376 and we're happy to revisit if we get more community feedback that this is needed as an SDK feature.

PR #2380, opened 2026-08-23
feat: add cross-language session message search helpers

Navanit-git:
## Summary

- add client-side session message search helpers for Node.js, Python, Go, .NET, and Java
- search only persisted `user.message` and `assistant.message` content returned by the existing history APIs
- support event-type filtering, case sensitivity, and literal/regular-expression matching using idiomatic APIs per language
- add focused Node.js, Python, and Go coverage

## Coordination

Relates to #2376.

This draft intentionally overlaps with #2378. That PR currently provides a Node.js-first implementation that recursively searches string values across event data; this draft proposes a narrower conversational-message contract and demonstrates cross-language API parity. I am opening it separately so the alternative semantics and implementation work are visible, and I am happy to coordinate or adjust based on maintainer direction.

## Validation

- Python focused tests: `uv run pytest test_search_messages.py` (3 passed)
- Python Ruff checks: passed
- VS Code diagnostics for all touched files: clean
- Node.js tests: not run because dependency installation was blocked by local npm certificate/cache failures
- Go and .NET checks: not run because those toolchains are unavailable locally
- Java build: not run because the Maven wrapper download was blocked locally


Navanit-git:
@MackinnonBuck  Could you please review this and advise on the API direction, particularly the narrower conversational-message search semantics and cross-language shape? 
cc @stephentoub @devm33

SteveSandersonMS:
Thanks @Navanit-git but we didn't confirm that we planned to add any such feature, so the PR is a little premature. I'll close this for now. We could always reopen in the future if relevant.

It's very unclear to me that https://github.com/github/copilot-sdk/issues/2376 is a feature we should add since it's no better than simply getting all messages and writing a predicate to select from that list.

PR #2384, opened 2026-08-23
fix(python): make pydantic imports lazy

### LINKED ISSUE #2381: [Python] `pydantic` unconditionally imported in `tools.py`, breaking low-level Tool API users
I'm building an application on top of the Python SDK with GitHub Copilot's help, and I'd like the application to have no native dependencies.

For this reason, I instructed Copilot to favor the low-level Tool API. However, `pydantic-core` was still a required dependency, so I asked Copilot to investigate.

Copilot traced it to two module-level statements in [`copilot/tools.py`](https://github.com/github/copilot-sdk/blob/cc0438d66e3e68c333537cb935d9425d4e4ed8d5/python/copilot/tools.py):

```python
from pydantic import BaseModel, ValidationError   # line 16
T = TypeVar("T", bound=BaseModel)                 # line 92
```

According to Copilot's analysis, both are only needed by `define_tool`. The `Tool`, `ToolInvocation`, `ToolResult`, and `ToolBinaryResult` dataclasses have no pydantic dependency. Yet any application that uses only the low-level `Tool(...)` API must still have `pydantic-core` installed.

This blocks the SDK from being used in environments where native extensions cannot be loaded or where a platform-specific wheel is unavailable.

### Reproduction

```python
import sys
sys.modules['pydantic'] = None  # simulate missing pydantic
from copilot.tools import Tool  # ModuleNotFoundError
```

### Expected Behavior

`from copilot.tools import Tool` should succeed without pydantic installed, as long as `define_tool` is never called.

### Copilot Suggested Fix

Make both references lazy:

```python
def define_tool(name=None, *, description=None, ...):
    from pydantic import BaseModel, ValidationError
    T = TypeVar("T", bound=BaseModel)
    ...

def _is_pydantic_model(t):
    try:
        return isinstance(t, type) and issubclass(t, BaseModel)
    except ImportError:
        return False

def _normalize_result(result):
    ...
    # only when serializing a pydantic model
    try:
        if isinstance(result, BaseModel):
            ...
    except ImportError:
        pass
```

The change should be backward-compatible, as existing `define_tool` callers are unaffected.

green3sf:
Fixes #2381

## Summary

- move the Pydantic model bound behind `TYPE_CHECKING`
- import `ValidationError` and `BaseModel` only when tool type detection, typed validation, or Pydantic result serialization actually needs them
- add a subprocess regression test that blocks all `pydantic` imports and constructs the low-level `Tool` dataclass

This leaves `pydantic` in the package's declared dependencies, so existing `define_tool` installations keep their current behavior. The change only removes the module-level runtime dependency and lets controlled environments use the low-level API when Pydantic is unavailable.

## Validation

- `python -m pytest -q test_tools.py --tb=short` — 42 passed
- `ruff==0.16.0 check .` — passed
- `ruff==0.16.0 format --check .` — 115 files already formatted
- `ty check --python <existing-venv> copilot` — passed
- `git diff --check` — passed

The wider unit suite requires the repository's Copilot CLI binary at collection/runtime. I did not install the Node dependencies locally; the changed tool test module and full static checks are green.

### DIFF
--- a/python/copilot/tools.py
+++ b/python/copilot/tools.py
@@ -13,10 +13,14 @@ from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, get_type_hints, overload
 
-from pydantic import BaseModel, ValidationError
-
 if TYPE_CHECKING:
+    from pydantic import BaseModel
+
     from .generated.rpc import CurrentToolMetadata
+
+    T = TypeVar("T", bound=BaseModel)
+else:
+    T = TypeVar("T")
 
 from .generated.rpc import (
     ExternalToolTextResultForLlm,
@@ -89,7 +93,6 @@ class Tool:
     is_terminal: bool = False
 
 
-T = TypeVar("T", bound=BaseModel)
 R = TypeVar("R")
 
 
@@ -144,7 +147,7 @@ def define_tool(
     *,
     description: str | None = None,
     handler: Callable[[Any, ToolInvocation], Any] | None = None,
-    params_type: type[BaseModel] | None = None,
+    params_type: type[Any] | None = None,
     overrides_built_in_tool: bool = False,
     skip_permission: bool = False,
     defer: Literal["auto", "never"] | None = None,
@@ -251,6 +254,8 @@ def define_tool(
                 if takes_params:
                     args = invocation.arguments or {}
                     if ptype is not None and _is_pydantic_model(ptype):
+                        from pydantic import ValidationError
+
                         try:
                             call_args.append(ptype.model_validate(args))
                         except ValidationError as exc:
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks a lot for this fix, @green3sf — the analysis and diff are clean, and we appreciate the effort (including the subprocess regression test).

That said, we've decided to close this without merging. As discussed on the linked issue (github/copilot-sdk#2381 (comment)), there's an easy workaround available today for this exact scenario (stubbing `pydantic` in `sys.modules` before importing `copilot`), and so far only a single user has reported needing pydantic-free imports of the low-level `Tool` API, with no other corroborating reports or reactions. Given that, we'd rather not add the lazy-import complexity to `tools.py` for a niche case with a known workaround.

If we see more evidence that this affects a broader set of users, we'll gladly revisit — and this PR is a good reference for what that fix would look like. Thanks again for the contribution!

green3sf:
Thank you for the thoughtful explanation and review. That makes sense given the current demand and available workaround. I appreciate the context, and I’m glad this PR can serve as a reference if the issue becomes more widely reported.

PR #2386, opened 2026-08-24
SDKs: expose MCP server instruction policy

### LINKED ISSUE #2379: Expose and serialize allowAllMcpServerInstructions in TypeScript SDK
## Problem

The Copilot runtime protocol supports `allowAllMcpServerInstructions`, but TypeScript SDK 1.0.11 does not provide a working `SessionConfig` path for it.

The generated RPC types contain the field, and the runtime describes it as including instructions from every MCP server instead of only allowlisted servers. However:

- `dist/types.d.ts` does not expose it on `SessionConfigBase`.
- `CopilotClient.createSession()` in `dist/client.js` does not serialize it into the `session.create` request.
- A consumer adding the property at runtime therefore has no effect.

This prevents evaluation harnesses and other SDK consumers from explicitly testing or enabling instructions from a non-allowlisted MCP server.

## Controlled reproduction

A one-turn probe used a temporary MCP server whose unique canary existed only in server-level initialization instructions. The tool description did not contain the canary, and the prompt prohibited tool calls.

- Stock SDK: agent returned `NO_MCP_INSTRUCTION_CANARY` (18,924 tokens, 1 turn, 0 tool calls).
- Consumer supplied `allowAllMcpServerInstructions: true` without changing SDK serialization: same negative result.
- Temporary SDK wiring that serialized the field into `session.create`, together with the consumer option: agent returned the exact instruction canary (19,028 tokens, 1 turn, 0 tool calls).

This isolates the missing SDK forwarding from model variance and MCP tool selection.

## Requested behavior

1. Add `allowAllMcpServerInstructions?: boolean` to the public TypeScript session configuration shared by create/resume paths as appropriate.
2. Serialize it to the runtime request and ensure resumed or updated sessions honor the same policy.
3. Preserve the secure default: omitted or `false` must continue to exclude instructions from non-allowlisted servers.

## Required side-by-side tests

Please cover both policy states using the same non-allowlisted MCP server fixture:

- **Disallowed:** omitted and explicit `false` do not expose the server instruction canary.
- **Allowed:** explicit `true` exposes the exact canary.
- The server's tools remain available in both states, proving the assertion concerns instruction visibility rather than MCP startup.
- Wire-level tests assert both `false` and `true` are serialized correctly where supplied.

The allow/disallow behavior should remain explicit rather than changing the default globally.

OllieinCanada:
Fixes #2379

## Summary

- expose the optional MCP server instruction policy in the Node.js, Python, Go, Rust, .NET, and Java session APIs
- forward explicit `true` and `false` values through both `session.create` and `session.resume`
- omit the field when it is not configured, preserving the runtime's allowlist-based default
- update the branch to current `main`

The generated runtime protocol already supports `allowAllMcpServerInstructions`. This fills the missing public create/resume path consistently across every SDK without changing the generated protocol schema.

## Security

Enabling this option broadens the session's instruction trust boundary. Each public API documents that caveat, and no SDK enables it by default or coerces an omitted value.

## Tests

Deterministic wire/serialization tests cover explicit `true`, explicit `false`, and omission for create and resume in every SDK.

Local validation:

- Node.js focused Vitest regression: 3 passed; typecheck passed
- Python focused client regression: passed
- Go focused request regression: passed
- Java focused create/resume regressions: 2 passed; JDK 25 compilation and Checkstyle passed
- .NET public create/resume wire regressions: 3 passed on .NET 8 with SDK 10
- `git diff --check` for the PR delta: passed

Rust is covered by the repository's fresh CI matrix; its local toolchain was unavailable in this workspace.


SteveSandersonMS:
Thanks for the Node support here — the fix is clean, well-tested, and matches the existing `mcpServers` pattern in `client.ts`.

Since `allowAllMcpServerInstructions` already exists in the generated RPC layer for all 6 SDKs (Go, Python, Rust, .NET, Java, Node), exposing it only in Node leaves the same gap open for the other five languages — anyone setting it there today has it silently dropped, same as the bug this PR fixes for Node. Could you extend this PR to expose and forward the field consistently in C#, Python, Go, Rust, and Java as well, following the same create/resume wiring approach?

For each language, focused unit/wire-level tests covering explicit `true`, explicit `false`, and omitted (mirroring the 3 Node tests here) would be great — full E2E MCP tests aren't needed beyond that unless a particular language's test setup makes it uniquely useful.

Once all languages are covered, please mark the PR ready for review. Thanks again!

OllieinCanada:
Implemented the requested cross-SDK expansion in `0b864461` and updated the branch to current `main`.

Node.js, Python, Go, Rust, .NET, and Java now all expose and forward the option for create and resume. Each language has wire/serialization coverage for explicit `true`, explicit `false`, and omission. Local focused checks passed for Node.js, Python, Go, and Java (including JDK 25 compilation and Checkstyle); fresh CI will validate the Rust and .NET toolchains. I have marked the PR ready for review.

OllieinCanada:
Follow-up `ed361202` strengthens the .NET coverage by exercising the public create/resume APIs through the repository fake server. All three true/false/omitted cases pass locally on .NET 8 with SDK 10.

emmanuelik0147-rgb:
Th

PR #2394, opened 2026-08-25
Support experimental model picker categories

andysalerno:
## Summary
- recognize the runtime's experimental model picker category in generated Rust types
- keep the provisional supplemental-model RPC untyped through Client::call
- verify raw response decoding and inspectable method-not-found fallback

## Compatibility
This adds no first-class supplemental RPC API. Older runtimes continue returning JSON-RPC -32601, and unknown future categories still decode as Unknown.

## Testing
- cargo test --manifest-path rust/Cargo.toml --features test-support --test session_test raw_client_call_supports_supplemental_model_compatibility

### DIFF
--- a/rust/src/generated/api_types.rs
+++ b/rust/src/generated/api_types.rs
@@ -30406,6 +30406,8 @@ pub enum ModelPickerCategory {
     /// Powerful model category optimized for complex tasks.
     #[serde(rename = "powerful")]
     Powerful,
+    #[serde(rename = "experimental")]
+    Experimental,
     /// Unknown variant for forward compatibility.
     #[default]
     #[serde(other)]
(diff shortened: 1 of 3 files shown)


andysalerno:
Companion implementation: github/copilot-agent-runtime#16936 and github/github-app#13436. This PR intentionally adds category compatibility only; the supplemental RPC remains private and is invoked through generic Client::call.

SteveSandersonMS:
Thanks for tackling the Rust side of this and for the clear test coverage on the raw `Client::call` compatibility path!

Since `ModelPickerCategory` is a shared protocol enum, we'd like to keep all generated SDKs aligned rather than patching this into Rust codegen alone. Could you extend the same schema change (and regeneration) to Node, C#, Python, Go, and Java as well, with per-language decoding/behavior tests for the new `experimental` value and the unknown-category fallback? E2E coverage isn't necessary here unless existing conventions in a given language already call for it — unit/integration-level tests matching what you've done for Rust should be sufficient.

Marking this as draft and adding `waiting on response` in the meantime — please mark it ready for review once the other languages are updated. Thanks!

PR #2397, opened 2026-08-25
Fix .NET in-process callback lifetime

minyitang:
## Summary

- open tracked runtime connections and drain them before releasing managed callback state
- retain callback state and surface an error if graceful draining fails
- preserve `ForceStopAsync` behavior with an immediate close that safely retains callback state when necessary
- root the netstandard callback host with a `GCHandle`, matching the modern .NET lifetime guarantee

## Compatibility

This is stacked on #2390 because `copilot_runtime_connection_open_tracked` and `copilot_runtime_connection_close_and_wait` are available in the bundled runtime starting with CLI `1.0.81-7`; #2390 updates the SDK to `1.0.81-9`.

## Validation

- built `net8.0`, `net10.0`, and `netstandard2.0` with zero warnings
- 217 .NET unit tests passed
- graceful and force-stop in-process FFI E2Es passed
- `dotnet format --verify-no-changes` passed

### DIFF
--- a/dotnet/src/Client.cs
+++ b/dotnet/src/Client.cs
@@ -663,7 +663,17 @@ public sealed partial class CopilotClient : IDisposable, IAsyncDisposable
 
         if (ctx.FfiHost is { } ffiHost)
         {
-            try { ffiHost.Dispose(); }
+            try
+            {
+                if (gracefulRuntimeShutdown)
+                {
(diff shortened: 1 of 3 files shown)


roji:
Superceded by https://github.com/github/copilot-sdk/pull/2622

PR #2437, opened 2026-08-28
Add Auto routing tier options across all SDKs

dereklegenzoff:
## Summary

Runtime PR https://github.com/github/copilot-agent-runtime/pull/16072 introduced `capi.autoTier`, but the SDKs' handwritten session options did not expose it. Python also discarded the new field during wire conversion.

- Add typed `efficiency`, `balance`, and `intelligence` options across Node.js, Python, Go, .NET, Java, and Rust, including public exports and fluent builders where applicable.
- Forward Python's `auto_tier` as `autoTier`; preserve omission and compatibility with the existing WebSocket option on create/resume.
- Use the generated AutoTier and lifecycle-event contracts already present on main. Retain main's CLI `1.0.83-0` dependencies and lockfiles; no runtime downgrade or additional dependency bump is needed.
- Reuse shared generated AutoTier definitions where applicable and cover typed lifecycle-event decoding. Regenerate bindings after merging main to avoid duplicate Go aliases.
- Document runtime-owned persistence: cold resume restores or explicitly overrides the tier, while a conflicting tier on a resident session is rejected. The SDK does not choose a default or implement routing state.

## Verification

- After merging main, targeted CAPI forwarding and lifecycle-event tests passed in all six languages. Node/Python source type checks and neighboring factory/ask-user API coverage also passed.
- Previously repeated real-process SDK verification against published CLI `1.0.82-1` using Node and Python with an offline loopback CAPI fixture: all 14 model turns had the expected `/auto` tier payload or omission, with zero legacy routing calls.
- Confirmed typed session start/resume event values in both SDKs, resident-session same/omitted/conflicting tiers, cold restoration, explicit overrides, and override durability.
- Earlier verification against a freshly built runtime containing the upstream change also passed 47 assertions across 15 model turns. Neither loopback run exercised production CAPI.

## Compatibility

The feature requires Copilot CLI `1.0.82-1` or later with V2 Auto routing. The branch now uses main's newer `1.0.83-0`; no follow-up runtime bump is pending. Omitting the option preserves existing behavior.

The generated .NET AutoTier uses the SDK's forward-compatible string-struct pattern, and Rust uses its generated enum with an Unknown variant. Existing canonical member names are preserved; runtime validation remains authoritative.


SteveSandersonMS (approved):
Reviewed the PR end-to-end and applied the required Java README sample fix before merge. The feature exposes an existing runtime capability consistently across all SDK languages, preserves omission semantics, and was manually validated against the bundled CLI.

PR #2453, opened 2026-09-01
[go] Avoid duplicate bundler license downloads

qmuntal:
## Summary

- Download the umbrella `@github/copilot` license once, after all requested platform bundles are built.
- Avoid the redundant license request and unused license artifact when Linux glibc and musl bundles are produced together.
- Keep the existing npm tarball URL construction unchanged across SDKs.

## Validation

- `go test ./cmd/bundler -count=1`
- `go test ./internal/embeddedcli -count=1`
- `go vet ./cmd/bundler`

### DIFF
--- a/go/cmd/bundler/main.go
+++ b/go/cmd/bundler/main.go
@@ -115,6 +115,10 @@ func main() {
 			os.Exit(1)
 		}
 	}
+	if err := downloadCLILicense(version, outputPath); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: failed to download CLI license: %v\n", err)
+		os.Exit(1)
+	}
 
 	// Generate the Go file with embed directive
 	if err := generateGoFile(
@@ -321,9 +325,6 @@ func buildBundle(info platformInfo, cliVersion, outputPath, goos string) (bundle
 		if err != nil {
 			return bundleArtifacts{}, fmt.Errorf("failed to hash existing output: %w", err)
 		}
-		if err := downloadCLILicense(cliVersion, outputPath); err != nil {
-			return bundleArtifacts{}, fmt.Errorf("failed to download CLI license: %w", err)
-		}
 		runtimeHash, err := sha256FileFromCompressed(runtimeArtifactPath)
 		if err != nil {
 			return bundleArtifacts{}, fmt.Errorf("failed to hash existing runtime.node: %w", err)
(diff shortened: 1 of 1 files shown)


go/cmd/bundler/main_test.go
```diff
@@ -101,6 +105,39 @@ func readTarGz(t *testing.T, path string) map[string]string {
...
+	if err != nil {
+		t.Fatalf("fetchNPMTarballURL() error = %v", err)
+	}
+	if got != wantTarball {
+		t.Fatalf("fetchNPMTarballURL() = %q, want %q", got, wantTarball)
+	}
+}
+
+func TestFetchNPMTarballURLRejectsMissingTarball(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
```
emmanuelik0147-rgb: Yes

SteveSandersonMS:
Thanks for this. The unit tests are solid and the license-download dedup is a real win. 

On the metadata-resolution part: this changes only Go to decouple from the hardcoded npm tarball URL pattern, while .NET, Python, and Rust keep using that same pattern. If the URL convention is at genuine risk of changing, it needs to be fixed everywhere; if not, the metadata resolution adds complexity without benefit.

## Cross-language consistency is the blocker

The exact same hardcoded URL-construction pattern is deliberately mirrored in three other language SDKs:
- [`dotnet/src/build/GitHub.Copilot.SDK.targets`](https://github.com/github/copilot-sdk/blob/main/dotnet/src/build/GitHub.Copilot.SDK.targets#L63)
- [`python/copilot/_cli_version.py`](https://github.com/github/copilot-sdk/blob/main/python/copilot/_cli_version.py#L162) (docstring explicitly notes it "Mirrors the .NET targets' URL layout")
- `rust/build/in_process.rs` (~line 84)

Fixing only Go creates inconsistency: if npm's convention ever changes, only Go survives; the other three break silently. This directly opposes the repo's stated goal of consistent patterns across all 6 language SDKs.

## Proposed changes

- Either drop the metadata-resolution part (keep only the license-dedup optimization)
- Or apply the same fix consistently across dotnet, python, and rust, and document the motivation

## Suggestion

Split the license-dedup optimization into its own focused PR so it can merge independently.

qmuntal:
Thanks, agreed. I checked the published npm metadata before deciding: all 812 `@github/copilot` versions and 2,413 version records across the eight platform packages use the URL convention currently constructed by the SDKs. There are ancient exceptions in unrelated packages (for example, old Express prereleases), so `dist.tarball` is technically more authoritative, but that does not justify introducing a Go-only divergence today.

I have force-pushed a repurposed version of this PR that drops metadata resolution and retains only the license-download deduplication. If tarball URL handling needs to change later, it should be handled consistently across Go, .NET, Python, and Rust.

SteveSandersonMS (approved):
Thanks for the quick update. I re-reviewed the revised PR and validated it locally. The metadata-resolution divergence is gone, and the remaining license-download dedup is focused, low-risk, and useful. This version looks good to merge.

PR #2470, opened 2026-09-02
Improve Rust permission confirmation reliability

jmoseley:
## Why

The Rust SDK discarded failures from `session.permissions.handlePendingPermissionRequest` and logged success unconditionally. A JSON-RPC error or transport write failure could therefore appear successful, while a confirmation with no response could remain pending after session teardown.

This is a follow-up robustness issue found during github-app incident analysis. It does not address the incident where the runtime hung before emitting `permission.requested`, since the SDK never received that request.

## What changed

- Log confirmation success only after the RPC returns successfully.
- Report delivery failures with structured `session_id`, `request_id`, and `method` fields.
- Cancel unresolved acknowledgement waits when the session shuts down and record that the delivery outcome is unknown.
- Add deterministic duplex-transport tests for JSON-RPC errors, write failures, missing responses, teardown, and the unchanged success path.
- Verify that a missing confirmation response does not block sibling session events or independent RPCs.

## Safety

This does not retry the confirmation, send a fallback decision, impose a fixed timeout, or synthesize a session error. Once a request has been written, the runtime may already have applied it even if the acknowledgement is lost. Retrying could duplicate or contradict the original decision.

## Validation

- `cargo test --no-default-features --features test-support --test session_test permission_confirmation -- --nocapture`
- `cargo test --all-features`

### DIFF
--- a/rust/src/session.rs
+++ b/rust/src/session.rs
@@ -1567,7 +1567,7 @@ fn spawn_event_loop(
                     _ = shutdown.cancelled() => break,
                     Some(notification) = notifications.recv() => {
                         handle_notification(
-                            &session_id, &client, &handlers, &command_handlers, notification, &idle_waiter, &capabilities, &open_canvases, &event_tx,
+                            &session_id, &client, &handlers, &command_handlers, notification, &idle_waiter, &capabilities, &open_canvases, &event_tx, &shutdown,
                         ).await;
                     }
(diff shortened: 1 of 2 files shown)


SteveSandersonMS (approved):
Reviewed and validated the Rust-only reliability fix, including failure and teardown behavior.

PR #2482, opened 2026-09-02
Use application terminology for client info

SteveSandersonMS:
## Summary

- rename the public client identity fields across Node.js, Python, Go, .NET, Rust, and Java to `applicationName`/`applicationVersion` and `integrationName`/`integrationVersion` (with language-idiomatic casing)
- preserve the existing runtime JSON-RPC wire keys (`editorName`, `editorVersion`, `extensionName`, and `extensionVersion`) as an internal compatibility boundary
- update handshake tests, feature documentation, examples, API comments, and the changelog with application-centric wording

## Why

The client-info API added in #2464 has not shipped in a release, so this is the safe point to avoid making editor-specific terminology part of the permanent public SDK contract. Copilot SDK clients include non-editor applications; application and optional integration identity describe that model without changing the runtime protocol.

## Validation

- Node.js formatting, typecheck, and targeted client-info tests
- Python formatting, lint, typecheck, and targeted client-info tests
- Go formatting and targeted tests
- .NET targeted unit tests (52 passed)
- Rust formatting and handshake tests (6 passed)
- documentation example validation for TypeScript, Python, Go, and C#
- manual TypeScript TCP JSON-RPC handshake confirming neutral public names serialize to the existing wire keys
- manual Java/Jackson serialization confirming the existing wire keys are emitted without duplicate public properties

The full Java Maven lane could not run locally because the environment JDK did not trust the intercepted Maven TLS certificate; the direct Java/Jackson validation passed, and CI will cover the Maven and Spotless lanes.

### DIFF
--- a/docs/features/client-info.md
+++ b/docs/features/client-info.md
@@ -1,6 +1,6 @@
 # Client info
 
-Client info identifies the application using the Copilot SDK and, when applicable, the Copilot integration within it. Set the optional `clientInfo` client option to attribute runtime telemetry for that connection to your application instead of the runtime's own build.
+Client info identifies the application using the Copilot SDK and, when applicable, a specific integration within it. An integration is an identifiable sub-part of the application through which the SDK is used, such as an extension or plugin. Set the optional `clientInfo` client option to attribute runtime telemetry for that connection to your application instead of the runtime's own build.
 
 ## When to set client info
 
@@ -12,10 +12,12 @@ Client info has four optional string fields. Set the fields you know and omit th
 
 | Field | Example | Meaning |
 |---|---|---|
-| `applicationName` | `"acme-developer-portal"` | Name of the application using the SDK |
-| `applicationVersion` | `"2.4.0"` | Version of the application using the SDK |
-| `integrationName` | `"copilot-assistant"` | Name of the Copilot integration within the application |
-| `integrationVersion` | `"1.5.0"` | Version of the Copilot integration within the application |
+| `applicationName` | `"vscode"` | Name of the application using the SDK |
+| `applicationVersion` | `"1.124.2"` | Version of the application using the SDK |
+| `integrationName` | `"copilot-chat"` | Name of the extension, plugin, or other application sub-part using the SDK |
+| `integrationVersion` | `"0.54.0"` | Version of that extension, plugin, or application sub-part |
+
+For a standalone application without a distinct integration, set only the application fields. For example, a developer portal could set `applicationName` to `"acme-developer-portal"` and `applicationVersion` to `"2.4.0"`, leaving both integration fields unset.
 
 The SDK sends client info once when it establishes the connection. The identity applies for the lifetime of that connection.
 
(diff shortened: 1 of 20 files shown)


docs/features/client-info.md
```diff
@@ -1,21 +1,21 @@
 # Client info
 
-Client info identifies the editor or extension that hosts your Copilot SDK application. Set the optional `clientInfo` client option to attribute runtime telemetry for that connection to your host instead of the runtime's own build.
+Client info identifies the application using the Copilot SDK and, when applicable, the Copilot integration within it. Set the optional `clientInfo` client option to attribute runtime telemetry for that connection to your application instead of the runtime's own build.
```
SteveSandersonMS: In the examples below, retain the IDE example content.

For non-IDE cases, give an example of only specifying an application and not an integration, assuming that is allowed.

docs/features/client-info.md
```diff
@@ -1,21 +1,21 @@
...
 | Field | Example | Meaning |
 |---|---|---|
-| `editorName` | `"vscode"` | Name of the host editor |
-| `editorVersion` | `"1.124.2"` | Version of the host editor |
-| `extensionName` | `"copilot-chat"` | Name of the Copilot extension within the host |
-| `extensionVersion` | `"0.54.0"` | Version of the Copilot extension within the host |
+| `applicationName` | `"acme-developer-portal"` | Name of the application using the SDK |
+| `applicationVersion` | `"2.4.0"` | Version of the application using the SDK |
+| `integrationName` | `"copilot-assistant"` | Name of the Copilot integration within the application |
+| `integrationVersion` | `"1.5.0"` | Version of the Copilot integration within the application |
```
SteveSandersonMS: The terminology "integration" is unclear so don't just say "name of the integration". What is an "integration"? You should clarify that this can be used in cases like app extensions/plugins or any other situation where your SDK usage corresponds to a specific sub-part of the application.

dotnet/src/Types.cs
```diff
@@ -554,24 +554,24 @@ public sealed class TelemetryConfig
...
     /// <summary>
-    /// Version of the host editor, e.g. <c>"1.124.2"</c>.
+    /// Version of the application using the SDK.
     /// </summary>
-    public string? EditorVersion { get; set; }
+    public string? ApplicationVersion { get; set; }
 
     /// <summary>
-    /// Name of the Copilot extension within the host, e.g. <c>"copilot-chat"</c>.
+    /// Name of the Copilot integration within the application.
```
SteveSandersonMS: Needs some further definition or example. People won't know what this means. Assuming it's optional, focus on that, e.g., "Optionally specifies a named integration within the application. This can be useful for plugins or extensions within an application."

SteveSandersonMS: Do this consistently across all languages.

docs/features/client-info.md
SteveSandersonMS: Updated the language examples to retain the VS Code/Copilot Chat scenario and added a standalone developer portal example that sets only application name and version.

docs/features/client-info.md
SteveSandersonMS: Clarified that an integration is an optional identifiable application sub-part, such as an extension or plugin, and updated the field descriptions accordingly.

dotnet/src/Types.cs
SteveSandersonMS: Updated the public integration name/version documentation consistently across Node.js, Python, Go, .NET, Rust, and Java. Each now explains that integration identity is optional and applies to a named extension, plugin, or comparable application sub-part.

devm33 (approved):

PR #2496, opened 2026-09-03
Support non-object arguments for external tools

stephentoub:
External tool overrides such as `apply_patch` can receive freeform JSON values, but the .NET SDK previously assumed every invocation root was an object and failed before entering the host handler.

This change preserves named-property expansion for object roots and binds any non-object JSON value to the function's sole schema parameter. When additional parameters are optional, it selects the sole required parameter; ambiguous schemas return a clear tool error instead of guessing. Fake-server coverage exercises string and numeric values, optional parameters, successful responses, and ambiguous rejection.

**Tests**
- `dotnet test dotnet\test\GitHub.Copilot.SDK.Test.csproj --filter "FullyQualifiedName~GitHub.Copilot.Test.Unit.ClientSessionLifetimeTests" --no-restore`

Generated by Copilot

### DIFF
--- a/dotnet/src/Session.cs
+++ b/dotnet/src/Session.cs
@@ -898,9 +898,16 @@ public sealed partial class CopilotSession : IAsyncDisposable
 
             if (arguments is JsonElement incomingJsonArgs)
             {
-                foreach (var prop in incomingJsonArgs.EnumerateObject())
+                if (incomingJsonArgs.ValueKind == JsonValueKind.Object)
                 {
-                    aiFunctionArgs[prop.Name] = prop.Value;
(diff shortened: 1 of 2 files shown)


SteveSandersonMS (approved):
Approving — this is a well-scoped, verified fix for a real crash when overriding apply_patch (and any other built-in tool that sends a bare string/number argument) in the .NET SDK. I reverted the source change locally and confirmed the new tests fail with the exact EnumerateObject-on-non-object exception this PR fixes, then confirmed they pass with the fix restored and the rest of the 57-test suite is unaffected. Confirmed this is .NET-specific (other SDKs pass raw JSON through without reflective binding), so no other-language changes are needed.

PR #2504, opened 2026-09-03
Upgrade agentic workflows to gh-aw 0.88.2

devm33:
## Summary

- upgrade agentic workflow locks from gh-aw `v0.83.1` to `v0.88.2` after the previous compiler version was blocked
- add the explicit no-shell settings required by the newer compiler to five issue workflows
- pin compiled-workflow verification and generated action locks to gh-aw `v0.88.2`
- regenerate all 11 workflow lock files, remove the obsolete generated maintenance workflow, and refresh the agentic-workflows skill

## Testing

- compiled and validated all 11 agentic workflows with gh-aw `v0.88.2`
- confirmed a second compilation produced no diff

## Context

These changes were split out of #2501 so the runtime release repair remains focused. This PR must land first because the gh-aw version currently on `main` is blocked and causes agentic-workflow activation checks on other PRs, including #2501, to fail before the agent runs.


SteveSandersonMS (approved):
I confirmed that gh-aw v0.83.1 is blocked upstream, which is causing activation checks to fail on other pull requests. This change upgrades the compiler to v0.88.2, regenerates the workflow locks, adds the required no-shell settings, and passes the compiled-workflow verification and current CI checks. Approved.

PR #2513, opened 2026-09-04
Add Rust model allowlist support on pinned SDK base

jpbufe3:
Integrations pinned to SDK commit `c672e716473da0be4da7ad807df8d21259623640` need the Rust model allowlist contract without pulling in later unrelated SDK changes. This compatibility PR ports only that surface onto the exact pinned base.

## Summary

- Add optional `allowed_models` fields and fluent builders to `SessionConfig` and `ResumeSessionConfig`, serialized as `allowedModels` and omitted when unrestricted.
- Add the generated typed `session.model.setAllowedModels` RPC request/result surface for replacing or clearing a running session's restriction.
- Cover defaults, Debug output, create/resume wire conversion, request serialization, method routing, session ID injection, and typed result deserialization.

The SDK preserves exact caller-provided IDs; the runtime remains responsible for validation, policy intersection, fallback selection, and enforcement across child, sub-agent, Auto, and remote execution paths.

## Testing

- `COPILOT_SKIP_CLI_DOWNLOAD=1 cargo test --all-features allowed_models`


SteveSandersonMS:
Thanks for the contribution! I looked into this and found `session.model.setAllowedModels` isn't implemented by the shipped Copilot CLI runtime (v1.0.81-11) — calling it against a real session returns `-32601 Unhandled method`, unlike other session methods (e.g. `session.abort`) which work fine. The codegen change also hand-injects this method into the schema rather than reading it from the real runtime contract, and it's Rust-only even though model-allowlisting would need to be consistent across all SDK languages.

Could you clarify the plan for landing this in the runtime first? Once the runtime actually implements `session.model.setAllowedModels` and it shows up in the real api.schema.json, we can regenerate for all six SDKs together (with a real end-to-end test against the runtime, not just mocked dispatch tests) rather than hand-injecting a Rust-only stub here.

I'm moving this to draft for tracking — please mark it ready for review once the runtime side has landed.

jpbufe3:
> Could you clarify the plan for landing this in the runtime first?

Agreed on the sequencing. The runtime implementation is github/copilot-agent-runtime#18943 and is still open; this SDK compatibility branch supports the dependent github/github-app#14584 work and should remain draft until the runtime PR lands.

The shipped runtime returning `-32601` is expected until that implementation is merged and released. Once the runtime publishes the canonical `api.schema.json`, the live RPC should be regenerated from that schema, the temporary Rust-only overlay removed, and broader cross-SDK generation/E2E coverage handled from the real runtime contract before this is marked ready.

jpbufe3:
https://github.com/github/copilot-agent-runtime/pull/18943has merged and is being staged for prerelease! Reopening for review cc @SteveSandersonMS

SteveSandersonMS:
Thanks for the update, and glad to see github/copilot-agent-runtime#18943 landed! I re-validated against the newest available runtime build (`@github/copilot@1.0.84-1`, published 2026-09-04) since that's the most recent CLI/runtime the SDK could actually pull:

- `schemas/api.schema.json` in that package still has no `session.model.setAllowedModels` entry and no `allowedModels` field anywhere.
- I spun up the real `copilot --server --stdio` binary from that package, created a live session, and called the method directly — same result as before:
  ```
  → session.model.setAllowedModels {"sessionId": "<real-id>", "allowedModels": ["gpt-5"]}
  ← {"error": {"code": -32601, "message": "Unhandled method session.model.setAllowedModels"}}
  ```

That's expected, since the runtime PR merged 2026-09-08, four days after that CLI build was cut — the fix hasn't shipped in any published runtime/CLI package yet ("staged for prerelease" isn't the same as available). The SDK's own codegen still fabricates the schema entry itself (`addModelSetAllowedModelsToApiSchema`) rather than reading it from a real, shipped `api.schema.json`, so nothing here would catch drift if the eventual real contract differs from what's hand-coded (field names, error semantics, etc.).

I'd like to hold this in draft until a published `@github/copilot` release actually contains the runtime change, at which point we should regenerate the Rust types directly from the real schema (dropping `addModelSetAllowedModelsToApiSchema`) and add a real end-to-end test that exercises `session.model.setAllowedModels` against that shipped runtime rather than only the mocked mock-server tests. Please re-open once you can point at a published CLI/runtime version where the real schema includes this method — happy to take another look right away at that point.

PR #2519, opened 2026-09-04
docs: fix six broken hook anchor links

rinceyuan:
## Summary

`docs/hooks/session-lifecycle.md` is the only file in `docs/` that uses the explicit heading-id extension:

```markdown
## Session start hook {#session-start}
```

GitHub-flavored Markdown does not implement that extension, so the braces render as literal text inside the heading and the id is never created. The generated slug is `session-start-hook`, which means every link aimed at `#session-start`, `#session-end` or `#agent-stop` lands at the top of the page instead of the section.

Six links are affected:

* `docs/features/hooks.md` - `onSessionStart`, `onSessionEnd`
* `docs/hooks/hooks-overview.md` - `onSessionStart`, `onSessionEnd`, `onAgentStop` (twice)

This drops the three explicit ids and retargets the links at the generated slugs, matching the other 50 files in `docs/`. It also removes the stray `{#...}` text from the rendered headings.

## Verification

Walked every relative link in `docs/` and resolved each target file and anchor against the headings actually present:

* before: 6 problems, all in these three files
* after: 0 problems across 51 files

Prose and links only, no code blocks touched, so `docs-validate` is unaffected.

### DIFF
--- a/docs/features/hooks.md
+++ b/docs/features/hooks.md
@@ -22,13 +22,13 @@ flowchart LR
 
 | Hook                                                                | When it fires                       | What you can do                            |
 | ------------------------------------------------------------------- | ----------------------------------- | ------------------------------------------ |
-| [`onSessionStart`](../hooks/session-lifecycle.md#session-start)     | Session begins (new or resumed)     | Inject context, load preferences           |
+| [`onSessionStart`](../hooks/session-lifecycle.md#session-start-hook)     | Session begins (new or resumed)     | Inject context, load preferences           |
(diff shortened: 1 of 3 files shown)


SteveSandersonMS (approved):
Approving — this is a clean docs fix. I independently verified the underlying bug (GitHub renders kramdown-style `{#id}` heading attributes as literal text rather than creating the anchor id) by diffing the live rendered HTML for docs/hooks/session-lifecycle.md before and after your change, and confirmed the new anchors (#session-start-hook, #session-end-hook, #agent-stop-hook) match GitHub's actual generated ids. Scope and diff size look right. Thanks for tracking down all six affected links.

PR #2535, opened 2026-09-04
Restore Rust dual-artifact CLI bundling

dmytrostruk:
## Summary

- restore the real full Copilot CLI/Node SEA as the artifact returned by `install_bundled_cli()`
- keep `install_bundled_runtime()` and normal `CliProgram::Resolve` on the native `copilot-runtime` wrapper with adjacent `runtime.node` and required assets
- embed separate, independently checksummed CLI and runtime payloads instead of aliasing wrapper bytes under the `copilot` filename
- preserve `bundled-in-process`, no-default-feature extraction, build caches, custom extraction directories, and all eight supported platform variants
- invalidate the wrapper-as-CLI cache entry from the regression by checking the installed CLI against the embedded SEA size
- add regression coverage for distinct artifacts and the full CLI `--binary-version` contract

Fixes the Rust SDK regression introduced by #2463.

## Validation

- `cargo +nightly-2026-04-14 fmt --all -- --config-path .rustfmt.nightly.toml --check`
- `cargo clippy --all-targets --features test-support,bundled-in-process -- --no-deps -D warnings -D clippy::unwrap_used -D clippy::disallowed_macros -D clippy::await_holding_invalid_type`
- `cargo test --test cli_resolution_test` — 7 passed
- `cargo test --no-default-features --test cli_resolution_test` — 8 passed
- `cargo test --all-features --test cli_resolution_test` — 7 passed
- generated and consumed both publish checksum snapshots, each containing all eight platform hashes

## Size impact

For Copilot CLI 1.0.83-5 on Darwin ARM64:

- full CLI payload: 83,552,873 bytes
- filtered runtime payload: 40,925,644 bytes with default features, 72,594,661 bytes with `bundled-in-process`
- the restoration adds 83,552,873 bytes of compressed embedded payload; normal `CliProgram::Resolve` still extracts only the runtime bundle, while the SEA remains lazy behind `install_bundled_cli()`


roji (approved):
LGTM, fixes unintentional regression introduced in #2463.

PR #2544, opened 2026-09-04
Add runtime-driven canary and unstable SDK publishing

MackinnonBuck:
## Problem

Maintainers need branch-capable unstable Node SDK releases and runtime-driven canary/unstable releases. npm trusted publishing recognizes only `.github/workflows/publish.yml`, so every production npm publication must execute directly in that workflow.

## Changes

- Make `publish.yml` the sole production release workflow and remove the separate `sdk-canary.yml` production workflow.
- Preserve existing stable/prerelease behavior across Node, .NET, Rust, Python, Java, Go, and GitHub Releases.
- Use one dispatch surface:
  - `dist-tag`: `latest`, `prerelease`, `unstable`, or `canary`
  - `version`: optional direct SDK version/base
  - `mode`: `publish` or `dry-run`
  - `test-policy`: runtime E2E handling with `required`, `advisory`, or `skipped`
  - `runtime`: optional automation JSON with exactly `{version,sha,run_id}`
- Keep normal SDK publishing simple: direct latest/prerelease/unstable releases leave `runtime` empty and use the default `required` test policy. Direct canary is rejected because canary requires exact runtime provenance.
- Keep latest/prerelease publish-only. Canary and unstable support dry-run; dry-run performs acquisition, testing when selected, packaging, and verification without registry, tag, or release mutations.
- Parse and validate runtime JSON and test policy once, then pass validated values through job outputs.
- Keep `required` E2E failures blocking, let `advisory` tolerate only the actual Ubuntu/macOS/Windows test-command failure with explicit warnings, and let `skipped` bypass the matrix without bypassing acquisition, packaging, manifest, or publication gates.
- Test runtime-backed releases with both the default and in-process transports on Ubuntu, macOS, and Windows.
- Record the selected test policy in the retained release manifest.
- Keep exact runtime acquisition, nine-package verification, retries, idempotent publication, and dry-run mutation suppression.
- Keep direct unstable public npm → Azure ordering and runtime unstable Azure verification → public npm ordering. Canary remains Azure-only.

## Versioning

Canary versions use:

`X.Y.(Z+1)-canary.<workflow_run_number>.g<sdk_sha7>`

Generated unstable versions use:

`<target_core>-unstable.<workflow_run_id>.g<sdk_sha7>`

Explicit direct unstable overrides must use `<core>-unstable` and produce:

`<core>-unstable.<workflow_run_id>.g<sdk_sha7>`

This keeps explicit and generated releases in the same numeric-first SemVer ordering. The source runtime run ID is provenance only. Re-running the same SDK workflow run preserves its release identity; a new dispatch receives a new version.

## Coordinated canary cutover

This PR must not merge until github/copilot-agent-runtime#20157 is ready for the coordinated sender/receiver cutover:

1. Pause the runtime repository's scheduled canary automation and drain any in-flight `runtime-canary` repository dispatches.
2. Merge this SDK receiver change, then immediately merge the runtime sender change that dispatches `publish.yml` through `workflow_dispatch`.
3. Run one runtime-backed canary `dry-run` and confirm the returned SDK workflow run completes successfully.
4. Resume scheduled runtime canary publication.

The old schedule must remain paused throughout the interval in which the SDK repository no longer accepts `runtime-canary` repository dispatches.


nodejs/scripts/runtime-package-acquisition.ts
```diff
@@ -0,0 +1,264 @@
...
+        }
+        values.set(key, value);
+    }
+    return {
+        runtimeVersion: values.get("--version") ?? "",
+        runtimeSha: values.get("--sha") ?? "",
+        registry: values.get("--registry") ?? "",
+        outputDirectory: values.get("--output") ?? "",
+    };
+}
```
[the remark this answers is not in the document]
MackinnonBuck: Fixed. `runtime-package-acquisition.ts` now requires exactly one non-empty value for each of `--version`, `--sha`, `--registry`, and `--output`, and rejects unknown, duplicate, missing, or empty arguments before resolving paths or performing any filesystem operation. Focused tests cover misspelled/missing output and the other invalid argument forms.

> [!NOTE]

.github/workflows/publish.yml
```diff
@@ -237,21 +271,461 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          [ "$RUNTIME_SOURCE" = "github-packages" ] ||
+            { echo "::error::Unstable runtime_source must be github-packages."; exit 1; }
+          [[ "$RUNTIME_VERSION" =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$ ]] ||
+            { echo "::error::runtime_version must be exact SemVer."; exit 1; }
+          [[ "$RUNTIME_SHA" =~ ^[0-9a-f]{40}$ ]] ||
+            { echo "::error::runtime_sha must be a lowercase full SHA."; exit 1; }
+          [[ "$RUNTIME_RUN_ID" =~ ^[0-9]+$ ]] ||
+            { echo "::error::runtime_run_id must be numeric."; exit 1; }
```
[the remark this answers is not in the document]
MackinnonBuck: This contract was superseded by the approved provenance policy. `runtime_run_id` is now informational provenance rather than an SDK release claim or trust root, and the cross-run ledger was removed. Runtime trust is established independently by strict GitHub Packages integrity plus exact package version, source SHA, platform metadata, repository metadata, and file-content validation. A new SDK workflow run may intentionally reuse the same runtime run.

> [!NOTE]

.github/workflows/publish.yml
```diff
@@ -237,21 +271,461 @@ jobs:
...
+          path: ${{ runner.temp }}/runtime-packages
+      - name: Select the acquired runtime
+        env:
+          COPILOT_SDK_RUNTIME_PACKAGE_DIR: ${{ runner.temp }}/runtime-packages
+          RUNTIME_VERSION: ${{ needs.unstable-plan.outputs.runtime_version }}
+        run: |
+          set -euo pipefail
+          node scripts/set-cli-version.js "$RUNTIME_VERSION" --local-package
+          runtime_path="$(npm run --silent prepare:runtime -- --print-path)"
+          "$runtime_path" --version | grep -F "$RUNTIME_VERSION"
```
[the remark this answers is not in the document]
MackinnonBuck: Fixed by removing the runtime-driven unstable path from `publish.yml` and consolidating it in `runtime-sdk.yml`. The receiver now acquires the exact requested package version and `releaseArtifacts.ts` accepts the ambient acquired package only when its `package.json.version` exactly equals the requested version; otherwise it falls through to the normal controlled resolution path.

> [!NOTE]

.github/workflows/sdk-canary.yml
```diff
@@ -1,428 +1,360 @@
...
-            }
-            if (!fs.statSync(path.join(path.dirname(runtime), 'runtime.node')).isFile()) {
-              throw new Error('runtime.node is not adjacent to the runtime wrapper');
-            }
-          " "$runtime_path"
-          legacy_path=$(npm run --silent prepare:runtime -- --print-legacy-path)
-          node "$legacy_path" --version | grep -F "$RUNTIME_VERSION"
+          node scripts/set-cli-version.js "$RUNTIME_VERSION" --local-package
+          runtime_path="$(npm run --silent prepare:runtime -- --print-path)"
+          "$runtime_path" --version | grep -F "$RUNTIME_VERSION"
```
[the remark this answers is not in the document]
MackinnonBuck: Fixed in the unified `runtime-sdk.yml` flow. Runtime selection is based on the exact acquired package version rather than substring matching, and the ambient runtime directory is used only when the platform package's `package.json.version` exactly matches the requested runtime version. The old canary workflow path shown here was replaced.

> [!NOTE]

.github/workflows/sdk-canary.yml
```diff
@@ -1,428 +1,360 @@
...
-          if [[ ! "$SDK_VERSION" =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$ ]]; then
-            echo "::error::Computed SDK canary version '$SDK_VERSION' is not valid semver."
-            exit 1
-          fi
-          echo "SDK canary version: $SDK_VERSION"
-          echo "SDK_VERSION=$SDK_VERSION" >> "$GITHUB_OUTPUT"
-
-      - name: Set package and runtime versions
+          PUBLIC_LATEST="$(node scripts/get-version.js current)"
+          BASE="${PUBLIC_LATEST%%-*}"
```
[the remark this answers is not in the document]
MackinnonBuck: Fixed by moving canary version planning into the tested `unstable-version.ts` SemVer implementation. It parses eligible release tags with the `semver` package instead of shell-splitting the raw version, so build metadata cannot leak into patch arithmetic. The baseline is also bounded by the immutable workflow creation time for same-run determinism.

> [!NOTE]

.github/workflows/publish.yml
```diff
@@ -485,3 +592,371 @@ jobs:
...
+        env:
+          NODE_AUTH_TOKEN: ${{ github.token }}
+        run: echo "//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}" > "$HOME/.npmrc"
+      - name: Download and validate all runtime platforms
+        env:
+          NODE_AUTH_TOKEN: ${{ github.token }}
+          RUNTIME_SHA: ${{ needs.validate-dispatch.outputs.runtime_sha }}
+          RUNTIME_VERSION: ${{ needs.validate-dispatch.outputs.runtime_version }}
+        run: |
+          npm run acquire:runtime-packages -- \
```
devm33, line 680: doesn't this need github token?

MackinnonBuck: This is written to `.npmrc` in the step just above

nodejs/scripts/unstable-version.ts
```diff
@@ -0,0 +1,206 @@
...
+        if (
+            !parsed ||
+            semver.valid(options.versionOverride) !== options.versionOverride ||
+            parsed.prerelease[0] !== "unstable"
+        ) {
+            throw new Error(
+                `Explicit unstable SDK version must be valid SemVer with an unstable prerelease: ${options.versionOverride}`
+            );
+        }
+        return `${parsed.major}.${parsed.minor}.${parsed.patch}-${parsed.prerelease.join(".")}.${options.runId}.g${options.sdkSha.slice(0, 7)}`;
```
roji: This allows a manual override to block subsequent generated releases for the same version core. For example, the accepted base `1.0.14-unstable.manual.1` produces `1.0.14-unstable.manual.1.34640000000.gabcdef0`. A later automatic run produces `1.0.14-unstable.34640000001.gabcdef0`, which SemVer considers *older*: a textual prerelease identifier (`manual`) sorts above any numeric identifier. `publishManifest()` then refuses publication with `refusing to rewind`. I reproduced this through the version helper and publisher with a controlled registry response.

Please keep manual and generated releases in the same numeric-first ordering. The simplest option may be to restrict the override to `<core>-unstable`, rather than supporting arbitrary prerelease suffixes. If labels are needed, put them after the shared run identity. Add a regression case that publishes/plans an override and then a later generated release with the same core.

.github/workflows/publish.yml
```diff
@@ -485,3 +592,371 @@ jobs:
...
+  runtime-test:
+    name: Test runtime (${{ matrix.os }})
+    needs: [validate-dispatch, runtime-plan, runtime-acquire]
+    if: needs.validate-dispatch.outputs.test_policy != 'skipped'
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
roji, line 772: This tests the new runtime only with the default transport. Ordinary Node CI already runs both `default` and `inprocess` via `COPILOT_SDK_DEFAULT_CONNECTION`, but those checks exercise the repository's pinned runtime, not the exact runtime supplied to this release. A runtime can therefore pass this release gate while breaking the SDK's FFI transport.

Please reuse the existing transport-matrix pattern for the runtime-backed gate. This should exercise the existing tests against the acquired runtime, not introduce another test suite or transport abstraction.

nodejs/test/release-workflows.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    it("preserves executable modes across runtime package artifact boundaries", () => {
+        expect(runtimeAcquireJob).toContain(
+            'tar -czf "$RUNNER_TEMP/runtime-packages.tar.gz" -C "$RUNNER_TEMP" runtime-packages'
+        );
+        for (const consumer of [runtimeTestJob, runtimePackageJob]) {
+            expect(consumer).toContain("if command -v cygpath >/dev/null 2>&1; then");
+            expect(consumer).toContain('runner_temp="$(cygpath -u "$runner_temp")"');
+            expect(consumer).toContain(
+                'tar -xzf "$runner_temp/runtime-package-artifact/runtime-packages.tar.gz" -C "$runner_temp"'
```
roji: This is representative of why I would substantially trim this file. The test claims to preserve executable modes, but only checks that particular shell strings exist; it never archives/extracts a fixture or checks its mode. Elsewhere the suite freezes job order, input order, helper names, and exact condition text without evaluating whether failed prerequisites actually prevent publication. Harmless refactoring breaks these assertions, while broken behavior can still satisfy them.

Please remove the bulk of the 235-line substring suite and retain a few checks of actual release behavior: dispatch validation, version ordering, package selection, and an archive/install smoke. If essential YAML wiring needs assertions, inspect its structure rather than requiring a particular spelling. I would not add a general Actions interpreter or another abstraction layer to replace this.

.github/workflows/publish.yml
```diff
@@ -485,3 +592,371 @@ jobs:
...
+          NODE_AUTH_TOKEN: ${{ github.token }}
+          RUNTIME_SHA: ${{ needs.validate-dispatch.outputs.runtime_sha }}
+          RUNTIME_VERSION: ${{ needs.validate-dispatch.outputs.runtime_version }}
+        run: |
+          npm run acquire:runtime-packages -- \
+            --version "$RUNTIME_VERSION" \
+            --sha "$RUNTIME_SHA" \
+            --output "$RUNNER_TEMP/runtime-packages"
+      - name: Archive validated runtime packages
+        run: tar -czf "$RUNNER_TEMP/runtime-packages.tar.gz" -C "$RUNNER_TEMP" runtime-packages
```
roji: This archives two copies of the runtime payload. Acquisition retains all eight original npm tarballs under `runtime-packages/tarballs` as well as the extracted platform directories, and this command includes both. All three test jobs and the packaging job download that archive, but their runtime resolution only consumes the extracted directories.

Please exclude the original `tarballs` directory from this handoff, retaining the extracted packages and acquisition metadata. Keeping the outer tar archive to preserve executable modes is useful; transferring the unused compressed copies alongside the extracted payload is not. This is a non-blocking size/time simplification.

.github/workflows/publish.yml
```diff
@@ -485,3 +592,371 @@ jobs:
...
+          SDK_REF: ${{ github.ref }}
+          SDK_SHA: ${{ github.sha }}
+          SDK_VERSION: ${{ needs.runtime-plan.outputs.sdk_version }}
+          TEST_POLICY: ${{ needs.validate-dispatch.outputs.test_policy }}
+          WORKFLOW_CREATED_AT: ${{ needs.runtime-plan.outputs.workflow_created_at }}
+          WORKFLOW_RUN_ID: ${{ github.run_id }}
+          WORKFLOW_RUN_NUMBER: ${{ github.run_number }}
+        run: |
+          npm run release:manifest -- create release-manifest.json .
+          npm run release:manifest -- verify release-manifest.json .
```
roji: Small simplification: the `create` command already writes the manifest and calls `verifyReleaseManifest()` before returning (`release-manifest.ts:224-242`). This immediately repeats the same verification and rehashes all nine tarballs, with no artifact boundary or intervening mutation.

Please drop this second invocation. Keep verification after downloading artifacts in the publication jobs, where there is a meaningful boundary to validate.

.github/workflows/sdk-canary.yml
```diff
@@ -1,428 +0,0 @@
...
-        description: "publish (tests must pass), publish-force (publish even if tests fail), or tests-only (run gate, never publish)"
-        required: false
-        type: choice
-        default: publish
-        options:
-          - publish
-          - publish-force
-          - tests-only
-  repository_dispatch:
-    types: [runtime-canary]
```
roji, line 44: Deleting this receiver before switching the sender creates a canary gap. The current `github/copilot-agent-runtime` main workflow still POSTs `event_type: runtime-canary` to the SDK repository, whereas the replacement here accepts only `workflow_dispatch`. The documented SDK-first merge order therefore leaves old dispatches with no receiver until github/copilot-agent-runtime#20157 lands; a successful repository-dispatch response is not evidence that an SDK run was started.

Please make the cutover explicit: pause and drain the old scheduled/in-flight flow, merge the receiver and sender changes in the coordinated window, then resume. Alternatively retain the old receiver until the sender migration is complete. This can be an operational step rather than permanent compatibility plumbing, but the merge order alone is insufficient.

roji (approved):
LGTM, it's lots of new code (~2300 net new lines) - maybe see if there's opportunity to reduce, but otherwise looks good and see comments below.

.github/workflows/sdk-canary.yml
MackinnonBuck: Updated the PR rollout plan to make this an explicit coordinated cutover rather than relying on merge order: this PR is blocked until github/copilot-agent-runtime#20157 is ready; pause and drain the old scheduled/in-flight canary flow, merge the SDK receiver and runtime sender changes in the coordinated window, validate one runtime-backed canary dry-run, then resume the schedule. The old schedule remains paused for the entire receiver gap.

nodejs/test/release-workflows.test.ts
MackinnonBuck: Removed the 235-line workflow substring suite. Dispatch validation, version ordering, package selection/integrity, and publication behavior remain covered by their focused tests. The executable-mode claim is now exercised through a real runtime handoff: the test creates source packages with executable payloads, archives the acquired directory with the workflow's exclusion behavior, extracts it, validates all eight packages, and checks the executable mode on platforms that expose POSIX mode bits.

.github/workflows/publish.yml
MackinnonBuck: Expanded the runtime-backed test matrix to `default` and `inprocess` for Ubuntu, macOS, and Windows, reusing the existing `COPILOT_SDK_DEFAULT_CONNECTION=inprocess` selection pattern from normal Node CI.

.github/workflows/publish.yml
MackinnonBuck: Removed the immediate second verification after `create`. The create command's built-in verification remains, and the publication jobs still verify after downloading the retained artifact.

nodejs/scripts/unstable-version.ts
MackinnonBuck: Restricted explicit overrides to exactly `<core>-unstable`, so both explicit and generated versions place the numeric workflow run ID immediately after `unstable`. Added regression coverage that plans an explicit release and a later generated release for the same core and verifies the latter sorts higher. Arbitrary prerelease suffixes and build metadata are now rejected.

.github/workflows/publish.yml
MackinnonBuck: The handoff archive now excludes `runtime-packages/tarballs` while retaining the eight extracted platform directories and `runtime-packages.json`. The new archive smoke coverage also verifies that the tarball directory is absent after extraction.

stephentoub (approved):

PR #2554, opened 2026-09-07
Fix update-protocol-version.ts templates that break the Go and .NET builds

rinceyuan:
`nodejs/scripts/update-protocol-version.ts` is the only generator for the five `SdkProtocolVersion` constant files, and it is wired up as `npm run update:protocol-version`. Its Go and C# templates have drifted from the files they are supposed to produce, so **running it today reverts two deliberate fixes and breaks the Go and .NET builds.**

| Template | Emits | Committed file has |
| --- | --- | --- |
| Go | `SdkProtocolVersion` / `GetSdkProtocolVersion` | `SDKProtocolVersion` / `GetSDKProtocolVersion` |
| C# | `namespace GitHub.Copilot.SDK;` | `namespace GitHub.Copilot;` |

## How it drifted

Both outputs are marked `DO NOT EDIT`, but both were hand-corrected later — for good reasons — without updating the templates:

| Date | PR | Change |
| --- | --- | --- |
| 2026-05-06 | #1164 | Last change to the generator (added the Rust output) |
| 2026-05-21 | #1343 | C# API review fixes — namespace settled on `GitHub.Copilot` |
| 2026-06-01 | #1527 | Consolidate Go initialism casing — `Sdk` → `SDK` |

Nothing caught it: `codegen-check.yml` runs `scripts/codegen`, which is a different generator and does not produce these files. No workflow runs `update:protocol-version`.

## Reproducing the breakage

On `main`, before this change:

```console
$ cd nodejs && npm run update:protocol-version
$ cd ../go && go build ./...
# github.com/github/copilot-sdk/go
.\client.go:1974:16: undefined: GetSDKProtocolVersion
```

```console
$ dotnet build src/GitHub.Copilot.SDK.csproj
dotnet/src/Client.cs(2174,26): error CS0103: The name 'SdkProtocolVersion' does not exist in the current context
  [GitHub.Copilot.SDK.csproj::TargetFramework=netstandard2.0]
  [GitHub.Copilot.SDK.csproj::TargetFramework=net8.0]
  [GitHub.Copilot.SDK.csproj::TargetFramework=net10.0]
```

`SdkProtocolVersion` is an `internal static class`, and `Client.cs` is in `GitHub.Copilot` with no `using` for the nested namespace, so moving it breaks the reference on every target framework.

## Verification after this change

Regenerating is now a no-op — the generator reproduces all five committed files exactly:

```console
$ cd nodejs && npm run update:protocol-version
Generating SDK protocol version constants for version 3...
  ✓ nodejs/src/sdkProtocolVersion.ts
  ✓ go/sdk_protocol_version.go
  ✓ python/copilot/_sdk_protocol_version.py
  ✓ dotnet/src/SdkProtocolVersion.cs
  ✓ rust/src/sdk_protocol_version.rs
Done!

$ git status --porcelain
                      # (no output — nothing regenerated differently)
```

(On a Windows checkout the five files show as modified after regenerating, because the generator writes LF while the working tree is CRLF. The content is identical — confirm with `git diff --ignore-all-space`.)

Builds pass with the regenerated files:

```console
$ cd go && go build ./... && go vet ./...      # exit 0
$ cd dotnet && dotnet build src/GitHub.Copilot.SDK.csproj   # succeeds, all 3 TFMs
```

## Also included

The header comment listed four output files; the generator has written five since #1164 added the Rust output. Added `rust/src/sdk_protocol_version.rs` to the list.

## Not addressed here

`java/sdk/src/main/java/com/github/copilot/SdkProtocolVersion.java` carries the same `Code generated by update-protocol-version.ts. DO NOT EDIT.` header, but this script does not write it — `codegen-check.yml` value-checks it separately instead. That looked like a distinct concern, so I left it alone.

The generator's existing Prettier deviations are also untouched: `nodejs` only runs `format:check`/`lint` over `src/**` and `test/**`, and the file does not satisfy `prettier --check` on `main` either, so reformatting would have been unrelated noise.


SteveSandersonMS:
Thanks for the careful write-up and validation. After looking at this path again, we're not using `nodejs/scripts/update-protocol-version.ts` in practice, and we're planning to remove it rather than continue maintaining it. Since this PR only updates that unused script, I'm going to close it rather than merge the maintenance fix.

PR #2555, opened 2026-09-07
Fix stale codegen references in copilot-instructions.md

rinceyuan:
`.github/copilot-instructions.md` points contributors and coding agents at a type-generation entry point that was removed in February. Four references are stale.

| Line | Says | Reality |
| --- | --- | --- |
| 16 | `nodejs/scripts/generate-session-types.ts` | File does not exist |
| 26 | `npm run generate:session-types` | Script does not exist |
| 71 | `scripts/codegen/java.ts` | Real path is `java/scripts/codegen/java.ts` |
| 72 | `npm run generate:session-types` | Script does not exist |

#464 (`RPC codegen`, 2026-02-13) deleted the script and replaced the npm task in the same commit:

```console
$ git show 7e069fd6 -- nodejs/package.json
-    "generate:session-types": "tsx scripts/generate-session-types.ts",
+    "generate": "cd ../scripts/codegen && npm run generate",

$ git log --diff-filter=D --oneline -1 -- nodejs/scripts/generate-session-types.ts
7e069fd6 RPC codegen (#464)
```

The document already had the correct command elsewhere, so it contradicted itself — line 67 says `cd nodejs && npm run generate` while lines 26 and 72 named the deleted task. This aligns them.

## Verification

`generate:session-types` survived nowhere but this file, and nothing references it after the change:

```console
$ git grep -rn -E 'generate:session-types|nodejs/scripts/generate-session-types'
                      # (no output)
```

Every path the file now claims resolves:

```console
$ ls -d scripts/codegen java/scripts/codegen/java.ts java/sdk/src/generated/java nodejs/src/generated
scripts/codegen  java/scripts/codegen/java.ts  java/sdk/src/generated/java  nodejs/src/generated
```

`scripts/codegen` is confirmed as the producer of the Node output that line 72 describes:

```console
$ grep -n 'nodejs/src/generated' scripts/codegen/typescript.ts
602:    const outPath = await writeGeneratedFile("nodejs/src/generated/session-events.ts", annotatedTs);
960:    const outPath = await writeGeneratedFile("nodejs/src/generated/rpc.ts", rpcTs);
```

## Scope

Only the four stale references. The Maven command on line 71 is left as-is — the `codegen` profile is real (`java/sdk/pom.xml`, whose own comment documents `mvn generate-sources -Pcodegen`), and its `workingDirectory` is `${project.parent.basedir}/scripts/codegen`, which confirms `java/scripts/codegen` as the correct generator location.


SteveSandersonMS (approved):
Verified: the four stale references to `nodejs/scripts/generate-session-types.ts` / `npm run generate:session-types` and the Java codegen path are correctly fixed. I independently confirmed the old script/task no longer exist, `npm run generate` correctly delegates to `scripts/codegen`, and `java/scripts/codegen/java.ts` is the real path. Repo-wide grep shows no remaining references to the deleted script/task. Scope is precisely the 4 lines described. LGTM.

PR #2561, opened 2026-09-08
perf(sdk): make CopilotClient.start() single-flight

### LINKED ISSUE #2560: CopilotClient.start() is not single-flight — concurrent auto-start spawns duplicate CLI processes
## TL;DR

Several concurrent first calls on one `CopilotClient` can each start a separate
Copilot CLI process. The client only tracks the last process handle
(`this.cliProcess`), so `stop()` cannot terminate the earlier ones — a duplicate
cold-start cost plus a real process leak. The fix makes `start()` single-flight:
all concurrent callers await one in-progress startup. A three-call reproduction
then spawns one CLI (instead of three) and leaves zero orphans after `stop()`.

## Impact

- One logical startup can pay for multiple CLI boots, native-addon loads, and
  settings reads (CPU/memory/disk contention).
- Orphaned CLI child processes survive `stop()`, and temp/session resources they
  own can remain in use.
- Triggered without any explicit `start()` call — e.g. firing several
  `createSession()` calls with `Promise.all()` on a fresh client.

## Root cause

- `start()` returns early only when the state is already `"connected"`. The
  `"connecting"` state has no stored in-flight promise, so a second concurrent
  caller re-enters the whole start body.
- `createSession()` and `resumeSession()` auto-start when `!this.connection`, so
  the race happens even when the app never calls `start()` directly.
- `startCLIServer()` assigns `this.cliProcess = spawn(...)`; a second spawn
  overwrites it and orphans the first, which `stop()` never terminates.

## Reproduction

```ts
const client = new CopilotClient(); // autoStart defaults to true
await Promise.all([
  client.createSession({ onPermissionRequest: approveAll }),
  client.createSession({ onPermissionRequest: approveAll }),
  client.createSession({ onPermissionRequest: approveAll }),
]);
// Expected: 1 CLI process; after stop(), 0 remain.
// Actual:   up to 3 CLI processes; after stop(), the extras are orphaned.
```

## Proposed fix

- Add a private `startPromise`. In `start()`: if connected, return; if a start is
  in flight, return that same promise; otherwise run the start body once and
  clear `startPromise` in a `finally` so a failed start can be retried.
- Move the existing start body into a private `doStart()`; startup order is
  unchanged.

## Acceptance criteria

- Concurrent `createSession`/`resumeSession`/`start()` on a fresh client spawn
  exactly one CLI process.
- After `stop()`, zero CLI processes remain.
- A failed start can be retried by a later `start()`.
- Regression tests covering the concurrent-start and retry-after-failure cases.

DonJayamanne:
## Summary

`CopilotClient.start()` was not single-flight: concurrent first calls (including
auto-start from `createSession()` / `resumeSession()`) could each spawn a
separate CLI process. The client only tracked the last `this.cliProcess`, so
`stop()` could not terminate the earlier ones — a duplicate cold-start cost and a
real process leak.

This guards `start()` with a shared in-flight `startPromise`: concurrent callers
await one startup, and the guard is cleared on completion so a failed start can
be retried. The original start sequence moved into a private `doStart()`; startup
order is unchanged and no public API changed.

Closes #2560

## Changes

- `nodejs/src/client.ts`: add `startPromise`; `start()` returns the in-flight
  start when one is running; `doStart()` holds the original startup sequence.
- `nodejs/test/client.test.ts`: regression tests for single-flight and
  retry-after-failure.

## Verification

- `npm test -- test/client.test.ts -t 'single-flight'` → 2 passed.
- `tsc --noEmit` → clean.
- Manual reproduction (3 concurrent `createSession()` on a fresh client):
  3 → 1 CLI process spawned; 2 → 0 orphaned after `stop()`.

## Notes

- This branch is based on an earlier `main`; a rebase may be needed before merge.

### DIFF
--- a/nodejs/src/client.ts
+++ b/nodejs/src/client.ts
@@ -448,6 +448,8 @@ export class CopilotClient {
     private runtimePort: number | null = null;
     private actualHost: string = "localhost";
     private state: "disconnected" | "connecting" | "connected" | "error" = "disconnected";
+    /** Shared in-flight start; concurrent callers await it instead of spawning another CLI. */
+    private startPromise: Promise<void> | null = null;
     private sessions: Map<string, CopilotSession> = new Map();
     private stderrBuffer: string = ""; // Captures CLI stderr for error messages
     /** Resolved connection mode chosen in the constructor. */
@@ -935,6 +937,20 @@ export class CopilotClient {
             return;
         }
 
+        // Concurrent callers share one in-progress start instead of each spawning a CLI.
+        if (this.startPromise) {
+            return this.startPromise;
+        }
+
+        this.startPromise = this.doStart();
+        try {
+            await this.startPromise;
+        } finally {
+            this.startPromise = null;
+        }
+    }
+
+    private async doStart(): Promise<void> {
         this.forceStopping = false;
         this.connectionClosed = false;
         this.processTransportError = null;
--- a/nodejs/test/client.test.ts
+++ b/nodejs/test/client.test.ts
@@ -62,6 +62,64 @@ describe("approveAll", () => {
 });
 
 describe("CopilotClient", () => {
+    it("start() is single-flight: concurrent callers share one startup", async () => {
+        const client = new CopilotClient({ autoStart: false });
+        onTestFinished(() => client.forceStop());
+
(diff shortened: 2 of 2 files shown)


DonJayamanne:
@copilot why is my CI job failing is it caused by my PR?

PR #2567, opened 2026-09-08
fix(rust): isolate GitHub token callbacks from request routing

### LINKED ISSUE #2424: Isolate Rust GitHub token callbacks from the global request router
## Summary

The Rust SDK handles `gitHubToken.getToken` by awaiting `GitHubTokenRegistry::dispatch` directly inside `SessionRouter`'s single request-routing loop.

Because the registered token provider is user-supplied async code, a slow or indefinitely pending callback prevents the router from dispatching unrelated CLI-to-SDK requests across every session owned by that client. A provider panic can also terminate the routing task, leaving subsequent requests unanswered.

This path was introduced with the session-scoped GitHub token provider support in #2412.

## Expected behavior

Token acquisition should be isolated from the global routing loop so one provider cannot stall unrelated sessions or request types. Provider panics should be contained and returned to the corresponding JSON-RPC request as an internal error rather than terminating request routing.

The implementation should preserve intentional ordering/concurrency semantics for requests targeting the same provider or session.

## Suggested coverage

- A pending token callback does not prevent an unrelated request from being routed.
- A panicking token callback receives an error response.
- Request routing remains operational after a provider panic.

xianjianlf2:
Fixes #2424.

A pending GitHub token callback currently holds up the single Rust request-routing loop; a callback panic terminates that loop without replying. Give each token-provider registration a lazy FIFO worker so other providers and session requests can continue while preserving callback order within a registration. Catch callback panics and return an internal RPC error without including the panic payload.

The registration owns and aborts its worker on retirement. Workers retain a weak client reference, so pending callbacks do not keep the client alive. No public API or protocol changes.

Three regression tests exercise blocked-provider isolation (including same-provider FIFO and an unrelated user-input request), panic recovery, and cancellation on session deletion, force-stop, and dropping session/client handles. The blocking and panic tests fail on the base commit.

Validation:
- 234 Rust unit tests and 135 session integration tests passed using in-memory JSON-RPC streams.
- Pinned nightly rustfmt check passed.
- Clippy across all targets with `test-support,bundled-in-process` and the repository's strict warning flags passed.
- `git diff --check` passed.

Validation used `COPILOT_SKIP_CLI_DOWNLOAD=1`; live CLI/provider E2E tests were not run. No new dependencies. Codex assisted with implementation and validation.

### DIFF
--- a/rust/src/github_token.rs
+++ b/rust/src/github_token.rs
@@ -2,11 +2,15 @@
 
 use std::collections::HashMap;
 use std::future::Future;
+use std::panic::AssertUnwindSafe;
 use std::sync::{Arc, OnceLock, Weak};
 
 use async_trait::async_trait;
+use futures_util::FutureExt;
 use parking_lot::Mutex;
 use serde_json::Value;
+use tokio::sync::mpsc;
+use tokio::task::JoinHandle;
 
 use crate::generated::api_types::{
     GitHubTokenAcquireReason, GitHubTokenAcquireRequest, GitHubTokenAcquireResult,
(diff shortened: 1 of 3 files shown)


SteveSandersonMS (approved):
I reproduced the router-blocking failure on the pre-PR commit using the new regression test, then confirmed the Rust GitHub token provider tests pass on this branch. The fix is scoped to the Rust token-provider dispatch path, preserves FIFO ordering within a registration, isolates callbacks across registrations, and contains provider panics by returning an RPC error instead of killing request routing.

The similar callback-awaiting pattern in extension launch handling is worth tracking separately, but this PR is a correct and minimal fix for #2424.

PR #2568, opened 2026-09-08
ci: check protocol version drift across all SDKs

### LINKED ISSUE #1165: Codegen Check workflow doesn't detect SDK protocol version drift
The `Codegen Check` workflow (`.github/workflows/codegen-check.yml`) regenerates schema-driven types via `cd scripts/codegen && npm run generate` and fails if the working tree is dirty. It does **not** run `cd nodejs && npm run update:protocol-version`, which is the script that regenerates the per-language SDK protocol version constants:

- `nodejs/src/sdkProtocolVersion.ts`
- `dotnet/src/SdkProtocolVersion.cs`
- `python/copilot/_sdk_protocol_version.py`
- `go/sdk_protocol_version.go`
- `rust/src/sdk_protocol_version.rs`

Source of truth: `sdk-protocol-version.json`. If someone edits `sdk-protocol-version.json` (or hand-edits one of the generated files) without running the script, CI won't catch it. Each SDK can drift independently.

### Proposal

Extend `codegen-check.yml`:

1. Add an `npm run update:protocol-version` step after `npm run generate`.
2. Extend the `paths:` trigger to include all the generated outputs and the script itself, so the workflow runs on PRs that touch any of them:
   ```yaml
   - 'sdk-protocol-version.json'
   - 'nodejs/scripts/update-protocol-version.ts'
   - 'nodejs/src/sdkProtocolVersion.ts'
   - 'dotnet/src/SdkProtocolVersion.cs'
   - 'python/copilot/_sdk_protocol_version.py'
   - 'go/sdk_protocol_version.go'
   - 'rust/src/sdk_protocol_version.rs'
   ```
3. Update the failure message to mention both regen commands.

### Context

I noticed this while working on #1164 (Rust SDK technical preview). I'd originally bundled this fix into that PR but reverted it to keep the Rust port surgical. Tracking it here as a follow-up.

<sub><em>&nbsp;&nbsp;Generated via Copilot (Claude Opus 4.7) <a href="https://adaptivepatchwork.com/ai-attribution/">on behalf</a> of @tclem</em></sub>

xianjianlf2:
Fixes #1165.

Codegen Check explicitly validates only Java's protocol constant, and changes to the other SDK constants do not trigger it. Extend the check to all six SDKs and include the constants plus the protocol-version generator in the workflow paths. A mismatch or missing declaration produces a file-specific annotation before installing codegen dependencies.

This validates the declared values against `sdk-protocol-version.json` without regenerating the files. It extends the existing Java consistency check and stays independent of the generator template fixes in #2554; generated source and language-specific declaration shapes remain unchanged.

Validation: parsed the workflow and executed its exact Node check in temporary fixtures for 15 cases: valid baseline, six individual version mismatches, six missing declarations, a source-version bump, and invalid source metadata. Verified all six constant paths and the generator trigger the workflow. Actionlint 1.7.7 and `git diff --check` passed. The full schema-generation workflow was not run locally. Codex assisted with this change and validation.

### DIFF
--- a/.github/workflows/codegen-check.yml
+++ b/.github/workflows/codegen-check.yml
@@ -15,6 +15,12 @@ on:
       - 'go/rpc/**'
       - 'rust/src/generated/**'
       - 'sdk-protocol-version.json'
+      - 'nodejs/scripts/update-protocol-version.ts'
+      - 'nodejs/src/sdkProtocolVersion.ts'
+      - 'dotnet/src/SdkProtocolVersion.cs'
+      - 'python/copilot/_sdk_protocol_version.py'
+      - 'go/sdk_protocol_version.go'
+      - 'rust/src/sdk_protocol_version.rs'
       - 'java/sdk/src/main/java/com/github/copilot/SdkProtocolVersion.java'
       - '.github/workflows/codegen-check.yml'
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks for looking into this and for the careful validation notes. After reconsidering the need, I don't think we should add more workflow/tooling here right now. The SDK protocol version is not something we expect to change frequently, and when it does change the manual regeneration/review step is small enough that extra automation in CI is likely more maintenance surface than value.

I'm going to close this PR, but I appreciate the cleanup-minded contribution.

PR #2569, opened 2026-09-08
Add max_output_tokens to ModelCapabilities limits

1fanwang:
## Summary

TypeScript callers cannot set an output token limit through the session config, even though the wire protocol accepts one. Setting a value fails to compile with TS2561, so the only way to reach the field is to cast away the type.

## Problem

The generated RPC schema declares an optional output token limit on the model capabilities override. The hand-written public model capabilities type omits it. The session config override is a deep partial of that public type, so the compiler rejects the field for a value the protocol already accepts.

## Solution

Add the optional field to the public limits type, next to the existing prompt token limit, so the public surface matches the generated schema.

## Testing Done

Compiled a probe that sets the field on a session config against the base branch and this branch, and ran the repository typecheck gate.

<details>
<summary>Raw logs</summary>

```console
$ printf 'import type { SessionConfig } from "./src/types";\nconst c: SessionConfig = { modelCapabilities: { limits: { max_output_tokens: 8192 } } };\nvoid c;\n' > probe-mot.ts
$ npx tsc --noEmit --skipLibCheck --target es2020 --moduleResolution node probe-mot.ts

# base branch:
probe-mot.ts(2,59): error TS2561: Object literal may only specify known properties, but
'max_output_tokens' does not exist in type '{ max_prompt_tokens?: number; ... }'.
Did you mean to write 'max_prompt_tokens'?

# this branch:
(no output, compiles cleanly)

$ npm run typecheck
> tsc --noEmit && tsc --noEmit -p tsconfig.test.json
(no output, both configs pass)
```

</details>


SteveSandersonMS (approved):
Confirmed: the generated RPC schema (`ModelCapabilitiesOverrideLimits`) already declares `max_output_tokens`, and this field is already present in the hand-written public wrapper types for Python, Rust, Java, Go, and .NET — Node's `ModelCapabilities` was the only one missing it. I reproduced the reported TS2561 compile error locally by reverting this change and confirmed the fix resolves it cleanly. Minimal, correctly-scoped, no further test/doc changes needed. Approving.

PR #2578, opened 2026-09-08
Document capi.enableWebSocketResponses as a resume option

1fanwang:
## Why

Someone resumes a session and it starts returning `400 input item ID does not belong to this connection`, or their proxy refuses WebSocket connections so resume never gets off the ground. Both are transport problems, and the SDK already lets them pick the transport on the resume call. They go to the session persistence guide, read the table of what can be reconfigured on resume, and the option is not there. Nothing in the docs suggests it exists, so the reasonable conclusion is that it does not and the session is lost.

The option is real. The resume config accepts it through the shared base config, every language binding exposes it, and it is covered by tests. Only the documentation was missing, and the table already lists its sibling `capi.autoTier`, which is what makes the omission read as deliberate.

After this change the reader finds the option in the same table they were already looking at, with a short section showing how to pass it.

## What changed

One row in the resume options table, plus a section giving the default, the Python spelling, a resume example, and how it relates to the equivalent environment variable.

## Testing Done

Documentation only, no code paths touched.

The option was absent from the docs before this change and present after:

```
$ git grep -c enableWebSocketResponses d5c9d06d -- docs/
(no output, exit status 1)

$ git grep -c enableWebSocketResponses HEAD -- docs/
HEAD:docs/features/session-persistence.md:3
```

The added example is also compiled by the existing docs validation, so a wrong option name or type would fail the build rather than render quietly.

<details><summary>Raw logs: docs validation (tail)</summary>

```
$ npm run validate:ts
TypeScript:
  ✅ 190 files passed

```

</details>

### DIFF
--- a/docs/features/session-persistence.md
+++ b/docs/features/session-persistence.md
@@ -243,6 +243,7 @@ When resuming a session, you can optionally reconfigure many settings. This is u
 | `excludedTools` | Disable specific tools |
 | `provider` | Re-provide BYOK credentials (required for BYOK sessions) |
 | `capi.autoTier` | Override the persisted Auto routing preference |
+| `capi.enableWebSocketResponses` | Choose the Responses API transport for the resumed session |
 | `reasoningEffort` | Adjust reasoning effort level |
 | `streaming` | Enable/disable streaming responses |
 | `workingDirectory` | Change the working directory |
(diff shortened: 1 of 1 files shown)


SteveSandersonMS (approved):
Verified: `capi.enableWebSocketResponses` is a real, tested field on `ResumeSessionConfig` across all 6 SDKs (Node, .NET, Python, Go, Rust, Java), shipped in #1711 but missing from this table alongside its sibling `capi.autoTier`. Re-ran `npm run validate:ts` in `scripts/docs-validation` myself and confirmed the new snippet compiles (190/190 passed). Rebased onto current main to clear the merge-state block.

PR #2580, opened 2026-09-08
Validate the examples that follow an unvalidated language fence

1fanwang:
## Why

A maintainer who wraps an example in `<!-- docs-validate: hidden -->` is asking for that example to be type-checked. For three examples in this repository, it silently was not, and nothing reported a problem. The most visible one is the trusted host-bundled block in the plugin directories guide, which carries those markers for exactly that reason.

Closing a hidden block sets a pending skip so the visible copy that follows is not validated twice. A fence whose language has no validator, `rust` among them, was never recognized as a code block at all, so it never consumed that pending skip. The flag survived and the next validated block absorbed it instead.

Rust examples appear in nine feature guides, usually as the last language tab before the next section, so any example following one is exposed. Before this change those three blocks were extracted and discarded. After it they are checked like every other example, and a broken one would fail the build.

## What changed

Fences in languages that have no validator are now tracked and consume a pending skip when they close, so the directive can no longer reach past its intended target.

## Testing Done

Three previously discarded examples are now extracted and validated:

```
auth/authenticate.md:412
features/plugin-directories.md:248
features/session-limits.md:164
```

The two TypeScript ones pass locally. The C# one needs a .NET toolchain I do not have, so CI covers it.

<details><summary>Raw logs: extraction before and after</summary>

```
$ git checkout upstream/main -- scripts/docs-validation/extract.ts
$ (cd scripts/docs-validation && npm run extract) | grep -E "typescript|csharp|Total|Skipped"
  typescript     189
  csharp         56
  Total          394
  Skipped        239

$ git checkout HEAD -- scripts/docs-validation/extract.ts
$ (cd scripts/docs-validation && npm run extract) | grep -E "typescript|csharp|Total|Skipped"
  typescript     191
  csharp         57
  Total          397
  Skipped        236

$ (cd scripts/docs-validation && npm run validate:ts) | tail -6
TypeScript:
  ✅ 191 files passed

```

</details>

### DIFF
--- a/scripts/docs-validation/extract.ts
+++ b/scripts/docs-validation/extract.ts
@@ -61,6 +61,7 @@ function parseMarkdownCodeBlocks(
   let skipNext = false;
   let wrapAsync = false;
   let inHiddenBlock = false;
+  let inUnvalidatedFence = false;
 
   for (let i = 0; i < lines.length; i++) {
     const line = lines[i];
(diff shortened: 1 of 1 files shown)


SteveSandersonMS (approved):
Manually reproduced the bug this fixes: reverted extract.ts to main and re-ran the extraction/validation pipeline — the pending-skip directive leaked past unvalidated fences (e.g. rust), silently dropping examples from type-checking (confirmed against the plugin-directories.md example added in #2581, which depends on this fix). The tracked-fence fix is minimal and correctly scoped. Approving.

SteveSandersonMS (approved):
Confirmed the fix: reproduced the skip-leak bug by diffing extraction output on main vs. this branch (typescript 189→191, csharp 56→57, matching the PR body), and ran real validation (`validate:ts`: 191/191 pass, `validate:cs`: 57/57 pass) to confirm the newly-included examples actually type-check. Minimal, single-file fix that reuses the existing hidden-block-skip-reset pattern. Rebased onto latest main to pick up the unrelated SourceLink vulnerability fix (#2587) that was blocking the C# check.

PR #2581, opened 2026-09-08
Document per-session plugin directories

1fanwang:
Stacked on https://github.com/github/copilot-sdk/pull/2580, which the added example needs in order to be type-checked. Once that merges this reduces to the documentation commit.

## Why

An SDK application wants one session to load a code-review plugin and another to load nothing, or it connects to a runtime it did not spawn. The reader opens the plugin directories guide and finds two ways to load a plugin: the `--plugin-dir` launch argument, and the trusted host-bundled startup option for plugins the host itself ships. Neither fits. The guide says plainly that loading a plugin directory "makes its extensions visible to every session created by the client", and that the SDK does not forward `--plugin-dir` to runtimes it did not spawn.

So the reader concludes per-session plugin sets are not supported, and reaches for a second client or a wrapper process.

They are supported. The session config takes `pluginDirectories`, it is typed in all six language bindings, and the SDK forwards it on both create and resume. Because it travels in the JSON-RPC payload rather than as a process argument, it also reaches an external runtime. It simply had no entry in its own guide.

After this change the reader finds it beside the two mechanisms they were already comparing.

## What changed

A section covering the per-session option: what it does, how paths resolve, how it interacts with config discovery and precedence, an example, and the spelling in each language.

## Testing Done

Documentation only, no code paths touched.

The option was absent from the guide before this change and present after:

```
$ git grep -c pluginDirectories upstream/main -- docs/features/plugin-directories.md
(no output, exit status 1)

$ git grep -c pluginDirectories HEAD -- docs/features/plugin-directories.md
HEAD:docs/features/plugin-directories.md:3
```

The added example is compiled against the SDK types by the docs validation, so a wrong option name or type fails the build.

<details><summary>Raw logs: docs validation</summary>

```
$ (cd scripts/docs-validation && npm run validate:ts) | tail -6
TypeScript:
  ✅ 192 files passed

```

</details>

### DIFF
--- a/docs/features/plugin-directories.md
+++ b/docs/features/plugin-directories.md
@@ -240,6 +240,51 @@ let client = Client::start(
 
 > The example above uses an stdio runtime connection — the default when the SDK bundles the CLI. If you connect to an external runtime via a URL (`forUri` / `ForUri`), pass `--plugin-dir` to the long-running CLI server when you start it; the SDK does not forward `--plugin-dir` to runtimes it didn't spawn.
 
+## Per-session plugin directories
+
+`--plugin-dir` is a launch argument, so it fixes one plugin set for the CLI process and every session created against it. When sessions need different plugin sets, or when the SDK is connected to a runtime it did not spawn, pass the directories on the session config instead. They travel in the `session.create` and `session.resume` payloads over JSON-RPC rather than as process arguments, so they reach an external runtime the same way the startup option does.
+
+<!-- docs-validate: hidden -->
+```typescript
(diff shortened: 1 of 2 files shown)


SteveSandersonMS (approved):
Confirmed `pluginDirectories` is a real, already-shipped session option typed and forwarded across all six SDKs (create + resume), and manually reproduced the `extract.ts` skip-leak bug that #2580 fixes — this PR's own new example silently escaped type-checking without it. Docs are accurate and consistent with existing option naming. Approving; will merge once #2580 lands and CI is green.

PR #2585, opened 2026-09-09
sdk: make client startup single-flight

DonJayamanne:
## Summary
- make Node.js `CopilotClient.start()` share one in-flight CLI startup
- serialize Python startup attempts and clean partial transports after failure or cancellation before retrying
- atomically publish one .NET connection task, clear failed or cancelled attempts before waking waiters, and clean a partially started FFI host
- add deterministic concurrency, retry, and cancellation regression coverage across the affected SDKs

## Why
Concurrent lazy-start callers could pass the initial state check before startup was published, causing duplicate CLI processes or connections. The language-specific guards now establish one startup owner while preserving safe retry behavior after failure.

## Validation
- Node.js focused startup tests: 2 passed
- Python client unit tests: 152 passed
- Python startup-error retry E2E: passed
- Python Ruff check/format and `ty check copilot`: passed
- .NET startup regressions: 2 passed
- .NET `ClientSessionLifetimeTests`: 125 passed
- .NET SDK build for `net8.0`, `net10.0`, and `netstandard2.0`: passed locally with NuGet audit disabled because of the upstream advisory below
- .NET formatter verification scoped to changed files: passed
- `git diff --check`: passed

## CI note
The current .NET jobs stop during restore, before compiling this diff, because the existing `Microsoft.SourceLink.GitHub 10.0.102` dependency now triggers GHSA-23fw-v26w-5fgq. This PR does not update that unrelated dependency or suppress repository CI audits.

### DIFF
--- a/dotnet/src/Client.cs
+++ b/dotnet/src/Client.cs
@@ -365,7 +365,56 @@ public sealed partial class CopilotClient : IDisposable, IAsyncDisposable
     /// </example>
     public Task StartAsync(CancellationToken cancellationToken = default)
     {
-        return _connectionTask ??= StartCoreAsync(cancellationToken);
+        var connectionTask = Volatile.Read(ref _connectionTask);
+        if (connectionTask is not null)
+        {
+            return connectionTask;
+        }
+
+        var completion = new TaskCompletionSource<Connection>(
+            TaskCreationOptions.RunContinuationsAsynchronously);
+        // Publish before startup begins because StartCoreAsync executes synchronously
+        // until its first incomplete await and can re-enter user-provided callbacks.
+        connectionTask = Interlocked.CompareExchange(
+            ref _connectionTask,
+            completion.Task,
+            null);
+        if (connectionTask is not null)
+        {
+            return connectionTask;
+        }
+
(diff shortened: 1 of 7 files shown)


DonJayamanne:
Addressed the cross-SDK consistency review in `cf091b903`:

- Python now serializes startup and cleans partial resources on failure and cancellation before retry.
- .NET now atomically publishes one shared startup task and clears failed/cancelled attempts before notifying waiters.
- Added deterministic concurrency, retry, and cancellation regressions; no changes were needed for Go, Java, or Rust.

DonJayamanne:
Fixed the Python matrix regression in `a6d7ad37b`. All eight jobs had the same single failing E2E: it expected a stale pipe after failed startup, but startup is now intentionally retryable and correctly returns the original invalid-argument stderr again. The updated expectation passes locally, along with 152 Python client tests, Ruff, and `ty`.

The .NET jobs are a separate pre-compilation restore failure from the existing `Microsoft.SourceLink.GitHub 10.0.102` dependency and newly published GHSA-23fw-v26w-5fgq; no unrelated dependency or audit suppression was added here.

PR #2586, opened 2026-09-09
Align Java formatting with shared just recipes

roji:
Java is the only SDK with a repository-managed pre-commit formatting hook; the other SDKs use explicit local commands backed by CI. The hook also checks working-tree contents rather than the staged snapshot, so it is not reliable enforcement of what gets committed.

Replace the hook and its automatic installation with `just format-java` (Spotless) and `just lint-java` (Spotless and Checkstyle), and include Java in the aggregate `just format` and `just lint` recipes. Update the development documentation with these commands and their Maven equivalents.

Existing Spotless and Checkstyle CI enforcement remains unchanged. Spotless continues to run explicitly in CI because `mvn verify` alone does not check formatting.


edburns (approved):
I approve this and will also state so in the PR.

Issue #1185, opened 2026-05-01
Support structured outputs (JSON schema) per message, passed through to provider


Halcyonhal9:
## Summary

The Copilot SDK should expose first-class support for **structured outputs** (JSON schema–constrained responses) when targeting model providers that natively support the feature — OpenAI (`response_format: { type: "json_schema", ... }`) and Anthropic (structured output / tool-schema). The schema must be settable **per message** and passed through unchanged to the provider.

## Motivation

The primary use case is **deterministic agent flows that operate on the output of Copilot Studio agents**.

In Copilot Studio, an agent turn is frequently a step inside a larger orchestrated agent flow — its output is not just rendered to a user, it's fed into:

- Agent flows that expect specific fields in the output
- Routing/branching logic that switches on a category, intent, or decision

These consumers require **stable, schema-validated JSON** from the agent turn. Today the SDK only emits free-form assistant text, which forces every agent flow to either:

1. Post-parse model prose with regex / JSON-extraction heuristics — brittle and silently breaks when the model rephrases.
2. Coerce JSON via a fake single-tool tool-call — adds a round-trip per turn, pollutes tool-use telemetry, and confuses `preToolUse` hooks and authoring UX.
3. Rewrite the outgoing HTTP body in an LLM interceptor to inject `response_format` — fights the SDK's own retry/streaming logic and is unsupported.

Both OpenAI and Anthropic already accept a JSON schema directly on the request. The SDK is the only layer blocking Copilot Studio from getting deterministic agent output end-to-end.

## Proposed API

Structured output must be **per message**, because different turns in an agent flow need different schemas (classify → plan → extract → summarize).

```ts
await session.sendMessage({
  content: "Classify this support ticket.",
  responseFormat: {
    type: "json_schema",
    schema: {
      name: "TicketClassification",
      strict: true,
      schema: {
        type: "object",
        properties: {
          category: { type: "string", enum: ["billing", "technical", "other"] },
          priority: { type: "string", enum: ["low", "medium", "high"] },
          summary: { type: "string" }
        },
        required: ["category", "priority", "summary"],
        additionalProperties: false
      }
    }
  }
});
```

### Requirements

1. **Per-message setting** — `responseFormat` accepted on `sendMessage` / equivalent. Optional; absent ⇒ current behavior.
2. **Schema passthrough** — Forwarded verbatim to the provider request:
   - OpenAI BYOM: `response_format: { type: "json_schema", json_schema: <schema> }`
[...]

PR #2590, opened 2026-09-09
feat: Add typed structured outputs for all six SDKs

SteveSandersonMS:
## Summary

Adds **provider-native structured output** to all six SDKs: provide a JSON Schema, or use an idiomatic typed helper to infer the schema and parse the result.

Addresses #1185. The runtime support shipped in CLI **1.0.86-0** via [runtime #19652](https://github.com/github/copilot-agent-runtime/pull/19652). This PR uses the generated contracts already on `main`; it contains no RPC codegen changes.

### How it works

- A schema belongs to one submitted run, including its tool calls, steering, and stop-hook corrections. Independent sends do not inherit it.
- Typed helpers wait for non-autopilot session idle and select the last matching root assistant message using `originatingMessageId`. This is not a promise of one assistant message per user message.
- Streaming stays text. Immediate steering must omit the schema. Timeouts/cancellation stop waiting, not agent work.
- Provider schema restrictions apply. Node/Zod and Python/Pydantic validate results; other languages use their normal JSON deserialization, not full JSON Schema validation.

Each language has nine E2E scenarios. All six reuse **the same eight recordings from actual CAPI calls**; admission-rejection cases make no provider calls. Coverage includes tools, streaming, concurrent schemas, steering, corrections, raw/batch RPCs, and schema clearing.

## TypeScript / Node.js

Pass a Zod schema as the second argument to infer, parse, and validate the result.

```ts
import { z } from "zod";

const answerSchema = z.object({ value: z.number().int() });
const answer = await session.sendAndWait("What is 19 + 23?", answerSchema);
console.log(answer.value); // inferred as number
```

## Python

Use a Pydantic model, as with custom-tool parameters.

```python
from pydantic import BaseModel, ConfigDict

class Answer(BaseModel):
    model_config = ConfigDict(extra="forbid")
    value: int

answer = await session.send_and_wait_typed("What is 19 + 23?", Answer)
print(answer.value)
```

Requires **Pydantic 2.11+**. Validation uses the schema's aliases consistently, including in nested models.

## Go

Use the package-level generic helper; schema inference reuses the existing custom-tool generator.

```go
type Answer struct {
    Value int `json:"value"`
}

answer, err := copilot.SendAndWait[Answer](ctx, session, copilot.MessageOptions{
    Prompt: "What is 19 + 23?",
})
if err != nil {
    return err
}
fmt.Println(answer.Value)
```

## C#

Use the generic overload with the same schema inference as custom tools.

```csharp
var answer = await session.SendAndWaitAsync<Answer>("What is 19 + 23?");
Console.WriteLine(answer.Value);

public sealed record Answer(int Value);
```

Optional `serializerOptions` control both schema inference and deserialization without being mutated. For Native AOT or reflection-disabled applications, supply source-generated options, such as `AnswerJsonContext.Default.Options`.

## Java

Annotate the result type and enable `CopilotResponseProcessor`, using the existing custom-tool annotation-processing approach and schema generator.

```java
@CopilotResponse
public record Answer(int value) {}

Answer answer = session.sendAndWait("What is 19 + 23?", Answer.class).get();
System.out.println(answer.value());
```

No new schema library. Custom Jackson mappings and recursive types require an explicit schema.

## Rust

Enable the existing `derive` feature and use Serde/schemars, as with typed custom tools.

```rust
#[derive(serde::Deserialize, schemars::JsonSchema)]
#[serde(deny_unknown_fields)]
struct Answer {
    value: i32,
}

let answer: Answer = session.send_and_wait_typed("What is 19 + 23?").await?;
println!("{}", answer.value);
```

Deserialization preserves `i128`/`u128` precision and rejects top-level JSON null.

## `send` vs. `sendAndWait`

**`send` returns a user-message ID.** Capture assistant messages, wait for `session.idle`, then parse the last message whose `originatingMessageId` matches that ID, if present. **Typed `sendAndWait` does this for you.**

These happy-path examples assume one send on an initially idle, non-autopilot session. They reuse the types above and omit session setup, routine imports, and error handling. The callback examples keep messages by origin because events can arrive before `send` returns its ID; Rust's subscription already buffers events.

Here `prompt` is `"What is 19 + 23?"`. For Go, C#, and Java, `schema` is this JSON Schema represented as a `map[string]any`, `JsonElement`, or `Map<String, Object>`, respectively:

```json
{"type":"object","properties":{"value":{"type":"integer"}},"required":["value"],"additionalProperties":false}
```

### TypeScript / Node.js

**Wait and parse:** `const answer = await session.sendAndWait(prompt, answerSchema);`

**Send and listen:**

```ts
const messages = new Map<string, string>();
let onIdle!: () => void;
const idle = new Promise<void>(resolve => { onIdle = resolve; });
const unsubscribe = session.on(event => {
    if (event.agentId) return;
    if (event.type === "assistant.message" && event.data.originatingMessageId)
        messages.set(event.data.originatingMessageId, event.data.content);
    if (event.type === "session.idle") onIdle();
});

const id = await session.send({ prompt, responseSchema: answerSchema });
await idle;
unsubscribe();
const content = messages.get(id);
if (content) {
    const answer = answerSchema.parse(JSON.parse(content));
    console.log(answer.value);
}
```

### Python

**Wait and parse:** `answer = await session.send_and_wait_typed(prompt, Answer)`

**Send and listen:**

```python
import asyncio
from copilot.session_events import AssistantMessageData, SessionIdleData

messages = {}
idle = asyncio.Event()

def on_event(event):
    if event.agent_id:
        return
    if isinstance(event.data, AssistantMessageData):
        messages[event.data.originating_message_id] = event.data.content
    if isinstance(event.data, SessionIdleData):
        idle.set()

unsubscribe = session.on(on_event)
message_id = await session.send(prompt, response_schema=Answer)
await idle.wait()
unsubscribe()
if content := messages.get(message_id):
    answer = Answer.model_validate_json(content)
    print(answer.value)
```

### Go

**Wait and parse:** `answer, err := copilot.SendAndWait[Answer](ctx, session, copilot.MessageOptions{Prompt: prompt})`

**Send and listen:**

```go
var messages sync.Map
idle := make(chan struct{}, 1)
unsubscribe := session.On(func(event copilot.SessionEvent) {
    if event.AgentID != nil { return }
    switch data := event.Data.(type) {
    case *copilot.AssistantMessageData:
        if data.OriginatingMessageID != nil {
            messages.Store(*data.OriginatingMessageID, data.Content)
        }
    case *copilot.SessionIdleData:
        idle <- struct{}{}
    }
})
defer unsubscribe()

id, err := session.Send(ctx, copilot.MessageOptions{Prompt: prompt, ResponseSchema: schema})
if err != nil { return err }
<-idle
if content, ok := messages.Load(id); ok {
    var answer Answer
    if err := json.Unmarshal([]byte(content.(string)), &answer); err != nil { return err }
    fmt.Println(answer.Value)
}
```

### C#

**Wait and parse:** `var answer = await session.SendAndWaitAsync<Answer>(prompt);`

**Send and listen:**

```csharp
var messages = new ConcurrentDictionary<string, string>();
var idle = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
using var subscription = session.On<SessionEvent>(e =>
{
    if (e.AgentId is not null) return;
    if (e is AssistantMessageEvent message && message.Data.OriginatingMessageId is { } origin)
        messages[origin] = message.Data.Content;
    if (e is SessionIdleEvent) idle.TrySetResult();
});

var id = await session.SendAsync(new MessageOptions { Prompt = prompt, ResponseSchema = schema });
await idle.Task;
if (messages.TryGetValue(id, out var content))
{
    var answer = JsonSerializer.Deserialize<Answer>(content, new JsonSerializerOptions(JsonSerializerDefaults.Web));
    Console.WriteLine(answer!.Value);
}
```

### Java

**Wait and parse:** `Answer answer = session.sendAndWait(prompt, Answer.class).get();`

**Send and listen:**

```java
var messages = new ConcurrentHashMap<String, String>();
var idle = new CompletableFuture<Void>();
try (var subscription = session.on(event -> {
    if (event.getAgentId() != null) return;
    if (event instanceof AssistantMessageEvent message && message.getData().originatingMessageId() != null)
        messages.put(message.getData().originatingMessageId(), message.getData().content());
    if (event instanceof SessionIdleEvent) idle.complete(null);
})) {
    String id = session.send(new MessageOptions().setPrompt(prompt).setResponseSchema(schema)).get();
    idle.get();
    String content = messages.get(id);
    if (content != null) {
        Answer answer = new ObjectMapper().readValue(content, Answer.class);
        System.out.println(answer.value());
    }
}
```

### Rust

**Wait and parse:** `let answer: Answer = session.send_and_wait_typed(prompt).await?;`

**Send and listen:**

```rust
use github_copilot_sdk::{MessageOptions, tool::schema_for};

let mut events = session.subscribe();
let id = session.send(MessageOptions::new(prompt).with_response_schema(schema_for::<Answer>())).await?;
let mut content = None;
loop {
    let event = events.recv().await?;
    if event.agent_id.is_some() { continue; }
    if event.event_type == "assistant.message" && event.data["originatingMessageId"] == id {
        content = event.data["content"].as_str().map(str::to_owned);
    }
    if event.event_type == "session.idle" {
        if let Some(text) = content {
            let answer: Answer = serde_json::from_str(&text)?;
            println!("{}", answer.value);
        }
        break;
    }
}
drop(events);
```


nodejs/src/types.ts
```diff
@@ -3395,6 +3403,20 @@ export interface MessageOptions {
...
+     * schema and origin, even when promoted to a follow-up after the model request finishes.
+     * Specifying a schema with mode "immediate" is rejected, even while idle.
+     * This is not a persisted session default and does not survive a context reset.
+     *
+     * sendAndWait still returns an assistant message event. For a typed result, pass a
+     * Zod-compatible schema as sendAndWait's second argument instead.
+     * Streaming events remain text and may include intermediate messages.
+     * Use rpc.send's responseFormat for provider-specific name, description and strict options.
+     */
+    responseSchema?: ZodSchema | Record<string, unknown>;
```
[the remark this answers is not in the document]
SteveSandersonMS: Implemented parity in 04ecfa0b: Python/Pydantic, Go/jsonschema-go, Java using the existing custom-tool schema generator, and Rust/schemars now expose raw-schema options and typed waits. Each has correlation/lifecycle coverage and real-runtime E2Es using the existing provider captures; Rust CI also enables derive so that coverage runs.

python/test_structured_output.py
```diff
@@ -0,0 +1,253 @@
...
+    if kind == "cancel":
+        waiting.cancel()
+        error = asyncio.CancelledError
+    elif kind == "disconnect":
+        session._mark_disconnected()
+        error = RuntimeError
+    else:
+        error = TimeoutError
+    with pytest.raises(error):
+        await waiting
```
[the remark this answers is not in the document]
SteveSandersonMS: This await is intentional: it waits for the asyncio task and propagates its timeout, cancellation, or disconnection exception into pytest.raises. Removing it would stop testing those failure paths and make pytest.raises fail because no exception was raised. Keeping the awaited task preserves the lifecycle regression coverage; this is a no-effect false positive.

dotnet/test/E2E/StructuredOutputE2ETests.cs
```diff
@@ -0,0 +1,430 @@
...
+    {
+        var environment = Ctx.GetEnvironment();
+        environment["COPILOT_CLI_ENABLED_FEATURE_FLAGS"] = "HYDRAFUSION,HYDRAFUSION_ROLLOUT";
+        await using var client = Ctx.CreateClient(environment: environment);
+        foreach (var model in new[] { "gpt-4.1", "hydrafusion" })
+        {
+            var config = StructuredSessionConfig();
+            config.Model = model;
+            config.OnPermissionRequest = PermissionHandler.ApproveAll;
+            await using var session = await client.CreateSessionAsync(config);
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in 70d73ccd: session creation now goes through Ctx.CreateSessionAsync(client, config), preserving the custom client environment while applying the selected provider. The rejection test passes against runtime 1.0.86-0 with all four backend configurations.

scripts/codegen/csharp.test.ts
```diff
@@ -0,0 +1,64 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------------------------------------------*/
+
+import assert from "node:assert/strict";
+import { test } from "node:test";
+import type { JSONSchema7 } from "json-schema";
+import { generateRpcCode } from "./csharp.js";
+
+for (const keyword of ["anyOf", "oneOf"] as const) {
+    test(`C# RPC preserves named single-variant ${keyword} objects`, () => {
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in 70d73ccd: scripts/codegen now has an npm test command using its existing tsx dependency, and Codegen Check invokes it. Both single-variant union regression cases pass through that command.

nodejs/src/session.ts
```diff
@@ -782,17 +801,52 @@ export class CopilotSession {
...
+        if (schemaOrTimeout !== undefined && typeof schemaOrTimeout !== "number" && !typedSchema) {
+            throw new TypeError(
+                "The second argument must be a timeout or a schema with toJSONSchema() and parse(). " +
+                    "Pass raw JSON Schema in options.responseSchema instead."
+            );
+        }
+        const effectiveTimeout =
+            (typeof schemaOrTimeout === "number" ? schemaOrTimeout : timeout) ?? 60_000;
+
+        if (typedSchema && options.responseSchema) {
```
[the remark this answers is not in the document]
SteveSandersonMS: Fixed in 77bee95d: the Node typed overload now rejects immediate steering before subscribing or sending JSON-RPC, matching the other five SDKs. Added a regression case that asserts both the error and that sendRequest was never called. Unformatted steering and raw-schema behavior are unchanged.

SteveSandersonMS:
On the coverage question in https://github.com/github/copilot-sdk/pull/2590#issuecomment-5728838975: yes, the 32 MiB ceiling and unsupported-provider rejection are shared runtime validation, deliberately covered through Node/.NET E2Es rather than duplicating those cases in all six SDKs. The other SDKs exercise raw-schema forwarding and typed results through the same runtime with shared provider recordings, plus language-specific correlation, correction, and failure-path unit coverage. The Node immediate-mode discrepancy from the earlier review was fixed in 77bee95d.

dotnet/src/Session.StructuredOutput.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            throw new ArgumentException("The typed overload infers its response schema. Use the untyped overload for an explicit response schema.", nameof(options));
+        }
+        if (options.Mode == "immediate")
+        {
+            throw new ArgumentException("Structured output cannot be requested on an immediate steering message.", nameof(options));
+        }
+
+        serializerOptions ??= AIJsonUtilities.DefaultOptions;
+        var typeInfo = (JsonTypeInfo<TResult>)serializerOptions.GetTypeInfo(typeof(TResult));
```
SteveSandersonMS, line 80: **[P2] Support fresh custom serializer options**

Passing ordinary, freshly constructed serializer options fails before sending, even with reflection enabled:

```csharp
var options = new JsonSerializerOptions {
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase
};
await session.SendAndWaitAsync<Answer>("What is 19 + 23?", options);
```

`GetTypeInfo` throws `NotSupportedException` because the options' resolver has not been initialized. The identical options work after an ordinary `JsonSerializer.Deserialize` call initializes them; omitted options and source-generated options also work. This was also reproduced against an actual client, with zero admitted user messages.

Please resolve the serialization metadata consistently with normal System.Text.Json behavior, while respecting reflection-disabled configurations and avoiding mutation of the caller's options.

python/copilot/session.py
```diff
@@ -1786,6 +1802,16 @@ async def send(
             params["requestHeaders"] = request_headers
         if display_prompt is not None:
             params["displayPrompt"] = display_prompt
+        if response_schema is not None:
+            schema = (
+                response_schema.model_json_schema()
```
SteveSandersonMS, line 1807: **[P2] Keep inferred aliases consistent with result validation**

Schema generation defaults to alias names, but the subsequent `model_validate_json` in the new typed helper respects the model's validation configuration. These can produce incompatible contracts:

```python
class Answer(BaseModel):
    model_config = ConfigDict(
        extra="forbid",
        validate_by_alias=False,
        validate_by_name=True,
    )
    value: int = Field(alias="answer")
```

The generated schema requires `answer` and forbids `value`. A real remote call returned the compliant `{"answer":42}`, but `send_and_wait_typed` rejected it because `value` was missing.

Please make schema generation and deserialization use consistent alias semantics, including nested models.

rust/src/session.rs
```diff
@@ -648,6 +743,86 @@ impl Session {
...
+        opts.response_schema = Some(crate::tool::schema_for::<T>());
+        let event = self.send_and_wait_structured(opts).await?;
+        let data: crate::session_events::AssistantMessageData = serde_json::from_value(event.data)?;
+        let value: Value = serde_json::from_str(&data.content)?;
+        if value.is_null() {
+            return Err(structured_output_error(
+                "structured response was JSON null, not a result",
+            ));
+        }
+        Ok(serde_json::from_value(value)?)
```
SteveSandersonMS: **[P2] Deserialize typed results without losing wide integers**

Converting the response through `serde_json::Value` breaks valid `i128`/`u128` results. Without `arbitrary_precision`, integers beyond `u64::MAX` become floating-point values in `Value`.

For a derived response containing `amount: i128`, a real remote call returned:

```json
{"amount":18446744073709551616}
```

`serde_json::from_str::<Answer>(&content)` succeeds on that exact response, but this helper fails with `invalid type: floating point 1.8446744073709552e+19, expected i128`.

Please retain the top-level-null rejection while deserializing `T` directly from the original JSON text.

SteveSandersonMS (commented):
Three issues in the newly introduced typed-output paths, reproduced during the console-app review. Pre-existing issues are excluded.

python/copilot/session.py
SteveSandersonMS: Fixed in ac02e37d. Typed validation now explicitly uses by_alias=True and by_name=False, matching model_json_schema() aliases throughout nested models and lists without changing model configuration. Added a nested-model regression and exercised disabled alias validation against the existing shared CAPI recording. The existing Pydantic dependency minimum is now 2.11 because that release introduced these per-call controls; the targeted unit/E2E suite passes on both the locked version and 2.11.0.

dotnet/src/Session.StructuredOutput.cs
SteveSandersonMS: Fixed in ac02e37d. Mutable serializer options are copied before metadata resolution, preserving the caller’s options and resolver. A default reflection resolver is supplied only when System.Text.Json reflection is enabled; supplied source-generated resolvers remain supported with reflection disabled. Added fresh-options regressions for both configurations and an explicit reflection-enabled CI step; the shared real-runtime E2E also uses fresh options. All 39 structured-output cases pass with reflection both enabled and disabled, and all target frameworks build without warnings.

rust/src/session.rs
SteveSandersonMS: Fixed in ac02e37d. The helper now deserializes Option<T> directly from the original JSON text, then rejects None, avoiding the lossy Value intermediate while retaining top-level-null rejection even when T itself is optional. Regressions cover i128::MIN, i128::MAX, 18446744073709551616, u128::MAX, and whitespace-wrapped null. Structured-output unit and shared runtime E2E suites pass on both transports.

PR #2595, opened 2026-09-09
docs: add in-process runtime guide

patniko:
## Summary

- explain when to host the Copilot runtime in process
- document configuration for all six SDKs
- cover native runtime resolution, lifecycle behavior, and experimental limitations

## Validation

- `git diff --check`
- docs validation passed for TypeScript, Python, Go, and C#
- Java docs validation was not run because Maven is unavailable in the local environment

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -25,6 +25,7 @@ How to configure and deploy the SDK for your use case.
 * [Default Setup (Bundled CLI)](./setup/bundled-cli.md): the SDK includes the CLI automatically
 * [Local CLI](./setup/local-cli.md): use your own CLI binary or running instance
 * [Backend Services](./setup/backend-services.md): server-side with headless CLI over TCP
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
I completed this PR by fixing the two issues the review had flagged:
- Added the missing `@AllowCopilotExperimental` opt-in (with import) to the Java example, matching the pattern documented in java/README.md.
- Added a row for in-process hosting to the decision matrix in choosing-a-setup-path.md so the cross-link isn't dangling.

Thanks for writing this guide — in-process hosting really did need a top-level doc, and everything else checked out against the current SDK implementations.

SteveSandersonMS (approved):
Verified the in-process hosting API surface against all six SDKs' source (TS, Python, Go, .NET, Rust, Java) and ran the TypeScript example live. Fixed the Java experimental opt-in and added the decision-matrix cross-link in a follow-up commit. LGTM.

PR #2596, opened 2026-09-09
Add seven-day Dependabot cooldowns for npm, PyPI, and NuGet

MackinnonBuck:
New dependency releases can be unavailable in downstream registries during their seven-day quarantine, preventing contributors from installing the SDK or its development dependencies.

Add `cooldown.default-days: 7` to all three npm entries and the PyPI and NuGet entries, with `@github/*` excluded from the npm cooldowns. Existing weekly schedules, grouping, ignore rules, and other ecosystems remain unchanged.

Dependabot security updates bypass cooldowns. Copilot CLI version updates are managed separately and are unaffected.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -3,6 +3,7 @@ multi-ecosystem-groups:
   all:
     schedule:
       interval: 'weekly'
+# npm, PyPI, and NuGet cooldowns accommodate downstream registry quarantine.
 updates:
   - package-ecosystem: 'github-actions'
     directory: '/'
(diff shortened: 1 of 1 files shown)


SteveSandersonMS (approved):
Thanks for this — the cooldown config looks correct and I confirmed `cooldown`/`default-days`/`exclude` are valid, currently-supported Dependabot options for npm, pip, and nuget. One small question for my own understanding: Dependabot's cooldown option is also supported for `gomod` and `maven`, but this PR leaves Go and Java out. Is that because those ecosystems don't route through the same quarantined mirror, or just out of scope for now? Either way this looks good to merge.

PR #2598, opened 2026-09-09
Stabilize Java OAuth lifecycle E2E test

roji:
The Java OAuth lifecycle E2E test intermittently misses the initial `mcp.oauth_required` event when the test MCP server challenges authentication before `createSession()` finishes registering event interest.

Add an opt-in first-challenge gate to the test server and release it from the lifecycle scenario only after session creation completes. Existing test-server consumers retain their normal behavior, while the lifecycle test still covers initial, refresh, upscope, and reauthentication token flows.

### DIFF
--- a/java/sdk/src/test/java/com/github/copilot/McpOAuthE2ETest.java
+++ b/java/sdk/src/test/java/com/github/copilot/McpOAuthE2ETest.java
@@ -114,9 +114,10 @@ public class McpOAuthE2ETest {
 
     @Test
     void testShouldRequestReplacementTokensAcrossMcpOauthLifecycle() throws Exception {
-        try (var oauthServer = OAuthMcpServer.start(ctx.getRepoRoot())) {
+        try (var oauthServer = OAuthMcpServer.start(ctx.getRepoRoot(), true)) {
             var serverName = "oauth-lifecycle-mcp";
             var observedReasons = new CopyOnWriteArrayList<McpOauthRequestReason>();
+            var observedRequest = new AtomicReference<McpAuthRequest>();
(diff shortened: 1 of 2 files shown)


java/sdk/src/test/java/com/github/copilot/McpOAuthE2ETest.java
```diff
@@ -336,6 +345,15 @@ static OAuthMcpServer start(Path repoRoot) throws Exception {
             throw new AssertionError("Timed out waiting for OAuth MCP server: " + stderr);
         }
 
+        void releaseInitialChallenge() throws Exception {
+            var client = HttpClient.newHttpClient();
+            var response = client.send(
+                    HttpRequest.newBuilder(URI.create(url + "/__release-initial-challenge"))
+                            .timeout(Duration.ofSeconds(10)).POST(HttpRequest.BodyPublishers.noBody()).build(),
```
stephentoub, line 352: Is 10 seconds still going to be flaky?

stephentoub (approved):

java/sdk/src/test/java/com/github/copilot/McpOAuthE2ETest.java
roji: No. The 10 seconds are timeouts, not delays or polling intervals. In the successful path, `createSession()` completes its `mcp.oauth_required` interest registration, the test immediately releases the local gate, and the server immediately returns the held initial challenge.

The ordering is now explicit rather than scheduler-dependent. The timeout only produces a clear failure if the local control request or test server hangs.

> [!NOTE]

PR #2600, opened 2026-09-10
Expose versioned session activity in the Rust SDK

ellismg:
## Summary

- pin generated protocol contracts to runtime commit `fe8a752a5442916234d59611eda30f68ba0821f9` until a containing Copilot CLI package is released
- regenerate the standard TypeScript, .NET, Python, Go, Rust, and Java contract outputs without hand-editing generated files
- add validated Rust access for `session.metadata.activity` and `session.activity_changed` using one shared `SessionActivitySnapshot`
- add explicit supported/legacy/method-unavailable/unknown-version handling and malformed-v1 protocol errors
- add a connection-scoped epoch/revision reducer that rejects superseded connections and delayed stale snapshots
- document subscribe-before-query ordering, activity semantics, process independence, and scoped cancellation APIs

## Compatibility

The executable runtime remains pinned to Copilot CLI `1.0.83`. The new activity contract requires runtime commit `fe8a752a5442916234d59611eda30f68ba0821f9` until a containing package release exists. Consumers should pin this SDK commit (`0a0417f118c07db242e850faed0f6526928ef81e`) until the next SDK release is published.

Capability detection is structural: only `contractVersion == 1` is supported. Missing method, legacy two-field responses, and unknown versions are unsupported/unknown rather than synthesized idle state.

The authoritative runtime head advanced from `3225633fbdface9b6cfb0997aea1356d66552d31` after CI fixes. Both `api.schema.json` and `session-events.schema.json` are byte-for-byte identical between those commits, so the follow-up only updates provenance pins; no SDK API or generated contract changed.

## Validation

- Rust nightly format and Clippy with all features/targets
- Rust activity unit tests, event-subscription delivery test, legacy-runtime E2E query test, library suite, examples, and docs
- shared multi-language code generation from the updated runtime head
- Java schema-fetcher/codegen tests using the updated runtime head
- Node typecheck and focused generated/codegen tests
- Python unit suite (`473 passed`)

Broad existing E2E suites were also exercised; environment/runtime setup failures outside this change remained in Node/.NET, while the focused activity and generated-contract validation passed.


SteveSandersonMS:
Thanks for the detailed work here — the capability-detection shape is sensible, and session activity is a legitimate SDK capability once the runtime contract is real. This PR isn't ready to merge because it is still pinned to an unreleased runtime contract (`fe8a752a...`) rather than a shipped bundled CLI/runtime schema, and current CI remains red.

For tracking, I've moved this to draft - please mark as ready to review when appropriate.

Please bring this back once:

1. The runtime contract has shipped in a Copilot CLI/runtime release and the generated files are regenerated from that shipped source rather than a private runtime commit pin.
2. The generated outputs are actually current and CI is green across affected SDKs, including Java compile/tests and Go formatting/codegen checks.
3. The new Rust reducer has coverage and a fix for stale query results from an older epoch arriving after a newer event.
4. Go's `session.activity_changed` event data preserves the activity payload instead of decoding into an empty struct.
5. Equivalent high-level activity APIs are added across the SDK languages. This is language-neutral runtime state, so it should not be Rust-only without a real product/technical reason, and no such reason is established here.
6. There is real-runtime/E2E validation showing the activity query/event behavior works against the shipped runtime.

Happy to re-review once the runtime ships and these are addressed.

PR #2616, opened 2026-09-11
Expose local cross-session discovery and delivery

dfrysinger:
## Summary

- add source-bound Node and Rust APIs for listing active local messageable sessions
- add exact-target cross-session delivery with typed refused, not-delivered, and ambiguous terminal errors
- keep authenticated recipient admission inside the runtime rather than exposing caller-controlled provenance through the SDK
- ensure generated session-scoped TypeScript calls always overwrite any caller-provided source session ID

## Runtime dependency

This SDK surface consumes the local transport, provenance, target-claim, and recipient-admission contract in github/copilot-agent-runtime#19586. The runtime owns wire security and recipient admission. The SDK exposes only source-bound discovery and delivery so CLI and App consumers do not implement separate local protocols. Same-computer discovery and delivery do not require remote control or remote export.

The unreleased discovery and delivery wire types live in the hand-written SDK facades until the runtime schema reaches the SDK pinned CLI release. Generated files remain reproducible from that pinned release.

## Validation

- Node typecheck and formatting
- 70 focused Node tests for discovery, delivery, event types, and source identity
- Rust compile check
- 2 Rust discovery tests
- 4 Rust delivery tests
- deterministic pinned-schema generation and whitespace checks


dfrysinger:
The official GitHub Copilot App consumer is now open: https://github.com/github/github-app/pull/15507\n\nThese PRs align Runtime, SDK, and App on one public local cross-session messaging protocol. The App vendors SDK commit d8aaeed60fee20ec1f33f18d36d00a7e8936d134 and proves bidirectional local App/CLI delivery without remote control or remote session export.

SteveSandersonMS:
Thanks for iterating on this. Three things to resolve before this is ready for review:

1. Capability gating: cross-session-messaging is hardcoded to `client_kind == "cli"` in the runtime rather than using the existing `enable_<feature>` + `ClientMode`-default pattern already used for `mcp-apps`/`session-store` (default off in `Empty` mode, on in `CopilotCli` mode, always caller-overridable). Please make this an explicit opt-in/opt-out flag like the others, not a categorical client-kind exclusion.
2. Cross-language consistency: this only lands in Node and Rust. Please add it consistently across all SDK languages.
3. Real E2E coverage: today's tests only assert JSON-RPC serialization/types; none exercise the actual runtime behavior (discovery, delivery, tool-level cross-session messaging) against a real running runtime. Please add E2E tests that drive this through an actual agent turn once the runtime supports it, not just mocked RPC calls.

I'm moving this to draft until the runtime dependency lands and the SDK surface (and its tests) can be built against real runtime behavior. Please mark it ready for review once that's true.

dfrysinger:
Superseded by https://github.com/github/copilot-agent-runtime/pull/19586. Runtime now owns the canonical `src/sdk` source tree and includes the reviewed high-level Node and Rust cross-session messaging facades, generated bindings, and closure fixes. The companion App integration remains at https://github.com/github/github-app/pull/15507.

PR #2622, opened 2026-09-11
Fix FFI close responsiveness and callback reclamation

roji:
Follow up on #2610 to fix three concrete defects: Node could block its event loop for five seconds during close with pending outbound traffic, the Rust lifecycle regression depended on thread scheduling, and Java permanently retained JNA callback wrappers after successful close.

- Run Node's native `connection_close` through Koffi's async facility so outbound callbacks can finish on the event loop. Retain callback state and keepalive until close reports quiescence, then unregister the callback and shut down the host.
- Relax only the Rust test's initial close-call count; retain its callback lifetime, eventual reclamation, and idempotence assertions.
- Release Java's wrapper root only after successful native close, preserving it on false/exception. Keep the registration lookup through the existing shutdown-before-close ordering so a later successful close can reclaim it.

Public shutdown completion/error contracts and the graceful-stop versus force-stop distinction are unchanged. No native ABI or cross-SDK cleanup redesign is included.

### Validation

- Node: focused unit/E2E tests, typecheck, ESLint, and formatting pass. Real-native E2E suite passed 10 repetitions. On macOS ARM64 with published runtime 1.0.84-4, an independent 10 ms timer was delayed 5,007 ms before the fix and 11-12 ms afterward. Public force-stop and callback-triggered disposal both reclaimed resources successfully.
- Rust: lifecycle regression passed 100 repetitions; formatting passes.
- Java: targeted `mvn verify`, Spotless, and Checkstyle pass. The regression exercises real JNA trampoline upcalls, failed-close retention, successful-close garbage collection, and both shutdown orderings; it passed five repeated runs. Three optional packaged-runtime loading tests were skipped because the native JAR was unavailable.

### DIFF
--- a/nodejs/src/ffiRuntimeHost.ts
+++ b/nodejs/src/ffiRuntimeHost.ts
@@ -407,15 +407,15 @@ export class FfiRuntimeHost {
         }
     }
 
-    /** Closes the FFI connection, shuts down the native host, and releases resources. */
-    dispose(): void {
+    /** Awaits the initial cleanup attempt; a non-quiescent close is retried in the background. */
+    async dispose(): Promise<void> {
         if (this.disposed) {
             return;
         }
         this.disposed = true;
         this.receiveStream.end();
         if (!this.starting) {
-            void this.tryFinalizeCleanup();
+            await this.tryFinalizeCleanup();
         }
     }
 }
--- a/nodejs/src/ffiRuntimeHost.ts
+++ b/nodejs/src/ffiRuntimeHost.ts
@@ -256,7 +256,7 @@ export class FfiRuntimeHost {
         } finally {
             this.starting = false;
             if (this.disposed) {
-                this.tryFinalizeCleanup();
+                void this.tryFinalizeCleanup();
             }
         }
     }
@@ -329,11 +329,11 @@ export class FfiRuntimeHost {
         }
         this.cleanupRetryTimer = setTimeout(() => {
             this.cleanupRetryTimer = undefined;
-            this.tryFinalizeCleanup();
+            void this.tryFinalizeCleanup();
         }, CLEANUP_RETRY_INTERVAL_MS);
     }
 
-    private tryFinalizeCleanup(): void {
+    private async tryFinalizeCleanup(): Promise<void> {
         if (this.cleanupInProgress) {
             this.scheduleCleanupRetry();
             return;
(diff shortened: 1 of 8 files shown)


nodejs/src/ffiRuntimeHost.ts
```diff
@@ -402,7 +415,7 @@ export class FfiRuntimeHost {
         this.disposed = true;
         this.receiveStream.end();
         if (!this.starting) {
-            this.tryFinalizeCleanup();
+            void this.tryFinalizeCleanup();
```
stephentoub: [P2] Await the full native cleanup lifecycle before resolving shutdown

`dispose()` now starts asynchronous cleanup and returns, so `CopilotClient.stop()` and `forceStop()` can resolve while close retries, callback unregistration, and host shutdown are still pending. That changes the documented completion semantics and can race an immediate restart or filesystem cleanup with the old native host.

Returning only the current `tryFinalizeCleanup()` promise would not be sufficient because the `false` path schedules a retry and then resolves. Please use a stable completion promise spanning all retries through successful cleanup or terminal quarantine, and await it from both client shutdown paths.

Generated by Copilot

stephentoub:
I reviewed the relevant Node and Java discussions from #2610 for context.

For Node, the five-second behavior was explicitly understood: the discussion explained that synchronous `connection_close` can wait for Koffi's event-loop callback relay and hit the runtime's quiescence timeout. The decision in #2610 was not to add a timing-based E2E because it would not prove callback reclamation safety; it was not an assertion that blocking the event loop was desirable. Moving close to Koffi's async facility is therefore a sound follow-up, subject to preserving shutdown completion semantics as noted inline.

For Java, process-lifetime retention of the JNA callback wrapper was also explicit and intentional in #2610—the resolution stated that the wrapper root would remain intact after detaching its delegate. Given the clarified ABI guarantee that a successful `connection_close` is the callback-quiescence barrier, this follow-up's narrower lifetime is valid: retain on `false` or exception, and release only after successful close. Preserving the registration across host-shutdown-before-close ordering also looks correct.

Generated by Copilot

stephentoub (approved):
The Java callback reclamation and Node async-close direction are sound in light of the #2610 discussions. Approving the overall change; please address the P2 shutdown-completion issue called out inline before merging.

Generated by Copilot

nodejs/src/ffiRuntimeHost.ts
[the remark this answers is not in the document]
roji: Agreed that async close makes even a successful first attempt finish after `dispose()` returns; a completion guarantee would need to span retries, which #2610 already schedules after disposal returns. The owner explicitly deferred that cleanup-completion/force-stop contract work and limited this PR to the three concrete fixes, so I am keeping the non-blocking close and callback-lifetime safety here and deferring the stable completion promise and client await changes.

roji: The cross-retry completion concern is valid, and awaiting only the current attempt would not solve it. The owner explicitly deferred the cleanup-completion redesign, including whether force-stop should await it, and limited this follow-up to the three concrete defects; resolving this as deferred rather than adding the stable completion promise/client awaits here, while retaining callback state until native quiescence.

roji: Correction to my earlier reply: 794dc8f9 restores the initial-success completion behavior lost by the async-close change, with `stop()` and `forceStop()` awaiting that attempt through callback unregistration and host shutdown. Per the owner's scope decision, `false` still completes the initial attempt while retaining resources for detached retries, as in #2610; the stable all-retries promise remains deferred, with regressions covering both outcomes and real-runtime responsiveness verified.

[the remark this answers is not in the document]
roji: Correction to my earlier deferral: the initial successful-close timing was a regression introduced here, now fixed in 794dc8f9 by returning the initial cleanup promise and awaiting it from both client shutdown paths through callback unregistration and host shutdown. A `false` result still retains resources, schedules the existing detached retry, and finishes that initial operation; the broader all-retries completion contract remains deferred, and focused tests plus real-runtime responsiveness/restart checks pass.

PR #2623, opened 2026-09-11
Fix replay race in Python persisted-session RPC test

roji:
The persisted-session RPC test enqueued an LLM turn without a replay snapshot, then raced metadata inspection and session detachment against inference. Delaying metadata polling reproduced unmatched model requests in 10/10 runs even though the test passed.

Add the missing replay fixture, await the completed turn, and assert its response before inspecting persisted state. Keep the real user message because it triggers runtime persistence, and preserve all existing list/find/inspect and discriminator assertions. No production code, timeout, retry, or workflow changes.

## Validation

- The completion check fails without the snapshot.
- Twenty delayed executions across default and in-process contexts passed, each with exactly one captured model request and no unmatched requests.
- Both implicated test modules and repeated paired runs passed locally; Ruff checks passed.
- [All nine Python workflow jobs passed](https://github.com/github/copilot-sdk/actions/runs/34617308704), including Windows/in-process with 902 passed and 10 skipped. The persisted-state and todos tests completed normally.

## Investigation limits

Discovered while investigating the merge-queue failure for #2622. This fixes the demonstrated replay race, but does not conclusively identify the cause of the original 300-second Windows timeout. A separate post-crash pytest-xdist scheduler deadlock was reproduced independently; its upstream fix is unreleased and is not included here.

### DIFF
--- a/python/e2e/test_rpc_server_e2e.py
+++ b/python/e2e/test_rpc_server_e2e.py
@@ -311,17 +311,19 @@ class TestRpcServer:
                 on_permission_request=PermissionHandler.approve_all,
             )
 
-            await session.send(
-                "Record a turn for sessions.list discriminator coverage", mode="enqueue"
+            # Persistence requires a user message. Finish its replayed turn before
+            # inspecting state or detaching, rather than racing an active inference.
+            message = await session.send_and_wait(
+                "Record a turn for sessions.list discriminator coverage"
             )
+            assert message is not None
+            assert message.data.content == "Recorded."
 
             listed = None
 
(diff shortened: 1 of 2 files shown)


ArshVermaGit (commented):
**Approve.** The fix is correct, tests are well covered, and the change is limited to the flaky test.

stephentoub:
@roji, the PR is empty?

roji:
Ah, interesting - copilot identified as already fixed in #2642 and "helpfully" decided to empty the PR... Thanks for looking.

PR #2632, opened 2026-09-12
[Chore] Handle Rust Unknown Enum Collision

MRayermannMSFT:
## What
Rust codegen gives the wire value `"unknown"` a stable `UnknownValue` variant when the catch-all already uses `Unknown`. The [successful branch workflow run](https://github.com/github/copilot-sdk/actions/runs/34661240430) generated the CLI update draft with this fix.

## Why
Copilot CLI `1.0.84-5` adds `"unknown"` to `CatalogTrustEligibility`. The existing name collision stops the automated CLI update workflow before it can create a pull request.


stephentoub:
This is duplicating https://github.com/github/copilot-sdk/pull/2631, right?

PR #2649, opened 2026-09-14
Fix the .NET manual tool resume sample build

rinceyuan:
`dotnet/samples/ManualToolResume.cs` is a file-based app (it opens with `#:project ../src/GitHub.Copilot.SDK.csproj`), and it does not build.

```console
$ cd dotnet/samples && dotnet build ManualToolResume.cs
ManualToolResume.cs(41,9): error GHCP001: 'GitHub.Copilot.Rpc.PermissionDecisionApproveOnce'
is for evaluation purposes only and is subject to change or removal in future updates.
Suppress this diagnostic to proceed.
```

The sample calls `HandlePendingPermissionRequestAsync` with `new PermissionDecisionApproveOnce()`, and that type is marked experimental in the generated surface:

```csharp
// dotnet/src/Generated/Rpc.cs
[Experimental(Diagnostics.Experimental)]
public partial class PermissionDecisionApproveOnce : PermissionDecision
```

`ExperimentalAttribute` reports as an error unless suppressed, so the sample cannot be built or run as shipped.

## The change

Suppress the diagnostic around that one call, matching how the rest of the repository already handles it:

- `dotnet/src/Client.cs` wraps experimental usages in `#pragma warning disable GHCP001` / `restore` in several places.
- `docs/features/usage-and-billing.md` tells readers to use exactly `#pragma warning disable GHCP001` or a project-level `<NoWarn>GHCP001</NoWarn>`.
- `docs/setup/in-process-runtime.md` lists "Allow the `GHCP001` experimental API diagnostic" as the .NET requirement.

I used the bare pragma with no explanatory comment, since that is the form used both in `Client.cs` and in the docs snippets.

## Verification

Both samples build after the change:

```console
$ cd dotnet/samples
$ dotnet build ManualToolResume.cs      # exit 0
$ dotnet build Chat.cs                  # exit 0
```

`Chat.cs` already built before this change; only `ManualToolResume.cs` was affected. The build produces no warnings, and the diff is two added lines with no other edits.

## Context

Nothing builds the samples in CI, which is why this went unnoticed — the same gap behind #2648 (the Go samples module). I have not added CI coverage here either, since where that belongs seemed like your call.

For completeness, I checked the other two sample sets while I was at it and they are fine: the Node samples type-check against `nodejs/src`, and every symbol the Python samples import resolves against `python/copilot`.


rinceyuan:
Thanks — I took the factory half of this and verified the rest, which did not hold up.

**Adopted:** the sample now calls `PermissionDecision.ApproveOnce()` instead of `new PermissionDecisionApproveOnce()`. That matches `dotnet/src/PermissionHandlers.cs` and the factory's own doc comment ("Use these for quick discoverability via `PermissionDecision.<dot>`"), so it is better sample style regardless.

**Kept:** the suppression, because removing it does not build. The diagnostic simply moves from the derived type to the base type:

```console
$ dotnet build ManualToolResume.cs      # factory, no pragma
ManualToolResume.cs(41,5): error GHCP001: 'GitHub.Copilot.Rpc.PermissionDecision'
is for evaluation purposes only and is subject to change or removal in future updates.
```

The base class carries the attribute, so every member of the hierarchy is affected:

```csharp
// dotnet/src/Generated/Rpc.cs
[Experimental(Diagnostics.Experimental)]
[JsonPolymorphic(...)]
public partial class PermissionDecision
```

`PermissionDecision.ApproveOnce()` is declared in the hand-written partial in `dotnet/src/PermissionDecision.cs` and carries no attribute of its own, but its return type is the experimental base, so the call site still needs the suppression.

Current state builds clean:

```console
$ dotnet build ManualToolResume.cs      # factory + pragma
exit 0, no errors or warnings
```

PR #2657, opened 2026-09-14
Add Python 3.15 support

brettcannon:
Python 3.15 is the newest supported Python release, but the package metadata and CI coverage currently stop at earlier versions. This change advertises 3.15 support and continuously verifies both ends of the supported range.

- Add the Python 3.15 PyPI classifier.
- Run the Python SDK test matrix on Python 3.11 and 3.15 across Linux, macOS, and Windows for both default and in-process transports.
- Allow prerelease Python resolution so CI can test 3.15 release candidates before GA and naturally use stable 3.15 afterward.

**Validation**
- Compiled the SDK and parsed its package metadata in a virtual environment using CPython 3.15.0rc2 free-threaded.
- `git diff --check`

The full local dependency/test run was not available because the sandbox denied uv temporary/cache writes; CI provides the complete test coverage.


brettcannon:
FYI [Python 3.15 is due out](https://peps.python.org/pep-0790/#schedule) October 1.

brettcannon:
The Python 3.11 failure looks to be some testing timeout issue and is not related to the PR.

PR #2659, opened 2026-09-14
Rust: Isolate hostless runtime cache

ellismg:
## Overview

Separate the non-bundled hostless runtime cache from the bundled CLI cache. Non-bundled in-process builds now extract to `<platform cache>/github-copilot-sdk/runtime/<version>`, while bundled CLI extraction remains at `cli/<version>`.

This prevents stale hostless-runtime cleanup from deleting a same-version bundled `copilot` executable used by another application. `COPILOT_CLI_EXTRACT_DIR` retains its existing direct-directory semantics.

## Changes

- Share hostless runtime path computation between `build.rs` and runtime resolution.
- Document the distinct cache paths on macOS, Linux, and Windows.
- Add regression coverage proving runtime cleanup cannot delete or overwrite the bundled CLI directory.
- Update focused non-bundled resolution tests for the new namespace.

## Validation

- `rustup run nightly-2026-04-14 cargo fmt --check`
- `cargo test --no-default-features --lib cache_paths::tests`
- Focused `cli_resolution_test` cases for conventional extraction, runtime resolution, and `COPILOT_CLI_EXTRACT_DIR`
- `cargo clippy --no-default-features --lib -- -D warnings`


ellismg:
This will be helpful for us on the copilot app team.  Right now we can run into an issue without this fix which causes the CLI we use as part of our dev builds when runing out of main to get deleted and replaced with an incompatible version, breaking things until we restart the dev version of the app and redownload the correct version.

jmoseley (approved):

PR #2663, opened 2026-09-15
Harden Rust CLI download retries

gimenete:
## Summary

- increase transient Copilot CLI download retries from 3 to 5 in both Rust build modes
- extend the bounded exponential-backoff window from 7 seconds to 31 seconds while preserving fail-fast behavior for non-transient failures

Fixes the transient download failure observed in [github/github-app CI run 34929070862](https://github.com/github/github-app/actions/runs/34929070862).

## Validation

- `cd rust && cargo +nightly-2026-04-14 fmt --check`
- `cd rust && cargo clippy --all-features --all-targets -- -D warnings`
- `cd rust && cargo test --all-features --lib` (248 passed)
- `cd rust && cargo test --all-features` reached the E2E target; 371 E2E tests could not initialize because `tsx` is not installed in the local workspace (`sh: tsx: command not found`)


ArshVermaGit (commented):
Looks good to me. The retry count is updated consistently for both in-process and out-of-process builds, and the existing fail-fast behavior for non-transient errors is preserved. The checks are passing as well. No issues from my side.

stephentoub (approved):

PR #2676, opened 2026-09-15
fix(rust): cut retained runtime-install memory by ~99%

### LINKED ISSUE #2675: Rust bundled runtime installation retains large native-image buffers on cold and warm startup
## Description

The Rust SDK's `install_bundled_runtime()` path allocates complete runtime archive entries and reads installed files into full-size buffers. `install_hostless_assets` handles `runtime.node`, then `install_runtime_pair` extracts and compares it again. On macOS, those allocations can leave substantial physical memory retained by the allocator even after the installer returns, including when every installed file is already valid.

This is an installer-only reproduction, not evidence of a live-object leak or a measurement of a client/model session. No authentication, CLI subprocess or service request is needed.

## Reproduction

At SDK revision `a675b55531a9dfc647ee015e32d74279568550f3`, build a release Rust executable with default `bundled-cli` features which calls `github_copilot_sdk::install_bundled_runtime()` once and stays alive for one second afterward. Run the executable with a fresh isolated `HOME`. For the warm case, first seed that same home's SDK cache using a separate process, then run the measured executable in a new process. Do not reuse the same process, because its `OnceLock` bypasses installation.

Measure process-lifetime peak RSS/physical footprint with macOS `/usr/bin/time -l`, and current RSS/physical footprint using `proc_pid_rusage(RUSAGE_INFO_V0)` after the one-second idle period. Keep allocation-stack logging disabled in these measurements.

## Observed SDK-only baseline

Apple M4 Pro, 48 GiB RAM, macOS 26.6.2 (25G83), arm64; rustc 1.94.0; release optimization level 3, debug level 1. Public bundled runtime `1.0.84-8`. Five independent processes per cohort; medians (minimum-maximum):

| Cohort | Retained physical footprint, MiB | Installer seconds |
| --- | --- | --- |
| Cold installation | 155.438 (154.563-156.047) | 0.880 (0.863-1.037) |
| Valid warm installation | 174.360 (172.735-174.907) | 1.031 (0.997-1.092) |

The installed `runtime.node` is 71,166,736 bytes, SHA-256 `839cd681c72cb92f27697d5e3e3ee96d7bb8df4234ffb5f7b442956828ec173a`. Filtered embedded runtime archive SHA-256: `6c43b789080fc06b25d406af8fae709daa99f0724c4d290cc8a31160c5a3ad64`.

A separate instrumented warm process recorded two page-rounded 71,172,096-byte VM allocations through the SDK runtime installer, one also through `std::fs::read`; after installation, `vmmap` reported 165.7 MiB in `MALLOC_LARGE (empty)` regions. Instrumented values are excluded from the comparison above.

## Expected behavior

Runtime extraction and existing-file verification should use bounded buffers, preserve exact output bytes/modes, permit valid read-only warm caches, and repair corrupt files with staged atomic replacement. The full CLI installer is a separate path and outside the memory optimization scope.

A focused fix with runnable reproduction and cold/warm/corrupt/truncated measurements is prepared.

mohamedmansour:
**Problem:** The Rust runtime installer repeatedly loaded entire native files into memory for extraction and comparison, even when the runtime was already installed. The buffers were freed, but the allocator retained the memory after installation. Fixes #2675.

**Fix:** Stream extraction and compare existing files in fixed-size chunks. Stage only missing or changed files, validate the archive, then replace files atomically. Matching files are reused without requiring a writable cache.

**Proof:** Matched SDK-only measurements on macOS arm64, runtime 1.0.85, five alternating pairs per case:

| Retained physical memory | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| Cold install | 156.19 MiB | 1.98 MiB | 98.7% |
| Warm install | 175.84 MiB | 1.86 MiB | 98.9% |

All 68 installed files remain byte-identical. Regression tests cover corruption repair, archive integrity, concurrent installation and read-only caches. Native Windows/Linux validation remains outstanding.


rust/src/embeddedcli.rs
```diff
@@ -326,90 +323,283 @@ fn install_hostless_assets(install_dir: &Path, archive: &[u8]) -> Result<(), Emb
...
+        }
+        if is_library {
+            #[cfg(feature = "bundled-in-process")]
+            if path != Path::new(RUNTIME_LIBRARY_NAME) {
+                continue;
+            }
+            #[cfg(not(feature = "bundled-in-process"))]
             continue;
         }
+        if !seen.insert(path.clone()) {
```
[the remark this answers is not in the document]
mohamedmansour: Reproduced the overwrite through the installer and fixed it in 9a1af09e: destinations are now separator-normalized and deduplicated case-insensitively before artifact selection, with non-portable Unicode/DOS/trailing-dot aliases rejected consistently without probing read-only caches. Added cold, warm, corrupt-cache, mixed-separator and native-library regression coverage; all 27 installer tests and all-feature/all-target Clippy pass locally on macOS arm64.

stephentoub:
There is one P2 behavioral regression that should be fixed before merge.

On Unix, `existing_runtime_file` requires the installed file's entire `0o777` mode to exactly match the tar header. That breaks valid immutable warm caches: a common `chmod -R a-w` changes `0644` to `0444` and `0755` to `0555` while preserving contents and the wrapper's executable bits. These files are then treated as stale, and the installer tries to create staging files in the non-writable cache and fails. The previous implementation compared contents only and reused such a cache.

The new read-only-cache test only removes write permission from the directories, so it does not exercise this case. Please make warm validation ignore irrelevant write-bit differences (while still requiring the executable bits needed by the wrapper), and extend the test to remove write bits from the installed files as well.

I did not find another P0/P1/P2 issue. The streaming design and archive/publish hardening otherwise look directionally sound.

Generated by Copilot

mohamedmansour:
@stephentoub Accepted and fixed in 61cb61c9: I reproduced the `Permission denied` failure from [your report](https://github.com/github/copilot-sdk/pull/2676#issuecomment-5691879139), changed Unix warm validation from `0o777` to `0o555` to ignore write-bit differences while retaining read/execute checks, and extended regression coverage to read-only installed files and each missing wrapper execute bit. The regression now passes, along with 486 non-replay tests, 25 default-feature installer tests, 30 doctests, formatting, and Clippy; full replay E2E validation remains blocked by local setup (initially missing `tsx`, then the macOS socket-path limit with project-local scratch paths).

rust/src/embeddedcli.rs
```diff
@@ -298,118 +306,336 @@ fn install_hostless_assets(install_dir: &Path, archive: &[u8]) -> Result<(), Emb
...
+        .mode()
+        .map_err(|e| EmbeddedCliError::new(EmbeddedCliErrorKind::Archive, e))?
+        & 0o555;
+    #[cfg(unix)]
+    let matches = {
+        use std::os::unix::fs::PermissionsExt;
+        matches
+            && metadata
+                .as_ref()
+                .is_some_and(|metadata| metadata.permissions().mode() & 0o555 == mode)
```
stephentoub: This still requires all read and execute bits to match exactly (`actual & 0o555 == expected & 0o555`). That fixes the reported `chmod -R a-w` case, but a valid owner-only immutable cache (`0644` → `0400`, `0755` → `0500`, with directories at `0500`) remains readable/executable by its owner and is still treated as stale; staging then fails because the cache is non-writable. Please make warm validation depend on actual readability and only the executable permission needed by the wrapper, and add owner-only read-only coverage.

Generated by Copilot

stephentoub (approved):
Approved with the remaining permission-mode concern noted inline.

Generated by Copilot

rust/src/embeddedcli.rs
mohamedmansour: Accepted and fixed in 4c217387: I reproduced the owner-only immutable-cache failure, removed exact mode matching, and now establish readability by hashing while checking executable access with the process's effective credentials through rustix (already in the lockfile). Added 0400/0500 cache coverage and no-reextraction tests for valid reduced permissions; all 35 installer tests, formatting, and all-feature/all-target Clippy pass.

rust/src/embeddedcli.rs
```diff
@@ -298,118 +304,314 @@ fn install_hostless_assets(install_dir: &Path, archive: &[u8]) -> Result<(), Emb
...
+        }
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => None,
+        Err(e) => return Err(EmbeddedCliError::new(EmbeddedCliErrorKind::Io, e)),
+    };
+    let matches = metadata
+        .as_ref()
+        .is_some_and(|metadata| metadata.len() == size);
+    if matches {
+        match fs::File::open(target) {
+            Ok(file) => return Ok(Some(file)),
```
stephentoub, line 571: This warm path now treats a byte-correct runtime wrapper as valid without checking that the current process can execute it, and the new test explicitly accepts mode `0o400`. That undoes the effective-credentials execute check added in `4c217387` after the earlier review feedback. On Unix, stripping the wrapper's execute bits makes `install_runtime` return `Ok(path)`, but launching that path fails with `EACCES`; subsequent installs continue preserving the unusable mode. Please require effective execute access for `RUNTIME_BINARY_NAME` only: modes such as `0o500`/`0o700` should remain reusable, while `0o400`/`0o600` should be repaired or rejected.

Generated by Copilot

mohamedmansour: Accepted and fixed in 1dfcddeb: the installer now uses a dependency-free POSIX faccessat binding with effective credentials for RUNTIME_BINARY_NAME only, repairs nonexecutable wrappers and rechecks before returning, while preserving 0500/0700 wrappers and readable 0400/0444 native libraries; Windows behavior is unchanged. I reproduced the public-API EACCES failure before the fix and verified successful runtime RPC pings after repair, immutable-cache reuse, and ACL-denial handling on macOS; all-feature tests (922 passed), no-default-feature library tests (232 passed), strict Clippy, and formatting pass, with Linux/Windows native validation left to PR CI.

stephentoub (approved):
The remaining runtime-wrapper execute-access issue is correctly addressed, and I found no new P0/P1/P2 concerns in the update. Approving the code; the current-head workflows still need to be authorized and pass before merge.

Generated by Copilot

PR #2700, opened 2026-09-16
nodejs: enforce exact npm dependency policy

roblourens:
## Summary

- pin the Node.js SDK's published direct production dependencies to exact versions
- default future npm saves to exact versions
- verify every external package/version in the resolved production closure from `package-lock.json` has been published for at least seven full days before release packaging
- include optional and `devOptional` production nodes while excluding strictly dev-only, local-link, and bundled entries
- verify packed release manifests preserve the exact source dependency requirements

## Why

The seven-day Dependabot cooldown added in #2596 delays dependency update PRs, but a published semver range can still resolve to a newer release immediately. Exact direct requirements keep downstream direct dependency selection deterministic while Dependabot continues to automate upgrades after the cooldown.

The lockfile closure check also prevents releasing when the complete production graph resolved and tested by CI contains a newly published external transitive or optional package. This includes Koffi's 18 `@koromix/*` optional platform packages.

This unblocks downstream mirrored feeds such as VS Code's ADO feed, where newly published public npm packages may not be available until their quarantine period ends.

## Scope

A library's `package-lock.json` is not published or honored for consumer installs. This check therefore proves that the release-time production graph is at least seven days old, but it does not freeze future consumer resolution of ranges declared by transitive dependencies. Fully freezing that graph would require publishing an `npm-shrinkwrap.json` or bundling dependencies, which is outside this change.

The selected direct production dependency versions and authoritative npm publication timestamps are:

- `koffi@3.2.1` — `2026-09-04T07:39:01.277Z`
- `vscode-jsonrpc@8.2.1` — `2024-05-21T13:53:09.125Z`
- `zod@4.3.6` — `2026-01-22T19:14:35.382Z`

`@github/*` packages remain exempt from the public npm publication-age check, matching the existing Dependabot cooldown configuration.

## Validation

- `npm ci --ignore-scripts` for the Node SDK, test harness, and samples
- `npm run format:check`
- `npm run lint` (five existing warnings, no errors)
- `npm run typecheck`
- `npm run build`
- `npm test` — 1,147 passed, 11 skipped
- dependency policy tests — 14 passed
- `npm run verify:dependency-policy` — 3 exact direct dependencies and 21 resolved production package versions passed the seven-day check
- policy runtime — 0.62 seconds locally
- `npm run pack:release`
- `npm run verify:release-packages` — nine release package tarballs verified
- `git diff --check`

(Written by Copilot)


stephentoub:
Blocking issue: the seven-day publication-age verification currently checks only the direct `dependencies` and `optionalDependencies` declared in the root `package.json`. It does not walk the resolved production graph in `package-lock.json`, so transitive or optional packages—such as Koffi's `@koromix/*` packages—can be newer than seven days while the check still passes. This leaves the transitive-dependency failure mode from the proposed policy uncovered. Please verify every resolved external production package/version, while accounting for the fact that exact top-level requirements alone do not freeze ranged transitive dependencies for downstream consumers.

Generated by Copilot

stephentoub (approved):
Approved with the blocking transitive-dependency policy gap documented in the separate review comment.

Generated by Copilot

roblourens:
Addressed the documented gap in 4e0ab199.

The policy now walks the `packages` map in `package-lock.json` and verifies every unique resolved external production package/version. It includes optional and `devOptional` production nodes (including all 18 Koffi `@koromix/*` platform packages) and excludes entries npm marks strictly dev-only, local links, and packages bundled inside another tarball.

The live check now covers 21 resolved production package versions and takes 0.62 seconds locally. I also added tests for optional, `devOptional`, strictly dev-only, nested, duplicate-version, link, bundled, and malformed lockfile entries.

The PR description now explicitly states the remaining npm limitation: this validates the graph resolved and tested at release time, but a library's `package-lock.json` does not freeze later consumer resolution. That stronger guarantee would require shrinkwrap or bundling.

(Written by Copilot)

PR #2741, opened 2026-09-21
Expose host-user hook opt-in across all six SDKs

SteveSandersonMS:
## Summary

Expose optional/nullable `enableHostUserHooks` configuration for both create and resume in Node.js, Python, Go, .NET, Java, and Rust.

- Resolve and explicitly serialize a boolean before the initial RPC: Empty defaults to false, CopilotCli defaults to true, and explicit false/true always wins.
- Preserve config copying/reuse conventions and SDK callback registration. Resume uses the current caller's choice or mode default rather than retaining a persisted grant.
- Regenerate all six SDK bindings from the paired locally modified runtime schemas, including the existing options-update surface where applicable.
- Update each SDK's public documentation and the multi-tenancy guide with source scope, managed-policy independence, and the fact that SessionFs does not sandbox host hook capabilities.
- Require protocol 4 for the paired breaking runtime contract; do not probe or fall back to older runtime behavior.

Companion runtime draft: https://github.com/github/copilot-agent-runtime/pull/22124. Both changes must be coordinated. No consuming application changes are required.

## Validation executed

- All six SDKs: targeted create/resume mode matrices, explicit-false serialization, config reuse/builders, callback preservation, protocol rejection, and language build/type checks passed.
- .NET: 49 targeted tests; Java: 139; Rust: 2 host-hook tests plus 10 protocol tests. Node/Python/Go targeted suites also passed; Go race checks and whole-tree vet passed.
- Rust generator regression suite: 19 passed.
- Node real-native integration: 32 passed on Node 24.21.0 across subprocess stdio, external TCP, and in-process FFI. Covers SessionFs, isolated host settings/directory fixtures, startup prompts, commands, HTTP, SDK callbacks, opposing sessions, live updates, reload, and warm/cold resume. No live model or real credentials were needed.
- Generated and handwritten counterparts were synchronized with the runtime's embedded SDK.

Integration used the locally modified runtime, not a bundled release:

```text
/workspace/copilot-agent-runtime/dist-cli/prebuilds/linux-x64/copilot-runtime
SHA256 32480e7e7710e567ee89a212c22b95bfeaeeb5edc166ee0bf36deb2c2fa79ffe
Adjacent runtime.node:
SHA256 a86674c487a1039aedc688b37e1dece59832d8cd6a05504a9b034a84764fdc70
```

## Generated-diff notes

Current local runtime schemas contain connector/event/command-result changes absent from the previously pinned SDK schema. Those unavoidable generated differences are retained. The protocol generator's existing Go capitalization/.NET namespace drift and missing Java output were corrected; the Rust generator now emits the command-result union, and existing Go assertions use its generated accessor. Generated bindings were not hand-patched.

Full SDK E2E suites were not run; real-native transport coverage is the focused 32-case Node suite above. No new transport support, public SDK options-update APIs, or hook capabilities were added.


SteveSandersonMS:
Superseded by https://github.com/github/copilot-agent-runtime/pull/22124 following the SDK source migration into src/sdk. The runtime PR is rebased onto current main and preserves all 158 paths from the SDK implementation commit 7dff12bf0c60ded9d3d6b3322ac982dffcfbca99. The later automated Java regeneration against the old pinned schema removed the new contract fields and is intentionally not carried forward; the consolidated PR retains bindings generated from the locally modified runtime schema. Further development and review should take place in the runtime PR.

